PR #22, opened 2014-10-30
Feature/nsmutabledict category

### LINKED ISSUE #21: Clean up SDL Structs with a category on NSMutableDictionary
Every setter method in every SDLRPCStruct subclass takes an object and sets it in the store. If the object passed into the method is nil, the object is removed from the store instead (because passing nil to setObject throws an NSInvalidArgumentException).  This makes every setter look like

```ObjC
-(void) setNavigationText:(NSString*) navigationText {
    if (navigationText != nil) {
        [store setObject:navigationText forKey:NAMES_navigationText];
    } else {
        [store removeObjectForKey:NAMES_navigationText];
    }
}
```

This can be cleaned up by creating a category on NSMutableDictionary and including it in the SDLRPCStruct class.  It would look like

```ObjC
- (void)setOrRemoveObject:(id)object forKey:(id <NSCopying>)key {
    if (nil != object) {
        [self setObject:object forKey:key];
    } else {
        [self removeObjectForKey:key];
    }
}
```

Then every setter can be reduced to

```ObjC
- (void)setNavigationText:(NSString *)navigationText {
    [store setOrRemoveObject:navigationText forKey:NAMES_navigationText];
}
```

justinjdickow:
fixes #21


joeljfischer:
Note that to be comprehensive, this category would have to be used ~370 more places

justinjdickow:
@hugh22 @adein Can somebody please review this change so we can merge the pull request?

mikeburke106:
This PR has been sitting in limbo for ~3 months and we've asked for updates multiple times with no response.  I propose that we discuss this PR in the next call and then allow 2 more weeks for a final review.  If all is good by the following call, I propose we mark this PR for approval and merge.

I understand that everyone is very busy, but please do what you can to put a priority on these pull requests, as they are the future of the project.  The more people we have reviewing them, the fewer mistakes will make it into the code.  Thanks everyone.

mikeburke106:
I just reviewed the commit diffs and everything looks good to me.

joeljfischer:
This needs to be into the `develop` branch

PR #33, opened 2015-01-20
Application Launching (SDL 4.0)

joeljfischer:
## Summary
This PR will involve several pieces to allow an iOS application to track available applications on the users phone that can be launched with SmartDeviceLink. This will involve 4 major pieces.

1. Head Unit tracking of the foregrounded application on a user's phone.
2. Head Unit requested querying of a server URL to retrieve a list of applications that are SDL compatible.
3. Parsing the response from the server, and determining which apps installed on the user's phone can be launched, and sending the filtered response to the HU.
4. Responding to a request from the HU to launch a specific app, and launching it.

### Tracking Foregrounded Application
The SDL Proxy object, upon opening a connection, will poll the application for its state, and send that across. It will also immediately set up observers on application state changes. Upon firing of those observers, it will recheck the application for its state, and push a new notification to the HU.

This PR also introduces the ability to send notifications from the Mobile API (issue #34). This is required because application state changes are sent via a pre-existing RPC Notification that was previously only sent from the HU, onHMIStatus. FULL is equivalent to foreground, LIMITED is equivalent to background.

### Query Apps Server HTTP Request
The HU will request, via a OnSystemRequest, that the mobile device retrieve an application list from the server. The application *may* (this has not yet been determined) cache the response for a specified period.

### Query Apps Filter Installed Applications
The server will respond with a set response, [described here](https://github.com/smartdevicelink/sdl_server/blob/master/docs/application_launching_v1.0.md) which will be parsed by the library. The library will then run a system call to determine which apps are installed on the phone. The library will then filter the list passed by the server so that only installed apps are represented, and forward it on to the HU.

### Launch App
The HU will send an OnSystemRequest with a URL field that contains the URL scheme of the app they wish to launch to the foregrounded application. The library running in that application will then attempt to launch the application specified by the url scheme passed in. This will, of course, foreground that application, and background the application making the call.

##### General
- [ ] Make sure the proxy can handle being at v4

##### Tracking Foreground State
- [x] Track Application State
- [x] Add the ability to send RPCNotifications (by sending any type of RPCMessage)
- [x] Send an OnHMIStatus notification to connected HU when application state changes

##### Query Apps Server HTTP Request
- [ ] Capture OnSystemRequests for Query Apps and Launch App commands
- [ ] Make a query apps web call to the server specified in the OnSystemRequest to retrieve a list of URLs
- [ ] Parse the response from the server, determine applications to be checked

##### Query App Filter Installed Applications
- [ ] Test each application to see if it's installed on the phone
- [ ] Filter the response from the server based on which apps are installed on the phone
- [ ] Send the filtered response to the Head Unit

##### Launch App
- [ ] Parse a Launch App OnSystemRequest
- [ ] Open the specified application from a Launch App OnSystemRequest


joeljfischer:
@justinjdickow does this initial PR need to do anything else?

joeljfischer:
This was mistakenly merged, and was reverted. See PR #45 now.

PR #50, opened 2015-01-22
[WIP] Documentation and misc updates

Mike-FoMoCo:
Made changes corresponding to issue #15 with commit 72ba45cc0a2853c9cfc10ecd8db70acf30082703.

PR #51, opened 2015-01-22
Add a CONTRIBUTING file

joeljfischer:
This is necessary to show people who wish to create a pull request information that they need to know. It will also contain information regarding the CLA when that's ready.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,7 @@ We use Gitflow as our branch management system. Please read up on it [here](http
 
 * All feature branches should be based on `develop` and have the format `feature/branch_name`.
 * Minor bug fixes should be based on `master` and have the format `hotfix/branch_name`.
-* All commits should involve logical units. Please do not pull all changed code in one commit, unless it is a very minor change.
+* All commits should involve logical units. Please do not put all changed code in one commit, unless it is a very minor change.
 * Work in progress pull requests should have "[WIP]" in front of the Pull Request title. When you are ready to merge, remove this tag and @mention `smartdevicelink/developers` to get it scheduled for review.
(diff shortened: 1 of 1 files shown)


adein:
@joeljfischer After talking with @mrapitis and @anildahiya, we have a few questions:

1. Hotfixes
    * How/Who determines if an issue/bugfix should be a hotfix?
    * The contributing guide says branch off of master, using "hotfix/branchname", and pull request back in to master.  However, looking through the closed pull requests, some hotfix pull requests with this format were accepted, some hotfix pull requests that didn't use this were also accepted, and some bugfix pull requests were rejected because they were in to master.  Maybe this is just because the process hadn't been defined yet?
    * I'm assuming hotfixes are only to fix bugs that are already in production code?  How does this affect releases?  Is there a new release tag, etc. for the bugfix release?
2. Deleting branches after merging
    * When a pull request is merged, how can/will developers know when to delete their branch in their forked repo?

justinjdickow:
1. 

  a. We are defining a hotfix as a pull request which resolves a documentation issue or bug in a non-API changing way. In contrast, adding public methods is always a feature. Said another way, anything that would only change the patch version in [semantic versioning](http://semver.org/) would be considered a hotfix.

  b. This is because we are still in the process of transitioning into the better defined git flow. The #50 pull request should not necessarily have been rejected because it was into master, but rather because there were documentation updates as well code changes which were not described in the pull request and should have been separated out.

  c. There might be a need to hotfix the develop branch, we're unsure if it's the same process to hotfix a release - we should discuss. When you hotfix master, you create a new patch version for master, and rebase develop and feature branches. The hotfix increases the patch version on master, and it gets a new tag.

  2. Once it's merged in it can be deleted, but it's pretty much completely up to them whether or not they delete the branch

 -joel, justin

PR #52, opened 2015-01-22
Feature/nsmutabledict category

justinjdickow:
Please see #22 for related comments


joeljfischer:
safeSet: might be a better name than setOrRemove:

joeljfischer:
@justinjdickow how would you feel about re-implementing this on 4.0? If not, I think this can be closed for now, but we can leave the issue open.

PR #54, opened 2015-01-26
Fix incorrect case with SDLHMIZoneCapabilities

joeljfischer:
WARNING: This is a backward-incompatible change

The case of the file and class are different. This resolves that change by making the acronym for HMI fully uppercase, changing the class to match the file.


joeljfischer:
Consider resolving via #71

Issue #15, opened 2014-10-28
HeaderDoc style documentation in files


joeljfischer:
[Tutorial on HeaderDoc](http://www.raywenderlich.com/66395/documenting-in-xcode-with-headerdoc-tutorial)
[Jazzy](https://github.com/realm/jazzy)

We should document all of our methods next to the methods in header files. This will provide us with in Xcode documentation generation for alt-clicking on methods; as well as the opportunity to generate html based on the documentation.

joeljfischer:
@Heath-FoMoCo @adein Can we get an update on this?

Heath-FoMoCo:
Yes.  We are currently in the process of moving our 2.3.1 iOS proxy release into  a pull request from GIT Hub.  This will include most recent changes and our documentation.  Expectation is early next week.

joeljfischer:
Awesome, thanks

PR #66, opened 2015-02-03
Feature/inline documentation

Mike-FoMoCo:
Addressed Issue #15, adding inline documentation changes only.


joeljfischer:
Please rebase this on master, then squash the commits into one, to keep the history simpler. If you're okay with it, we can transfer this branch into the repository and do it ourselves as well.

joeljfischer:
In those files where the newline is deleted after `@end`, please add the newline to the end of the file. It sounds like a minor thing, but we are seeing warnings about it.

joeljfischer:
It appears that the newline issues may have been caused by a Windows machine editing the files and appending a Windows line ending (CRLF) instead of an OSX one (LF). If you can use a text editor that lets you modify your line endings (Sublime Text 3 has this, for instance), it would be great if you could modify those to match Unix style endings.

Heath-FoMoCo:
@joeljfischer Is there a specific file the CRLF is an issue on?  Or is it all files?
What do you mean by warnings?  Is it a compile warning in xCode?
Is this something that can be rectified during the merge?

We aren't seeing this on our end right now -- so a little clarification would help.

joeljfischer:
If you go over the the files changed tab at the top of this PR, you'll see a few files that have their `@end` line changed with new terminating characters. We were seeing warnings within the git repository client we are using. I'm guessing that Xcode will probably reformat the files correctly, but I haven't yet had a chance to pull down the PR to find out.

Mike-FoMoCo:
I can see where the newlines are removed when looking at the "files changed" tab, but I don't see any terminating characters replacing them, even when pulling the file fresh and opening directly in Sublime. I don't doubt your claims, just having trouble seeing it on our end. Looks like there's only 12 instances of it, is this something that can be fixed during the merge as Heath suggested?

joeljfischer:
It's entirely possible that I misread the new line endings. Since there were two characters there, I assumed it was a CRLF. It could be "no newline" instead. I do know that Xcode likes newlines at the end of all files. With that said, I just pulled down the PR and I'm not seeing any errors or warnings regarding it. @justinjdickow was the one who brought it to my attention. If he has anything to add about the warnings he saw, he can add them, otherwise I don't think it's an issue.

justinjdickow:
If you look at the pull request it appears there are no new lines at the end of a few of the files. I'm not sure that's necessarily an issue, just something that GitHub warns us about.

joeljfischer:
I'm reviewing the documentation. It appears that some files remain undocumented, largely ones that developers shouldn't have to interface with, such as `SDLEnum`, `SDLFunctionID` et al. Is that intentional at this stage?

I noticed that many methods documented with `@result` start their description with 'returns'. This results in a the line reading 'Returns return the dimension' for example.

In addition, there are some files that seem like they should have documentation that don't, or don't fully.

##### Proxy
* SDLProxy
* SDLJingle
* SDLLockScreenManager
* SDLProxyListener
* SDLProxyFactory
* SDLRPCRequestFactory
* SDLTTSChunkFactory

##### Enums
* SDLAmbientLightStatus
* SDLCharacterSet
* SDLDisplayType
* SDLGlobalProperty
* SDLIgnitionStableStatus
* SDLPowerModeQualifications
* SDLPowerModeStatus
* SDLPredefinedLayout
* SDLPrerecordedSpeech
* SDLRequestType
* SDLResult
* SDLTBTState
* SDLTimerMode
* SDLTouchType
* SDLTriggerSource
* SDLVehicleDataType

I was too lazy to go through the rest, except to note that the struct `SDLScreenParams` was also missing documentation, I only noticed it because it's also in the wrong group.

joeljfischer:
With all of that said, I'm probably okay just pulling this in and improving from there. Some is better than none.

joeljfischer:
Should be built against the new develop architecture.

joeljfischer:
Closed in favor of #96

PR #67, opened 2015-02-03
Feature/2.3.1

Mike-FoMoCo:
I've included the 2.3.1 release notes below:

**Updates from AppLink 2.3 to 2.3.1 (iOS)**

**Changes / Enhancements – iOS**
- Unregister method for transport (allows the transport to be called sooner increasing speed)
- Prevent unrelated accessories from causing a disconnect
- Better background connection on Gen1 head units.
- Better Multi-App experience on Gen1.1 head units

**Fixed Issues – iOS**

- Fixed “notifications weren’t setting their message type” bug.
- Fix for infinite loop when disconnected while waiting for connection. Call up one
level higher so we re-perform the search to see what external accessories are
connected.
- Changed the SDLProtocol to recursively call the processMessages method in
the event that more than 1 message is in the buffer at once.

**Note:**

- Developers should NOT retain a reference to the protocol/transport objects within your app. (SDLIAPTransport, SDLAppLinkProtocol) Doing so may cause connectivity issues.

Edit (Joel): Changed Ford names to SDL ones


justinjdickow:
@Mike-FoMoCo What are the changes for 2.3.1?

Mike-FoMoCo:
Sorry about that, I've included the 2.3.1 release notes below:

**Updates from AppLink 2.3 to 2.3.1 (iOS)**

**Changes / Enhancements – iOS**
- Unregister method for transport (allows the transport to be called sooner increasing speed)
- Prevent unrelated accessories from causing a disconnect
- Better background connection on Gen1 head units.
- Better Multi-App experience on Gen1.1 head units

**Fixed Issues – iOS**

- Fixed “notifications weren’t setting their message type” bug.
- Fix for infinite loop when disconnected while waiting for connection. Call up one
level higher so we re-perform the search to see what external accessories are
connected.
- Changed the AppLinkProtocol to recursively call the processMessages method in
the event that more than 1 message is in the buffer at once.

**Note:**

- Developers should NOT retain a reference to the protocol/transport objects within your app. (FMCIAPTransport, FMCAppLinkProtocol) Doing so may cause connectivity issues.

sdl_ios/SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -269,6 +279,12 @@ - (void)dealloc {
     [[NSNotificationCenter defaultCenter] removeObserver:self name:UIApplicationWillEnterForegroundNotification object:nil];
```
joeljfischer: All NSNotificationCenter observations can be removed from an observer by calling `[[NSNotificationCenter defaultCenter] removeObserver:self]`, which will clean up a few lines of code and make sure you don't miss removing any observations at dealloc time.

sdl_ios/SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -269,6 +279,12 @@ - (void)dealloc {
...
 }
 
+- (void)dealloc {
+    [self unregister];
+    [self closeSession];
```
joeljfischer: Close session is called twice, it's also called within `unregister`

sdl_ios/SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -295,16 +311,15 @@ - (void)openSession {
         } else {
             if ([self.protocolString isEqualToString:CONTROL_PROTOCOL_STRING]) {
```
joeljfischer: This if statement no longer does anything and could be removed

sdl_ios/SmartDeviceLink/SDLDataStreamHandlingDelegate.h
```diff
@@ -0,0 +1,14 @@
...
+//  SmartDeviceLink
+//
+
+#import <Foundation/Foundation.h>
+#import "SDLAbstractProtocol.h"
+
+@interface SDLDataStreamHandlingDelegate : NSObject <NSStreamDelegate>
+
+@property (assign) SDLServiceType serviceType;
+@property (strong, nonatomic) SDLAbstractProtocol *protocol;
```
joeljfischer: If this is strongly referenced by the proxy, then it could be weak in here, and just test to make sure it doesn't go to nil before using it.

sdl_ios/SmartDeviceLink/SDLProtocol.m
```diff
@@ -14,21 +15,22 @@
 #import "SDLRPCPayload.h"
 #import "SDLDebugTool.h"
 #import "SDLPrioritizedObjectCollection.h"
+#import "SDLDataStreamHandlingDelegate.h"
 
 
-const NSUInteger MAX_TRANSMISSION_SIZE = 512;
+const NSUInteger MAX_TRANSMISSION_SIZE = 1024;
```
joeljfischer: Why was this increased, why wasn't it increased before, is there a limit we should be aware of? Where does this number come from?

sdl_ios/SmartDeviceLink/SDLProtocol.m
```diff
@@ -237,22 +253,52 @@ - (void)processMessages {
...
     [self sendDataToTransport:message.data withPriority:header.serviceType];
+    
+}
 
+- (void)sendRawDataStream:(NSInputStream *)inputStream withServiceType:(SDLServiceType)serviceType {
+    
+    SDLDataStreamHandlingDelegate *streamDelegate = [SDLDataStreamHandlingDelegate new];
+    streamDelegate.serviceType = serviceType;
+    streamDelegate.protocol = self;
+    objc_setAssociatedObject(inputStream, @"RetainedDelegate", streamDelegate, OBJC_ASSOCIATION_RETAIN);
```
joeljfischer: This whole idea of using associated objects to retain the stream delegate onto the input stream provided by developers is simply asking for bugs and heartache. If the developer doesn't realize they have to retain the input stream they pass in themselves (which they shouldn't have to), then this stream delegate will immediately be released and die.

sdl_ios/SmartDeviceLink/SDLDataStreamHandlingDelegate.m
```diff
@@ -0,0 +1,53 @@
...
+        case NSStreamEventHasBytesAvailable:
+        {
+            // Grab some bytes from the stream
+            const int bufferSize = 1024;
+            uint8_t buf[bufferSize];
+            NSUInteger len = [(NSInputStream *)stream read:buf maxLength:bufferSize];
+            if(len > 0)
+            {
+                NSData* data = [NSData dataWithBytes:buf length:len];
+                [self.protocol sendRawData:data withServiceType:self.serviceType];
```
joeljfischer: I don't understand why we want developer to pass in an input stream if all we're going to do is read and send the raw data. They have the full file, why don't they just pass it to us, a file handle for the file, or the raw data itself?

sdl_ios/SmartDeviceLink/SDLProtocol.m
```diff
@@ -55,6 +57,20 @@ - (id)init {
 	return self;
 }
 
+- (void)storeSessionID:(UInt8)sessionID forServiceType:(SDLServiceType)serviceType {
+    [_sessionIDs setObject:[NSNumber numberWithUnsignedChar:sessionID] forKey:[NSNumber numberWithUnsignedChar:serviceType]];
+}
+
+- (UInt8)retrieveSessionIDforServiceType:(SDLServiceType)serviceType {
+    
+    NSNumber *number = [_sessionIDs objectForKey:[NSNumber numberWithUnsignedChar:serviceType]];
```
joeljfischer: This could be `NSNumber *number = _sessionIDs[@(serviceType)];`

joeljfischer:
I would like for these to be addressed. In addition to these, the fact that this PR makes api removals means that this is a major version change, so the earliest this could be merged is 4.0. Before it is merged, it would also need to be rebased on develop and squashed.

sdl_ios/SmartDeviceLink/SDLIAPTransport.m
Heath-FoMoCo: Can you please add more context to this statement?
Which `if statement`?  Why do you feel it is no longer needed?

Heath-FoMoCo:
@joeljfischer Which APIs changed for this pull request?  I am a bit confused; can you get more specific?

sdl_ios/SmartDeviceLink/SDLIAPTransport.m
joeljfischer: Sure, the lines

```objc
if ([self.protocolString isEqualToString:CONTROL_PROTOCOL_STRING]) {
                [SDLDebugTool logInfo:@"Session Not Opened (Control Protocol)" withType:SDLDebugType_Transport_iAP toOutput:SDLDebugOutput_All toGroup:self.debugConsoleGroupName];
            } else {
                [SDLDebugTool logInfo:@"Session Not Opened" withType:SDLDebugType_Transport_iAP toOutput:SDLDebugOutput_All toGroup:self.debugConsoleGroupName];
            }
```

used to contain the connection retry. Now they only log. If the logs are still necessary, then they can remain, but if not, then they could be removed since no active code is being run within them.

joeljfischer:
Sure. This is the list I got.

SemVer major version changes:
* `SDLAbstractProtocol` public protocol conformance changed
* `SDLAbstractProtocol` method `[sendEndSessionWithType:sessionID:]` became `[sendEndSessionWithType:]`
* `SDLProtocol` method `[sendEndSessionWithType:sessionID:]` became `[sendEndSessionWithType:]`
* `SDLProtocol` method `[handleBytesFromTransport]` removed
* `SDLProxy` iVar `rpcSessionID` removed
* `SDLProxy` iVar `bulkSessionID` removed
* `SDLProxy` property `protocol` type changed
* `SDLProxy` method `[initWithTransport:protocol:delegate:]` input variable type changed
* `SDLProxy` method `[getProtocol]` removed
* `SDLProxy` method `[putFileStream::]` removed (was deprecated)

SemVer Minor version changes
* `SDLRPCNotification` `[init]` methods added
* `SDLTransport` `[unregister]` method added

sdl_ios/SmartDeviceLink/SDLIAPTransport.m
Heath-FoMoCo: Thanks.
The SDLDebugTool also makes data visible to us VIA Siphon.  We need this information since multiapp/background connection is so finicky we have to have these debug lines to ensure unexpected behavior is normal.

Heath-FoMoCo:
Thanks.  We are reviewing these changes.

sdl_ios/SmartDeviceLink/SDLRPCNotification.m
```diff
@@ -1,9 +1,28 @@
...
+}
+
+- (id)initWithDictionary:(NSMutableDictionary*) dict
+{
+    if (self = [super initWithDictionary:dict])
+    {
+        messageType = NAMES_notification;
+    }
+    return self;
+}
```
jacobkeeler: Some things to note about this fix:
1. The incorrectly set `messageType` issue does not just affect Notifications, but Responses as well
2. Unfortunately, what you currently have will not quite fix the issue. The reason for this is that while the `messageType` variable **will** be set correctly, this value is used in creating the internal dictionary, and thus the dictionary will be improperly structured.
```objc
-(id) initWithName:(NSString*) name {
	if (self = [super init]) {
		function = [[NSMutableDictionary alloc] initWithCapacity:3];
		parameters = [[NSMutableDictionary alloc] init];
		messageType = NAMES_request;
		[store setObject:function forKey:messageType];
		[function setObject:parameters forKey:NAMES_parameters];
		[function setObject:name forKey:NAMES_operation_name];
	}
	return self;
}
```
As you can see, the `messageType` is used as a key in the dictionary, and this key-value pair is set while the `messageType` is still "request".
```objc
messageType = NAMES_request;
[store setObject:function forKey:messageType];
```
The simplest way to fix this issue would be to remove these two lines from [SDLRPCMessage initWithDictionary] and add them to the override of initWithDictionary in SDLRPCNotification, SDLRPCResponse, SDLRPCRequest, replacing `names_request` with `names_notification`, `names_response`, and `names_request` respectively.

joeljfischer:
Note that #71 could alleviate the issue of API changes that we are facing. Looking for feedback regarding that proposal.

sdl_ios/SmartDeviceLink/SDLProtocol.m
justinjdickow: If an associated object _is_ to be used, it should be a category on NSInputStream and the key should be a selector of the getter in order to guarantee its uniqueness. See the [NSHipster article of Associated Objects](http://nshipster.com/associated-objects/) for more information. Also please note the last two paragraphs..

>Associated objects should be seen as a method of last resort, rather than a solution in search of a problem (and really, categories themselves really shouldn't be at the top of the toolchain to begin with).

>Like any clever trick, hack, or workaround, there is a natural tendency for one to actively seek out occasions to use it—especially just after learning about it. Do your best to understand and appreciate when it's the right solution, and save yourself the embarrassment of being scornfully asked "why in the name of $DEITY" you decided to go with that solution.

joeljfischer:
Closed in favor of #91

PR #57, opened 2015-01-27
Cocoapods distribution

joeljfischer:
:bowtie: Ready to merge :bowtie: 

Closes #5 & #6 

## Summary
This branch is intended to alter distribution methods for this repository.

1. Build a iOS 8+ [dynamic framework (video)](https://developer.apple.com/videos/wwdc/2014/#416)
2. Add the ability to run tests on that framework
3. Add an example app that embeds the framework target to the same project
4. Write a podspec and fix errors currently preventing a [Cocoapods](http://cocoapods.org) build from working
5. Have the ability to use [Carthage](https://github.com/Carthage/Carthage) for distribution

See [this presentation](https://docs.google.com/presentation/d/1Si4PhqcnATXts2HC7kTPN3G-EXdBCrm9PPiZ1x87v54/edit?usp=sharing) for an overview.

##### Disadvantages
~~This would remove the ability to build a "fake framework" target that we currently use for builds. This means that the only way to continue to get iOS 7 support is through Cocoapods (or manually adding the Library directory to their project). We could add an additional target to continue the fake framework support, if we wished, or stick with Cocoapods as the only iOS 7 support.~~

Dynamic frameworks provide iOS 7 support! Weakly linking them means that despite a warning, everything should work fine in iOS 7 as long as the framework also targets iOS 7+.

### Build an iOS 7+ framework
iOS moves pretty quickly; dynamic frameworks are worlds better than static libraries and fake frameworks. This means that we should be working towards an implementation that can build proper dynamic frameworks. They're easy to import, and must be supported by cocoapods' current beta release.

### Testing
Dynamic framework targets come with a test target automatically. This makes it much easier to implement tests on the library code. It also opens up continuous integration as a possibility.

### Example App
Dynamic framework targets come with the ability to be embedded in app targets within the same project. This will make shipping an example app about a million times easier.

### CocoaPods
CocoaPods is by far the largest / most comprehensive resource for importing open-source projects into Xcode. We really need to support it. Unfortunately, the current import setup does not work for either frameworks or cocoapods (which will be building frameworks when it can in their next release). This PR supplies a podspec which will need to be verified.

Verifying a podspec is usually pretty easy. You commit, tag, and run a `lint` command line command to test your repo is setup properly. If it isn't, you commit and tweak until it works. That cannot be done on this repository, unfortunately. We will have to create a test repository that we can tweak, figure out what works, and move that work back here.

[...]

Issue #85, opened 2015-02-17
Remove Siphon Server


joeljfischer:
Major version change

The siphon server cannot currently be used by developers. It requires Ford-proprietary, closed-source, Windows-only software right now. Alternatives should be discussed.

idamaster:
Perhaps we can open source the Siphon server software, add it here, and put together a roadmap/timeline for its development/purpose/etc.

Heath-FoMoCo:
Open sourcing the siphon software it not an option.  Siphon was develop with the purpose of Ford use only and making changes to be open source would not be beneficial.

joeljfischer:
Honestly, in addition to that, I do believe there are better options available. My preferred one that I would like to investigate is using an open-source HTTP server to serve a webpage from the device to the local network, which could then be viewed from any web browser that [supports sockets](http://caniuse.com/#feat=websockets). We're currently awaiting feedback from the lawyers regarding the use of MIT & Apache 2.0 software in our open-source projects for this and other purposes.

This would likely be implemented as another weakly linked framework, the main framework would look for it at runtime and use it if available, and no-op if it's not there. This would also allow developers to automatically use the framework in debug builds only, which is a vast improvement.

joeljfischer:
@Heath-FoMoCo is this still in use? Do you think that Relay supersedes it? It seems like Relay and this attempt to accomplish the same thing, but Relay does it better.

asm09fsu:
@joeljfischer so NSLog (when compiled with iOS 10) actually uses the new Unified Logging system. This means that for now, we do not have to do special modifications to handle that, however if we'd like to improve the ability to filter, we can investigate creating a [custom log object](https://developer.apple.com/reference/kernel/1643798-os_log_create?language=objc)

PR #87, opened 2015-02-23
Remove Siphon Server

joeljfischer:
Fixes #57, this removes the siphon server and all associated APIs. See #85 for more information.


Heath-FoMoCo:
Before we can remove Siphon we have to have an equivalent coded.
Without we would be debugging iOS apps blindly.

joeljfischer:
I understand that, but I was under the impression that SmartDeviceLink apps cannot be debugged through the Siphon Server as it exists, and that this is a Ford only feature.

I'm in total agreement that getting a replacement out there is a top priority, but since this is a major version change, I would say it is better to remove it while we have other major changes batched here, and push for the new system (a minor version change) as soon as possible, whether it be before 4.0 is out or shortly after.

Am I off base here?

Heath-FoMoCo:
It is a Ford feature used to debug partner apps.
When a partner has issues we use this tool to determine if it is app logic or module behavior causing the issue.  Removing this would cripple this support .

mikeburke106:
Pardon my ignorance on the iOS libs, but I believe (in the Android code, at least) if the SiphonServer isn't connected, the logs will just be displayed via LogCat or whatever native log mechanism exists.

From SdlTrace.java:

```java
// Attempt to write formatted message to the Siphon
if (false == writeMessageToSiphonServer(msg)) {
	// If writing to the Siphon fails, write to the native log
	NativeLogTool.logInfo(SdlTrace.SYSTEM_LOG_TAG, msg);
}
```

Is there nothing like this set up in iOS yet, or is this not sufficient for our debugging purposes?

Heath-FoMoCo:
Correct; it is not sufficient.
On Android you can plug into the debugger and get live logging.  This is not possible with iOS plugged into a module.
With iOS you have to do “cable swap” to get logs (module to Mac) and there is always a chance that you will miss the entries.  Painful and not robust enough.

What we need is a way to debug with the cable plugged in the module.   Live code debugging, (Bluetooth, Wi-Fi, magic, anything) with breakpoints, would be awesome.  Right now, we believe, iOS makes this impossible.

mikeburke106:
Ohhhhh ok, I see.  So, the real cause of the problem is that iOS devices can only connect to SDL via USB and it can't write logs to USB at the same time.  I would agree, then, that this can't be removed until we have an alternative implemented.

joeljfischer:
I'm working on a replacement right now, this won't be merged in until that's done. It should be ready for 4.0.

PR #89, opened 2015-02-23
Fix SDLHMIZoneCapabilities class case

joeljfischer:
Fixes #54, based on the new distribution branch #57, can be merged after that one is.


joeljfischer:
Recreate against new develop branch

Issue #64, opened 2015-01-30
If an SDLRPCMessage object is initialized using initWithName, the messageType is always set to "request"


jacobkeeler:
Regardless of the type of message, if an `SDLRPCMessage` object is initialize using `initWithName`, then the `messageType` is automatically set to "request".
```
messageType = NAMES_request;
```
This should instead set the `messageType` to whatever is appropriate based on the message.

A simple way of doing this would to override the `initWithName` function in `SDLRPCRequest`, `SDLRPCNotification`, and `SDKRPCResponse`, and set the `messageType` in that override instead.

For example, the override in `RPCNotification` might look like:
```
-(id) initWithName:(NSString*) name {
	if (self = [super initWithName:name]) {
		messageType = NAMES_notification;
		[store setObject:function forKey:messageType];
	}
	return self;
}
```

joeljfischer:
This should be checked to see if it affects #45

Issue #34, opened 2015-01-20
Ability to send any type of RPCMessage subclass


joeljfischer:
Foreground app tracking requires the ability to send RPC Notifications to the head unit, we may as well include the ability to send any type of message. This issue will be resolved as part of the PR [#33](https://github.com/smartdevicelink/sdl_ios/pull/33).

PR #99, opened 2015-03-09
App Launching for 4.0

joeljfischer:
Based on #45, refactored for new project structure

:cat: Not Ready to Merge :cat: 

## Summary
This PR will involve several pieces to allow an iOS application to track available applications on the users phone that can be launched with SmartDeviceLink. This will involve 4 major pieces.

1. Head Unit tracking of the foregrounded application on a user's phone.
2. Head Unit requested querying of a server URL to retrieve a list of applications that are SDL compatible.
3. Parsing the response from the server, and determining which apps installed on the user's phone can be launched, and sending the filtered response to the HU.
4. Responding to a request from the HU to launch a specific app, and launching it.

### Tracking Foregrounded Application
The SDL Proxy object, upon opening a connection, will poll the application for its state, and send that across. It will also immediately set up observers on application state changes. Upon firing of those observers, it will recheck the application for its state, and push a new notification to the HU.

This PR also introduces the ability to send notifications from the Mobile API (issue #34). This is required because application state changes are sent via a pre-existing RPC Notification that was previously only sent from the HU, onHMIStatus. FULL is equivalent to foreground, LIMITED is equivalent to background.

### Query Apps Server HTTP Request
The HU will request, via a OnSystemRequest, that the mobile device retrieve an application list from the server. The application *may* (this has not yet been determined) cache the response for a specified period.

### Query Apps Filter Installed Applications
The server will respond with a set response, [described here](https://github.com/smartdevicelink/sdl_server/blob/master/docs/application_launching_v1.0.md) which will be parsed by the library. The library will then run a system call to determine which apps are installed on the phone. The library will then filter the list passed by the server so that only installed apps are represented, and forward it on to the HU.

### Launch App
The HU will send an OnSystemRequest with a URL field that contains the URL scheme of the app they wish to launch to the foregrounded application. The library running in that application will then attempt to launch the application specified by the url scheme passed in. This will, of course, foreground that application, and background the application making the call.

##### General
- [x] Make sure the proxy can handle being at v4
- [x] Refactor SDLRPCNotification, SDLRPCMessage, etc., so that it sets it's messageType correctly. ~~See PR #67~~ See #64 

##### Tracking Foreground State
- [x] Track Application State
- [x] Add the ability to send RPCNotifications (by sending any type of RPCMessage)
- [x] Send an OnHMIStatus notification to connected HU when application state changes
- [x] Remove app state observers when session closes
- [x] Properly log and handle errors

##### Query Apps Server HTTP Request
- [x] Create new SystemRequest enum type for Query Apps
- [x] Capture OnSystemRequests for Query Apps and Launch App commands
- [x] Make a query apps web call to the server specified in the OnSystemRequest to retrieve a list of URLs
- [x] Parse the response from the server, determine applications to be checked
- [x] Properly log and handle errors
- [x] Determine if NSURLSessionUploadTask should be NSURLSessionDataTask

##### Query App Filter Installed Applications
- [x] Test each application to see if it's installed on the phone
- [x] Filter the response from the server based on which apps are installed on the phone
- [x] Send the filtered response to the Head Unit
- [x] Properly log and handle errors

##### Launch App
- [x] Create new SystemRequest enum type for Launch App
- [x] Parse a Launch App OnSystemRequest
- [x] Open the specified application from a Launch App OnSystemRequest
- [x] Properly log and handle errors

##### Other
- [x] Test changes to query apps did not adversely affect Policy Server requests
- [x] Refactor into new develop project layout


joeljfischer:
This fixes #64

SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
```diff
@@ -41,108 +43,124 @@ - (void)logRPCSend:(SDLProtocolMessage *)message;
...
-	if (self = [super init]) {
+    if (self = [super init]) {
         _version = 1;
         _messageID = 0;
         _sessionID = 0;
         _recieveQueue = dispatch_queue_create("com.sdl.recieve", DISPATCH_QUEUE_SERIAL);
         _sendQueue = dispatch_queue_create("com.sdl.send.defaultpriority", DISPATCH_QUEUE_SERIAL);
         prioritizedCollection = [SDLPrioritizedObjectCollection new];
-
+
```
hugh22: Please get rid of these.

SmartDeviceLink-iOS/SmartDeviceLink/SDLInterfaceProtocol.h
```diff
@@ -17,6 +17,7 @@
 - (void)handleBytesFromTransport:(NSData *)receivedData;
 - (void)sendStartSessionWithType:(SDLServiceType)sessionType;
 - (void)sendEndSessionWithType:(SDLServiceType)sessionType sessionID:(Byte)sessionID;
-- (void)sendRPCRequest:(SDLRPCRequest *)rpcRequest;
+- (void)sendRPCRequest:(SDLRPCRequest *)rpcRequest __deprecated_msg("use -sendRPC: instead");
```
hugh22: Why the name and type change?

SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocolHeader.m
```diff
@@ -38,20 +38,27 @@ - (NSString *)description {
...
+        case 2:
+        case 3:
+        case 4: {
+            return [[SDLV2ProtocolHeader alloc] initWithVersion:version];
+        } break;
+        default: {
+            return [[SDLV2ProtocolHeader alloc] initWithVersion:2];
+        } break;
+    }
 }
```
hugh22: Why do you have to init a version specific object WITH a version number? These methods probably shouldn't even exist. And it might be better to have any invalid version numbers (case 0 and default) not just continue happily.

SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
```diff
@@ -41,108 +43,124 @@ - (void)logRPCSend:(SDLProtocolMessage *)message;
...
+    // Build the message payload. Include the binary header if necessary
+    switch (self.version) {
+        case 1: {
+            // No binary header in version 1
+            messagePayload = jsonData;
+        } break;
+        case 2: // Fallthrough
+        case 3:
+        case 4:
+        default: {
```
hugh22: If no case is matched, probably should be treated as an error.

SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocolHeader.m
joeljfischer: Because we are considering the protocol header version as a separate version. So it's a ProtocolHeader version 2, since in version 2, 3, 4, the header is all the same, but we need a specific version number for the protocol header itself. This is to avoid creating classes that mirror each other with hardcoded version values as the only difference.

joeljfischer: Regarding version 0, I believe there was a case where it might be sent, I'll have to ask @justinjdickow to see if he remembers why we left that. It may have been just because we weren't sure, and the old code did the same. The default case was possibly the same reason. We can reevaluate that, because I'd definitely prefer to fail faster than to leave corner cases.

SmartDeviceLink-iOS/SmartDeviceLink/SDLInterfaceProtocol.h
joeljfischer: Because we needed the ability to send RPC Notifications from the app side for foreground / background notifications, which piggyback onHMIStatus.

SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
joeljfischer: If I had to guess, I would say they were automatic Xcode changes in response to spacing inconsistencies? I can't really tell what happened here.

SmartDeviceLink-iOS/SmartDeviceLink/SDLProxy.m
```diff
@@ -173,232 +204,404 @@ - (void)handleProtocolSessionStarted:(SDLServiceType)sessionType sessionID:(Byte
...
+    
+    // Get data dictionary from the bulkData
+    NSError *error = nil;
+    NSDictionary *JSONDicationary = [NSJSONSerialization JSONObjectWithData:request.bulkData options:kNilOptions error:&error];
+    if (error != nil) {
+        [SDLDebugTool logInfo:@"OnSystemRequest failure: notification data is not valid JSON." withType:SDLDebugType_RPC toOutput:SDLDebugOutput_All toGroup:self.debugConsoleGroupName];
+        return nil;
+    }
+    
+    return JSONDicationary;
```
joeljfischer: Misspelling

SmartDeviceLink-iOS/SmartDeviceLink/SDLRequestType.h
```diff
@@ -7,14 +7,16 @@
...
-+(SDLRequestType*) AUTH_ACK;
-+(SDLRequestType*) PROPRIETARY;
++ (SDLRequestType*)HTTP;
++ (SDLRequestType*)FILE_RESUME;
++ (SDLRequestType*)AUTH_REQUEST;
++ (SDLRequestType*)AUTH_CHALLENGE;
++ (SDLRequestType*)AUTH_ACK;
++ (SDLRequestType*)PROPRIETARY;
++ (SDLRequestType *)QUERY_APPS;
++ (SDLRequestType *)LAUNCH_APP;
```
joeljfischer: I don't know why these are inconsistent, but they should all be consistent.

joeljfischer:
@hugh22 I merged in the current develop code to this branch, which appears to make following along with the code changes much easier. I'm pretty confident that the merge went well, but there were a number of errors to be resolved so just keep that in mind as well.

SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocolHeader.m
joeljfischer: @hugh22 I'm testing again, and I'm seeing a version incoming that is 0. It seemed to be coming via attempting to parse a zero byte 'receivedData'. I traced this to a TCP bug, which I fixed.

joeljfischer:
@hugh22 If you could take another look and see if your issues are resolved, that would be great!

hugh22:
Looks good to me.

PR #88, opened 2015-02-23
Remove ISDLProxy protocol

joeljfischer:
Fixes #70, fixes #19. This is based on the new distribution branch #57 and can be merged after that one is

PR #104, opened 2015-03-10
Remove isdlproxy

joeljfischer:
#88 but for the new project structure. This simply removes a completely unused protocol and it's accompanying file.

Note: This will actually be merged to develop.


justinjdickow:
Needs to be reopened on develop

PR #112, opened 2015-03-11
SDLEnums now store values in an immutable NSArray

### LINKED ISSUE #42: SDLEnum types should return an immutable array for +[values]
Currently they return mutable arrays, which means that anybody could accidentally modify the array, which could easily result in bugs.

joeljfischer:
Fixes #42


joeljfischer:
@smartdevicelink/ios I really need someone else to look at this. I don't know if there was something I'm missing about why it ought to be a mutable array.

PR #116, opened 2015-03-12
Give SDLDebugToolConsole protocol its own file

### LINKED ISSUE #114: SDLDebugToolConsole should have it's own file
All protocols should have their own files so they can be imported into headers separately from the classes they are attached to.

joeljfischer:
Fixes #114


justinjdickow:
Remove copyright from top of new files

PR #120, opened 2015-03-17
Tome

tomhoag:
Changed SmartDeviceLink scheme to run app
changed button.textLabel.setTitle --> [button setTitle:forState:]


joeljfischer:
Hey thanks for the PR. However, there are multiple schemes in there, including one that already runs the example project. The SmartDeviceLink scheme is purely for building the dynamic framework.

In addition, please separate differing features into different pull requests. I cannot pull in this pull request, because although the button state change is good, the scheme change is not.

joeljfischer:
In addition, we are still working on getting our CLA in place, so we unfortunately cannot accept PRs from outside of Ford until that is cleared up.

PR #124, opened 2015-03-18
Create a clang-format file and format everything

### LINKED ISSUE #107: Use clang-format to format all code according to style guide
Style guide has to be created first, but we can then use clang-format to format all commits.

http://clang.llvm.org/docs/ClangFormat.html

joeljfischer:
Fixes #107


joeljfischer:
This will be closed in favor of only having the clang-format file for now, the actual format will happen right before the release.

PR #153, opened 2015-03-31
Add SDLDialNumber RPC

joeljfischer:
- [x] Response Spec


joeljfischer:
This still needs a response spec

PR #162, opened 2015-04-02
Add new RequestType enum values

joeljfischer:
Add new RequestType enum value specs


joeljfischer:
Closed in favor of #163

PR #187, opened 2015-05-01
IAP Connection Refactor

joeljfischer:
Transfer of PR #105 to internal branch.

Updates based on AppLink 2.3.2 (iOS) Changes / Enhancements – iOS

* Significant internal changes to the iAP transport layer for stability and reliability especially in situations where multiple AppLink apps are running concurrently and for connection/disconnection events.
* The  SDLSyncProxy getters getProxyVersion, getTransport and getProtocol have been removed and replaced with properties.
* Introduced a delay before creating the EASession object upon connection to the module via USB cable in order to reduce collisions with other apps that may also be trying to connect at the same time. Delay will be from 0 to 10 seconds.
* The SDLProxy method putFileStream: : has a more descriptive signature now -
(void)putFileStream:(NSInputStream *)inputStream withRequest:( SDLPutFile *)putFileRPCRequest;
Debug tool now logs thread index as part of message instead of logical layer. Fixed Issues – iOS
* Fixed bug with AudioPassThru on Sync Gen 1.1 modules.
* Fixed a typo in the SDLProxyListener protocol. The parameter to onOnLockScreenNotification is now the correct type.
* Set build settings "DEPLOYMENT_POSTPROCESSING" to YES. This suppresses 400+ warnings when an app wants to build using dSYM.
* A/V Streaming additions


idamaster:
Hi @joeljfischer -- what was left on this PR? I think there was still an issue with renaming that needed to occur? Anything else? We wanted to get this in so we can start the remainder of PR's with @hugh22's other fixes.

joeljfischer:
I haven't been working on it recently because I'm trying to get 4.0 out the door and all my attention has been on trying to bugfix the app launching branch. There's still a few more things I need to look at on this branch, and do merge. It may take a few days before I'm able to get back to it.

joeljfischer:
Couple of quick questions to @hugh22 

`SDLIAPTransport disconnect` method only disconnects the data session, not the control session, is this intentional?

`SDLIAPTransport accessoryDisconnected` doesn’t do anything if control session disconnects, is this intentional?

If those states are intentional, I'm just looking for a reason to put into comments in those methods so we know why those don't deal with the control session.

Thanks!

hugh22:
Yep, this is on purpose. The control session doesn't remain open and so doesn't need to get explicitly disconnected. It gets closed either when it succeeds and has obtained the protocol index or when it fails and has timed out. We don't technically consider ourselves "connected" until we're open on the data session.

joeljfischer:
:+1:

joeljfischer:
@smartdevicelink/ios Just finished merging develop into this PR, so it should be ready to merge the other way. There were a *lot* of conflicts. It compiles, and I've done basic connection / show testing on core and a TDK 3. Any further testing you guys could do would be :tada:

Issue #210, opened 2015-05-28
The handleRPCDictionary method in SDLProxy.m uses reflection at the wrong time


adein:
This method was refactored in the latest merge to develop.  In the new code, the creation of the 'newMessage' variable is performed before 'Response' is determined/appended.  This results in all of the RPC response callbacks receiving a 'RPCRequest' based object instead of 'RPCResponse' based object.

e.g., onGetVehicleDataResponse receives an object of the type SDLGetVehicleData (instead of SDLGetVehicleDataResponse).

adein:
I'm not sure, but I believe this was also the reason that the app I was testing didn't think that SYNC3 had graphics support, even though it does.  Once this is fixed and merged, I'll test and confirm.

adein:
@joeljfischer I found out why I wasn't getting the graphics support in the RAI Response callback: There are returns inside of if-statements in the method for several specific RPCs (RAI, SystemRequest, etc).  The return statements prevent the RPC callback from being invoked near the end of the method, which results in the app not being notified of the RPC response.

joeljfischer:
You're right, those shouldn't have been in there.

joeljfischer:
#213 was merged into develop, see if that fixes your issue.

adein:
I believe everything seems to be working again, but more testing is needed since this is such a critical part of the code.

joeljfischer:
Okay, see #235

PR #213, opened 2015-05-28
Fix issue #210

joeljfischer:
Append 'response' to constructed RPC object before sending it to function handlers in `[SDLProxy handleRPCDictionary:]`


joeljfischer:
@adein This should fix #210

adein:
@joeljfischer Looks good.  I'll merge this in to a local branch and give it a try.

joeljfischer:
Removed the errant return statements

PR #214, opened 2015-05-29
add the AV session methods

hugh22:
AV methods

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLProxy.h
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLProxy.h
@@ -46,6 +46,15 @@
 + (void)enableSiphonDebug;
 + (void)disableSiphonDebug;
 
+- (void)startAudioSession;
+- (void)sendAudioData:(NSData *)data;
+- (void)stopAudioSession;
+
+- (void)startVideoSession;
+- (void)sendVideoData:(NSData *)data;
+- (void)stopVideoSession;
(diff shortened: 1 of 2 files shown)


joeljfischer:
As we discussed, can we remove the public access to these methods? Unless necessary, I'd rather not implement support for these broad methods without a very good reason to do so. It may require a restructure along with #220, to encapsulate video & audio data transfer better.

hugh22:
You want to remove access to starting and stopping the sessions? That's interesting. I thought you were talking about removing the access to sending hardware encoded data. Also, just because we "discussed it" doesn't mean the idea was approved.

joeljfischer:
Well, I think that it could be easier to use if it were encapsulated within a single, purpose-built class. Abstracting away "sessions" and raw data transfer into a class that is a bit more semantic for what you're doing and can track if a session is active, for example, seems to be a better structure. Thoughts about that?

hugh22:
Guess it just depends on if you want to give users flexibility or limit & control them. I tend to lean towards the former.

joeljfischer:
It's not my intention to limit & control. My thought is to give them tools that should make their development easier by building on the raw blocks. For instance, the class can warn you that data will not be sent if you didn't open a session, and can include the touch events responses in the same class. Basically creating one entry point. I don't think that's limitation and control. I just want to give users tools to do the right thing easily.

joeljfischer:
Closed for #286

PR #220, opened 2015-06-01
Feature/video encoding accept hw output

hugh22:
Add method to process CMSampleBufferRef produced by hardware encoder.


joeljfischer:
Any thoughts on moving the hardware decoder out of the proxy class into its own class? It would allow for better separation of concerns and better flexibility for future changes. I'm not sure the best way to structure it to send data back to the head unit, though.

joeljfischer:
Closed in favor of #286

PR #222, opened 2015-06-01
allow variable MTU

hugh22:
let MTU be specified at runtime


hugh22:
Nevermind, apparently this is not supposed to go live.

PR #236, opened 2015-06-19
SDLProxy start transport on initialization

### LINKED ISSUE #211: Connecting the phone to SYNC before launching the app results in the app not attempting to connect
It seems that the proxy doesn't check for EA connections when the proxy is instantiated; it only checks when a connection event occurs.  This results in apps not connecting if the app is launched after the device is connected to SYNC.

joeljfischer:
* Start up transport using a simple `connect` method call instead of adding the call to the end of the run loop.

* Fixes #211


joeljfischer:
@adein Updated to only contain the changes needed to fix #211

Issue #251, opened 2015-07-09
sendEndSesssionWithType:(SDLServiceType)serviceType method not defined in SDLProtocol.m class 


anildahiya:
Unfortunately during merging in develop branch i.e. commit SHA1 id : e6c2b6c7bc592c3154895936b34ebac7437f548d

following method definition got erroneously updated : 
-- (void)sendEndSessionWithType:(SDLServiceType)serviceType {
+- (void)sendEndSessionWithType:(SDLServiceType)serviceType sessionID:(Byte)sessionID {

i.e. sendEndSesssionWithType:(SDLServiceType)serviceType method got changed to  sendEndSesssionWithType:(SDLServiceType)serviceType sessionID:(Bytes)sessionID in SDLProtocol.m class.

It's leading to a crash since the original method sendEndSesssionWithType:(SDLServiceType)serviceType method is no longer found.

PR #252, opened 2015-07-09
potential fix for avoiding crash in session for av.

anildahiya:
Changes for fixing issue #251  with help of commit sha1 id : 00bd37e9


joeljfischer:
FYI, you don't have to put the SHA of your change, it shows the commits and files changed in tabs on PRs.

It would also be helpful to reference the issue you are fixing. #251 in this case.

joeljfischer:
@anildahiya Since there is already a release branch, and this PR is targeting that release, this PR should target the release/4.0.0 branch, not the develop branch.

joeljfischer:
@anildahiya @adein Since this was a blocker for #286, it is fixed in that PR.

PR #272, opened 2015-08-05
SDLGlobals for global version & MTU size handling

### LINKED ISSUE #256: MTU Size for Protocol Version 3
Please see: https://github.com/smartdevicelink/protocol_spec/issues/4

joeljfischer:
:cat: Ready for Review :cat:

#### Summary
This creates a new singleton class that has as project (not public level) global instance that stores the connected protocol version and can currently also be used to retrieve the maxMTU size based on that version.

#### Detail
This separates out version handling into a separate class called SDLGlobals. It is a singleton class that can be accessed from anywhere in the code. It defaults to a protocol version of 1. When a head unit version is set, it will automatically calculate the protocol version and the max MTU size based on that.

- [x] Needs to be updated based on new MTU discussions
- [x] Needs tests

Fixes #256

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLGlobals.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLGlobals.m
@@ -37,8 +37,6 @@ static const NSUInteger maxProxyVersion = 4;
     }
     
     _protocolVersion = 1;
-    
-    // This will cause maxMTUSize to assert unless changed before maxMTUSize is called. This is by design.
     _maxHeadUnitVersion = 0;
     
     return self;
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLGlobals.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLGlobals.m
@@ -8,6 +8,16 @@
 
 #import "SDLGlobals.h"
 
+static const NSUInteger maxProxyVersion = 4;
+
+
+@interface SDLGlobals ()
+
+@property (assign, nonatomic) NSUInteger protocolVersion;
+
+@end
+
+
 @implementation SDLGlobals
 
 + (instancetype)globals {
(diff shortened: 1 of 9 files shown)


Heath-FoMoCo:
@joeljfischer Are you developing the "future proofing" hard coded MTU?  If not I will assign someone.

joeljfischer:
Yeah, I'll just modify this current PR with it in the next few days.

justinjdickow:
@adein this looks good to me - just waiting for you to take a look for approval to merge

@joeljfischer

SmartDeviceLink-iOS/SmartDeviceLink/SDLGlobals.m
```diff
@@ -0,0 +1,77 @@
...
+        case 4: {
+            // If the head unit isn't running v3/4, but that's the connection scheme we're using, then we have to know that they could be running an MTU that's not 128k, so we default back to the v1/2 MTU for safety.
+            if (self.maxHeadUnitVersion > maxProxyVersion) {
+                return 1024;
+            } else {
+                return 128000;
+            }
+        } break;
+        default: {
+            NSAssert(NO, @"Unknown version number: %@", @(self.protocolVersion));
```
adein: Do our build settings specify that assertions are removed from Release builds?  If they are not removed, do we really want users to experience an app crash from an assert if a developer fails to update to a new proxy?  If asserts are removed, would returning 0 as the MTU size cause a different crash due to buffers being allocated to 0 bytes long?  It looks like it would be ok at first glance, but if we could test with a higher protocol number to make sure we're appropriately 'future-proofed', I'd be happier.

joeljfischer: This wouldn't crash if a developer failed to update. This uses the negotiated protocol version, which ought to always be <= maxProxyVersion. The assert is mainly so that if we update the proxy to, say, v5 and are testing our changes, then we will assert, and will know we need to update our code in that place. 

As for in release...I'm undecided. At one point I did have it throw an exception which would then force a crash even in release. I didn't necessarily want that to play out though, but at the same time, if this code actually reaches this point we're pretty much screwed. If we return 0 (or NSNotFound or something), perhaps we could have the caller realize something went wrong and shut down the whole proxy?

adein: Yeah.  If there's a simple way to make it just abort the whole process in that scenario, that would be best IMO.

joeljfischer: Okay. I think that's something that can be addressed in a 4.0.X update. It shouldn't be critical, because it is impossible for this code to run right now with line 49 as it is.

adein:
:thumbsup:  Looks good to me.

PR #273, opened 2015-08-06
ChangeRegistration RPC new parameters for 4.0

joeljfischer:
* Added AppName, TTSName, NGNMediaScreenAppName, and VRSynonyms

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLChangeRegistration.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLChangeRegistration.m
@@ -57,50 +57,50 @@
 
 - (void)setAppName:(NSString *)appName {
     if (appName != nil) {
-        parameters[NAMES_appName] = appName;
+        parameters[NAMES_appName] = [appName copy];
     } else {
         [parameters removeObjectForKey:NAMES_appName];
     }
 }
 
 - (NSString *)appName {
-    return parameters[NAMES_appName];
+    return [parameters[NAMES_appName] copy];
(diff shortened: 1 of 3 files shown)


joeljfischer:
@adein @justinjdickow @asm09fsu This still needs to be reviewed.

SmartDeviceLink-iOS/SmartDeviceLink/SDLChangeRegistration.m
```diff
@@ -55,4 +55,52 @@ - (SDLLanguage *)hmiDisplayLanguage {
     }
 }
 
+- (void)setAppName:(NSString *)appName {
+    if (appName != nil) {
+        parameters[NAMES_appName] = appName;
```
adein: @joeljfischer Since you're intending the 'copy' property type above, but then we specify a custom setter method, shouldn't the setter method call copy on the object passed in before storing it?  Same thing for all of the copy properties.

joeljfischer: yup you're right

adein:
@joeljfischer LGTM.  I assume using copy is a change we'd want to make to most of the public properties in the RPC classes, as well as the manual copy in the setters like you just implemented?

joeljfischer:
@adein yeah, ideally at some point we should get to that. Changing all the mutable arrays to immutable ones, and implementing copying on rpcs so they can be copied as well would be other points on the list of "would like to have happen eventuallyish"

Issue #228, opened 2015-06-04
IAP protocolIndexTimer should be reset whenever creating a new control session


joeljfischer:
In `-[SDLIAPTransport createIAPControlSessionWithAccessory:]` there is a nil check for `protocolIndexTimer` before recreating it. This check should probably be removed so that the timer is restarted if the method is called again.

joeljfischer:
https://github.com/smartdevicelink/sdl_ios/blob/release/4.0.0/SmartDeviceLink-iOS/SmartDeviceLink/SDLIAPTransport.m#L189

joeljfischer:
Fixed in #274

PR #274, opened 2015-08-07
Reset protocol index timer when IAP control session is recreated

joeljfischer:
Fixes issue #228

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLIAPTransport.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLIAPTransport.m
@@ -185,10 +185,9 @@ int const streamOpenTimeoutSeconds = 2;
 
     if (self.controlSession) {
         self.controlSession.delegate = self;
-
-        if (self.protocolIndexTimer == nil) {
-            self.protocolIndexTimer = [[SDLTimer alloc] initWithDuration:protocolIndexTimeoutSeconds];
-        }
+        
+        // This will restart the timer if it already exists
+        self.protocolIndexTimer = [[SDLTimer alloc] initWithDuration:protocolIndexTimeoutSeconds];
 
         __weak typeof(self) weakSelf = self;
         void (^elapsedBlock)(void) = ^{
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLIAPTransport.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLIAPTransport.m
@@ -186,8 +186,12 @@ int const streamOpenTimeoutSeconds = 2;
     if (self.controlSession) {
         self.controlSession.delegate = self;
         
-        // This will restart the timer if it already exists
-        self.protocolIndexTimer = [[SDLTimer alloc] initWithDuration:protocolIndexTimeoutSeconds];
+        if (self.protocolIndexTimer == nil) {
+            self.protocolIndexTimer = [[SDLTimer alloc] initWithDuration:protocolIndexTimeoutSeconds];
+        } else {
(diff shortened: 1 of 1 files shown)


joeljfischer:
@adein do you think that this one should also change? It looks like it's the same situation https://github.com/smartdevicelink/sdl_ios/blob/release/4.0.0/SmartDeviceLink-iOS/SmartDeviceLink/SDLProxy.m#L175-L184

adein:
@joeljfischer It depends on what the concern is.  The SDLTimer start: method will stop/cancel and then restart a timer if it is called a second time.  So if the issue is simply about restarting the timer, just calling start: again should be fine.

If the concern has to do with the objects/variables captured by the elapsed/canceled blocks no longer being valid (I'm not sure if this is applicable without looking at each one), then the blocks should simply be recreated (or changed to use getters).

joeljfischer:
The concern was more about the restarting. It should be okay then.

SmartDeviceLink-iOS/SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -185,10 +185,9 @@ - (void)createIAPControlSessionWithAccessory:(EAAccessory *)accessory {
 
     if (self.controlSession) {
         self.controlSession.delegate = self;
-
-        if (self.protocolIndexTimer == nil) {
-            self.protocolIndexTimer = [[SDLTimer alloc] initWithDuration:protocolIndexTimeoutSeconds];
-        }
+        
+        // This will restart the timer if it already exists
```
justinjdickow: cancels the current timer if it exists

SmartDeviceLink-iOS/SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -185,10 +185,9 @@ - (void)createIAPControlSessionWithAccessory:(EAAccessory *)accessory {
...
+        self.protocolIndexTimer = [[SDLTimer alloc] initWithDuration:protocolIndexTimeoutSeconds];
```
adein: I still believe we should call cancel explicitly.  Autorelease pools are drained at the end of every runloop iteration (in this case, I believe it's the main/UI runloop).  Setting an object reference to nil (retain count to 0) will result in the dealloc method being called *after* all events in that runloop have been processed.  I doubt this will ever actually cause an issue, but logically it makes more sense to me to explicitly cancel the timer before creating the new one, instead of relying on the runloop/autorelease pool to do it when it gets around to it.

joeljfischer: I'm pretty sure that this is a released object, not autoreleased. This should be a case where ARC puts in manual retain / release calls. An autorelease from a pool draining would just decrement the retain count by 1 anyway, the object's dealloc is enqueued to the runloop / queue as soon as the retain count reaches 0. Calling cancel would enqueue that call the same way.

That said, we may want to make the code more semantic / explicit that we're cancelling the timer by messaging 'cancel', but I'm very nearly positive that calling it will be no different than not, in this case.

joeljfischer: Do you have any more thoughts? If you think it would be better to be explicit about the `cancel` call, I can work that in. It shouldn't be too bad.

adein: I'd like if it were added, if only for code clarity.

joeljfischer: Made the change, let me know.

adein:
Looks good.

Issue #267, opened 2015-08-04
onOnLockScreenNotification argument is an ENUM instead of the notification object


adein:
In SDLProxyListener.h, the argument in the method definition is: (SDLLockScreenStatus*) notification.  It should be (SDLOnLockscreenStatus*) notification.

A branch containing the changes will pushed soon.

joeljfischer:
Fixed with #289

PR #276, opened 2015-08-10
onOnLockscreenNotification bugfix

adein:
The proxy passes the notification object (SDLOnLockscreenStatus), but the SDLProxyListener protocol currently defines the method as expecting the enum (SDLLockScreenStatus).

Fixes issue #267


joeljfischer:
The conflicts here are because this code is based on an older version of the file that has incorrect formatting that was since fixed (on 2015/07/01). Can you resolve this by either recreating based on the current version of the file, or by merging in the target branch to the current one?

Issue #271, opened 2015-08-04
PutFile Streaming is not aligned with Android


adein:
The methods/delegates/etc for streaming PutFile need to be updated to match the functionality in the Android proxy.

Pull requests will be added soon.

joeljfischer:
What milestone is this intended for?

joeljfischer:
This is closing because file streaming will radically change in 4.1

PR #277, opened 2015-08-10
Streaming Putfile enhancement

adein:
Changes for PutFile to allow streaming similar to Android.

Issue #271


joeljfischer:
This is a pretty big change, any chance we can push it to 4.1?

Couple of issues I see:

1. Based on develop from June 2. This will need to be updated by merging in the target branch to this branch. A lot seems to have changed, so this might be a headache (sorry!).
1. It would be nice if it followed the style guide, but I think we said we weren't going to enforce it until post 4.0.0, so no worries if it can't happen.
1. It would be really nice to have tests, but since that class doesn't have any tests built yet and that's a major project, I certainly wouldn't expect them unless we pushed the release back.

I haven't yet tested or thoroughly reviewed the code.

adein:
I'm not sure about 4.0 vs 4.1.  @Heath-FoMoCo - does this need to be 4.0?

adein:
@joeljfischer Yes, this can be delayed to 4.1.

joeljfischer:
:+1:

PR #280, opened 2015-08-10
Feature/heartbeat

### LINKED ISSUE #264: Heartbeat Implementation for iOS
See #106

justinjdickow:
[READY FOR REVIEW]

  * Updates the SDLTimer class to support repeat
  * Implements responding to heartbeats with heartbeat ACK independent of version
  * If version is 3 or greater, the proxy will now send a heartbeat every 5 seconds. If a heartbeat ACK is received, when the timer fires at 5 seconds the proxy will send another heartbeat. If the timer fires and no heartbeat ACK has been received, the proxy is disposed.
  * Any protocol message that is received on the transport (except for heartbeat and heartbeat ACK) will reset the heartbeat timer back to 5 seconds.

Resolves #264

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
@@ -326,7 +326,6 @@ const UInt8 MAX_VERSION_TO_SEND = 4;
 
 #pragma mark - SDLProtocolListener Implementation
 - (void)handleProtocolSessionStarted:(SDLServiceType)serviceType sessionID:(Byte)sessionID version:(Byte)version {
-    NSLog(@"Handle session started for service Type: %@", @(serviceType));
     self.sessionID = sessionID;
     [self storeSessionID:sessionID forServiceType:serviceType];
 
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
@@ -39,6 +39,7 @@ const UInt8 MAX_VERSION_TO_SEND = 4;
 @property (strong) NSMutableData *receiveBuffer;
 @property (strong) SDLProtocolReceivedMessageRouter *messageRouter;
 @property (nonatomic) BOOL heartbeatACKed;
+@property (nonatomic, strong) SDLTimer *heartbeatTimer;
 @end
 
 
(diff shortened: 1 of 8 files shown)


SmartDeviceLink-iOS/SmartDeviceLink/SDLTimer.h
```diff
@@ -11,7 +11,8 @@
 @property (assign) float duration;
 
 - (id)init;
```
joeljfischer: since you're changing it, can you also change this to an instancetype?

asm09fsu:
@justinjdickow is there a reason why within the onProtocolMessageReceived:, the heartbeatTimer is restarted, but the heartbeatACKed is not set to YES? Also, would it be wiser to move the actual restarting of the timer, as well as those checks to the handleHeartbeatACK method?

@adein

justinjdickow:
@asm09fsu The requirement is that any protocol message received just resets the timer, heartbeatACKed is only set to yes when a heartbeat is ACKed. No it would not be wiser, the timer needs to reset whenever a protocol message is received, not when a heartbeat ACK is received.

adein:
@justinjdickow Ah, ok.  I was wondering about that too, and I didn't notice that in the spec.  It's still a bit weird though.  Having an ACK seems redundant when ANY message will restart the timer.  I guess it's covering the case where nothing else is being exchanged except the heartbeat itself.

justinjdickow:
@adein Right, so it basically delays heartbeats until no other messages are being sent.

adein:
@justinjdickow The only other thing I noticed is that the heartbeat timer is linked to the time of the last protocol message that was received, instead of (or in addition to) being linked to whenever the last heartbeat request was sent.  So if a heartbeat is sent after a protocol message is received (and there is no following traffic), the other side (for the ACK) won't get the full amount of time to respond to the heartbeat request as it should.  Hopefully this makes sense?

justinjdickow:
@adein Remember that the timer repeats, so when the timer fires (i.e. a heartbeat is sent), the timer is set to fire again a full period later (the hardcoded 5.0), at which time it checks if the previous heartbeat was ACK'd and sends a new heartbeat or disconnects. So the time to respond to a heartbeat is always 5 seconds or more.

SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
```diff
@@ -325,6 +362,10 @@ - (void)destructObjects {
         self.messageRouter = nil;
         self.transport = nil;
         self.protocolDelegate = nil;
+        if (self.heartbeatTimer != nil) {
+            [self.heartbeatTimer cancel];
```
joeljfischer: No need to cancel here. When it is nil'd, it automatically cancels in `-[SDLTimer dealloc]`.

adein:
@justinjdickow After looking through in more detail, I see that both the heartbeat timeout (elapsed) block and the sending of heartbeat requests are combined (inside the block).  I was thinking they were done separately (I thought heartbeats would always be sent on a schedule, not just when no other traffic is occurring).  Everything makes sense now, thanks.

joeljfischer:
:+1: LGTM, the only thing that was weird to me when reviewing was the same thing @asm09fsu brought up. Maybe add a comment explaining why it doesn't set heartbeatACKed to YES? I get that it's following the spec, but it does seem counter-intuitive at first readthrough.

justinjdickow:
@joeljfischer @asm09fsu I'm fine with setting heartbeatACKed to yes when any message is received if it makes more sense to you guys

Issue #156, opened 2015-03-31
Video Streaming


joeljfischer:
We need to have a conversation around how this ought to be implemented

Heath-FoMoCo:
What do you mean by "navigation"?

joeljfischer:
@Heath-FoMoCo Fixed the title, sorry about that

Heath-FoMoCo:
Thanks,
We have this "mostly" implemented.  @hugh22 can explain details.

joeygrover:
@Heath-FoMoCo @hugh22 is it possible for you to create a pull request for this so the community can see progress?

hugh22:
Which part? Raw h.264 Video streaming capability is in the pull request submitted since February 26. You can show the community some progress by pulling that in.

After our meeting with Julius I was under the impression that you folks were going to handle the migration of moving the hardware encoding in to the proxy. Is that correct?

joeygrover:
I think we briefly discussed how it would be a lot easier to merge if that pull request  #105  was broken up into the specific pull requests that relate to their respective fixes/features. Then we can hopefully get all those changes into the main branch.

idamaster:
perhaps we can discuss what was in the pull request at the next meeting and see what makes sense to logically break down?

joeljfischer:
I would prefer to do it 'offline', since it might take a while and I don't think it's really beneficial to a public call. I also need a bit of time to look over it as well, which I haven't had time to do with the 4.0 features still running behind.

justinjdickow:
We have the same issue here as #106 - need a priority because this is a v 3 feature. I also do not understand how an application figures out whether a head unit its connected to with v 3 or higher actually supports video streaming (by failing to open the video streaming service?)

PR #286, opened 2015-08-11
Video / Audio Streaming

### LINKED ISSUE #204: End Service ACK 0x05 and End Service NACK 0x06 are missing from SDLFrameData enum and not used
I think the proxy doesn't check whether services are ended properly before disconnecting transports which might be fine, but for stopping h.264 and pcm data streaming, these will likely need to be implemented.

joeljfischer:
:dog: [Ready for Review] :dog:  

#### Summary
This PR will be used to enable video streaming via the proxy library. The idea is to create a new high level object that the developer can choose to create and subscribe to that will enable them to integrate video streaming capability into their application.

#### Detail
Eventually this feature should become split out and optional to the developer, so that they don't even have the files for it if they do not want to use it. However, for now it will be a part of the core library. The idea is to have an object that the developer can create, set up, and subscribe to, that will be passed into the proxy so that the proxy can interact with it as it needs to. This object will give the developer notifications such as if the Head Unit ended the service, and the ability for the developer to pass in a new frame when ready.

Resolves #204, #156


joeljfischer:
@adein @Heath-FoMoCo We've tested audio and video streaming. Let me know when you've reviewed it on your end.

asm09fsu:
Looks good to me!

Issue #268, opened 2015-08-04
Minimum iOS version supported is 8.0


adein:
This requires app developers to only support 8+ if they use the SDK.  Many app developers wish to support as early as 5 or 6, and are not happy about this (new) requirement. The only 2 instances we've found that require iOS 7/8 are avoidable.  They are:

1. Using NSURLSession (as opposed to supporting NSURLSession and NSURLConnection)
2. Using beginBackgroundTaskWithName (as opposed to beginBackgroundTaskWithExpirationHandler)

Pull requests will be added soon.

joeljfischer:
Quick comment is that even though the framework target's minimum version is 8.0 because that's the minimum for frameworks, the minimum for the Cocoapods is currently 7.0, and it can go lower. So as long as the code itself is compatible with pre-7.0, we can change the podspec to allow lower than 7.0. The framework target would have to remain 8.0. We could theoretically add a static lib target that targets lower than 8.0 as well, if it becomes needed. Let me know if that makes sense haha.

adein:
Are we using dynamic framework linking for some reason?  That's only required if you're including/linking against another framework from within ours..?

joeljfischer:
I'm not sure what you mean. I was talking about the library itself generating a dynamic framework; it currently isn't linking other frameworks in. Eventually the idea is that it will link frameworks in, i.e. for the logging framework which would be an 'outside' project that's linked in. Dynamic framework linking is nice because it can be runtime / optional. Cocoapods also allows us to have dependencies like this as well (for 7.0 or less). However, a static target wouldn't allow for this, and I'm not sure what the solution would be.

adein:
Correct.  We've always used a static framework in the past because that's all that was supported in earlier versions of iOS.

We can't just tell developers that their apps will only work on 8 and later with our SDK.  Even though a large percentage of users (at least in the US) update to the latest iOS, there are still users (especially in other markets like Asia) that cannot or will not update, and many app developers target those regions/audiences.

We need some sort of solution to support iOS (5 or 6, ideally 5)+; whether it's creating a duplicate/static version of the framework or something else.

joeljfischer:
Right, that's what Cocoapods is meant to cover, it can go back as far as we need to. I realize that we may have to create a static version, I was just hoping to wait for developers to complain before doing it, because it's so much extra work compared to Cocoapods & dynamic frameworks.

adein:
[...]

PR #288, opened 2015-08-12
Add abstraction for NSURLSession and NSURLConnection to support earlier iOS versions

adein:
Fixes item 1 of issue #268 

Changes SDLProxy to use a connection object that checks iOS version and uses either:

* SDLURLConnection handles NSURLConnection (for iOS 6 and below)
* SDLURLSession handles NSURLSession (for iOS 7 and above)


joeljfischer:
This is closed in favor of #290

PR #289, opened 2015-08-12
Fix parameter type for onOnLockscreenNotification

adein:
Change method parameter type for onOnLockscreenNotification from (SDLLockscreenStatus *) to (SDLOnLockscreenStatus *).


joeljfischer:
I'm going to reformat before pushing 4.0.0 final, but it should be noted that this would change a lot of formatting from correct to incorrect. This ought to only change the one line in order to keep the git history clean.

adein:
Yeah, I don't know why all of those changes are in here.

SmartDeviceLink-iOS/SmartDeviceLink/SDLProxyListener.h
```diff
@@ -57,69 +57,69 @@
 @class SDLUnregisterAppInterfaceResponse;
 @class SDLUnsubscribeButtonResponse;
 @class SDLUnsubscribeVehicleDataResponse;
+@class SDLOnLockscreenStatus;
```
joeljfischer: This should be placed in correct alphabetical order.

adein:
@joeljfischer Same question: will these be fixes during the merge, or should Infosys fix them in their PR branch?

joeljfischer:
Hmm...ideally it should be done before the merge, so that it is it's own separate commit, and not a part of the merge commit.

joeljfischer:
:+1: LGTM

PR #290, opened 2015-08-17
iOS 6 compatibility

joeljfischer:
:horse: [READY FOR REVIEW] :horse: 

This creates two new "wrapper" classes for `NSURLConnection` that makes the API more usable in light of removing `NSURLSession`. The class that would be interacted with is `SDLURLSession` that mirrors fairly closely the `NSURLSession` API, while not having the complexity that is unneeded at this point.

Instances of usage of `NSURLSession` have been altered to use `SDLURLSession` instead.


joeljfischer:
@adein This really just needs to be tested. I really would like to do unit tests, but we may have to smoke test for now and I'll write them right after 4.0.0. If I have time, I'll definitely write the unit tests before.

joeljfischer:
@adein @justinjdickow @asm09fsu This is ready for review. To test, I used SPT + Gen 3, and I added unit tests for SDLURLRequestTask, but I haven't for SDLURLSession.

SmartDeviceLink-iOS/SmartDeviceLink/SDLURLRequestTask.h
```diff
@@ -0,0 +1,45 @@
...
+
+@end
+
+
+@interface SDLURLRequestTask : NSObject
+
+@property (weak, nonatomic) id<SDLURLRequestTaskDelegate> delegate;
+@property (assign, nonatomic) SDLURLRequestTaskState state;
+
+- (instancetype)initWithURLRequest:(NSURLRequest *)request completionHandler:(SDLURLConnectionRequestCompletionHandler)completionHandler;
```
adein: This should have the NS_DESIGNATED_INITIALIZER macro.

adein:
@joeljfischer LGTM. I'll see if Infosys can fit in some testing and then I think we'll be good.

joeljfischer:
@adein okay, sounds good.

adein:
@joeljfischer This resulted in a crash on the iOS 6 device that Infosys used for testing.  We'll try to get more details.

joeljfischer:
The title was changed because additional changes are needed besides just removing NSURLSession. Specifically Base64 changes.

Issue #241, opened 2015-06-19
Proxy allows attempting to send an RPC when not connected


joeljfischer:
See #238

adein:
If this could be fixed as a hotfix, it would be another one I'd like to get in ASAP if possible.  It allows developers to create apps that will never be able to reconnect to SYNC until they are terminated and relaunched.  At the same time, it's not easy for devs to root-cause the connection issue.

If we're only making it so that the proxy uses a BOOL to track connection state and prevent sending an RPC when not connected, I think it should be a pretty easy/quick change.

joeljfischer:
Yeah, probably a BOOL within the `sendRPC:` method of `SDLProxy` would be sufficient. The proxy isn't currently set up to know whether or not it's connected because it's assumed to be destroyed if not connected however, so that would be a bit of a change to make that work well.

adein:
In my testing, I made a class BOOL (readonly public, readwrite private), that gets updated in the onProxyClosed / onProxyOpened events before the proxy listener delegate is called.

PR #298, opened 2015-09-09
[WIP] Developer API Enhancements Phase 1

### LINKED ISSUE #223: Refactor PutFile streaming to not use associated objects
Associated objects should only be used as a last resort, when modifying a system framework with new properties is really the only option. With PutFile streaming I don't believe we're at that point, this code could be refactored to be cleaner and less fragile. I have also seen situations where this code does not seem to work at all, and an alternate method to PutFile had to be used.

adein:
@joeljfischer has taken over this post :smile:

Fixes #223, #241 

This post was last rewritten on 2016/07/27.

**Not Ready For Merge**

This adds a new high-level abstraction layer on top of the current SDL layers such as SDLProxy. When 5.0 rolls around, many of those old layers will be removed and the functionality will be hidden behind the new high level layers being added here. Additional functionality will be built on top of the new developer api for quicker and more intuitive SDL use.

Features included:
* Quicker and easier initial setup with ahead of time configuration.
* RPC request/response bookkeeping (linking a handler block to a correlation ID).
* Commands and button subscription handler bookkeeping.
* Automatic correlation ID management.
* Lifecycle manager to do most of the startup / shutdown bookkeeping for you. Define your configuration and the manager takes care of everything else.
* A file manager to keep track of what's on the head unit and manage adding and deleting files.
* A permissions manager to help keep track of what you're allowed to do and when permissions change.
* A lock screen manager to automatically show and hide a lock screen with additional options.

To-Do:
- [ ] Add and Improve Documentation, example apps, and guides
- [x] Add a subscribable state object for keeping track of the RAIR data, HMI status, etc.
- [x] Deprecate the objects that will be removed in the next major version (which is pretty much all of them). However, they will not be replaced immediately, which will cause un-fixable warnings for developers. We need to decide if this is better, or just replacing them in 5.0 is better.
- [x] Group files by public / private to make clearer which Dev API classes are to be interacted with and which will cause version changes by altering.
- [ ] SDLFile store data to RAM instead of disk, or, store data to disk asynchronously when needed.
- [ ] Localize lock screen text
- [x] Localize errors (?)
- [x] Handle ChangeRegistration automatically as needed (need more definition around use cases)
- [ ] Logging Improvements: Configuration to turn on / off some logging such as Siphon, file logging. Filtering? Integrate a library such as SuperLogger?
- [ ] Consider removing publicly facing state strings in order to avoid potential minor / major version changes to update them.
- [ ] Handle manager startup error cases
- [ ] Make sure all TODOs are accounted for

To Test:
- [ ] LifecycleManager hash resumption use cases are accounted for
- [ ] PermissionManager
- [x] FileManager under load


adein:
Oops, it looks like there's some unnecessary autorelease pools that I forgot to remove.  I'll fix them in another commit.

justinjdickow:
Notes from Aaron from our previous convo:

* Manage lock screen within proxy
* Find a way to not have to cast the RPCResponse in the response handler blocks
* Have 1 protocol for notifications, with all of them optional. check respondsTo before callbacks
* Make SDLProxyBase a singleton, not using inheritance, & change the name to SDLManager or SDLController
* Deprecate SDLProxy
* 2 handler blocks for sendRPC: success and failure

Issue #306, opened 2015-09-18
SDLHexUtility causes crash when streaming audio over TCP/IP.


asm09fsu:
Looks like when we are streaming audio over TCP/IP, we get occasional crashes on the device. After digging into the code, it looks like the culprit is SDLHexUtility and it's performance. The memory allocation on this function is very high, sometimes upwards of 300+MB, which is causing the crash.

joeljfischer:
We probably need an `@autoreleasepool` here: https://github.com/smartdevicelink/sdl_ios/blob/release/4.0.0/SmartDeviceLink-iOS/SmartDeviceLink/SDLTCPTransport.m#L65

I'm guessing that would fix it since it would force the autorelease of the previous iteration's temp variables.

asm09fsu:
@joeljfischer yes, this fixes the memory allocation issue. I would highly recommend changing this to a C based function. The execution time is 97% (on average) faster on just the parsing alone.

I went ahead and added this as a PR, located here: https://github.com/smartdevicelink/sdl_ios/pull/307

joeljfischer:
Fixed in #307

PR #307, opened 2015-09-18
Fix SDLHexUtility's memory allocation and performance.

asm09fsu:
The purpose of this PR is to fix the memory allocation (by using an autoreleasepool) and the performance time (by using a C function).

Issue https://github.com/smartdevicelink/sdl_ios/issues/306

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLHexUtility.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLHexUtility.m
@@ -7,16 +7,32 @@
 
 @implementation SDLHexUtility
 
-+ (NSString *)getHexString:(UInt8 *)bytes length:(int)length {
-    NSMutableString *ret = [NSMutableString stringWithCapacity:(length * 2)];
-    for (int i = 0; i < length; i++) {
-        [ret appendFormat:@"%02X", ((Byte *)bytes)[i]];
+static inline char itoh(int i) {
+    if (i > 9) return 'A' + (i - 10);
+    return '0' + i;
+}
+
+NSString* getHexString(NSData *data) {
+    NSUInteger i, len;
+    unsigned char *buf, *bytes;
+    
+    len = data.length;
+    bytes = (unsigned char*)data.bytes;
+    buf = malloc(len*2);
+    
+    for (i=0; i<len; i++) {
+        buf[i*2] = itoh((bytes[i] >> 4) & 0xF);
+        buf[i*2+1] = itoh(bytes[i] & 0xF);
     }
-    return ret;
+    
+    return [[NSString alloc] initWithBytesNoCopy:buf
+                                          length:len*2
+                                        encoding:NSASCIIStringEncoding
+                                    freeWhenDone:YES];
 }
 
 + (NSString *)getHexString:(NSData *)data {
-    return [self getHexString:(Byte *)data.bytes length:(int)data.length];
+    return getHexString(data);
 }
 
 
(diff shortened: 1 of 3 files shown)


SmartDeviceLink-iOS/SmartDeviceLink/SDLHexUtility.m
```diff
@@ -7,16 +7,32 @@
...
+    for (i=0; i<len; i++) {
+        buf[i*2] = itoh((bytes[i] >> 4) & 0xF);
+        buf[i*2+1] = itoh(bytes[i] & 0xF);
     }
-    return ret;
+    
+    return [[NSString alloc] initWithBytesNoCopy:buf
+                                          length:len*2
+                                        encoding:NSASCIIStringEncoding
+                                    freeWhenDone:YES];
```
joeljfischer: If this fails, then the buffer isn't freed. You'll need to check that the return of the `init` method is not nil. If it is, call `free(buf)` and return nil.

See the "special considerations" on the apple documentation for the method.

SmartDeviceLink-iOS/SmartDeviceLink/SDLTCPTransport.m
```diff
@@ -63,8 +63,10 @@ - (void)connect {
 
 - (void)sendData:(NSData *)msgBytes {
     dispatch_async(_sendQueue, ^{
-        NSString *byteStr = [SDLHexUtility getHexString:msgBytes];
-        [SDLDebugTool logInfo:[NSString stringWithFormat:@"Sent %lu bytes: %@", (unsigned long)msgBytes.length, byteStr] withType:SDLDebugType_Transport_TCP toOutput:SDLDebugOutput_DeviceConsole];
+        @autoreleasepool {
+            NSString *byteStr = [SDLHexUtility getHexString:msgBytes];
+            [SDLDebugTool logInfo:[NSString stringWithFormat:@"Sent %lu bytes: %@", (unsigned long)msgBytes.length, byteStr] withType:SDLDebugType_Transport_TCP toOutput:SDLDebugOutput_DeviceConsole];
+        }
```
joeljfischer: I would wrap the entirety of the code in `dispatch_async` block in the `autoreleasepool`

asm09fsu: What is the reasoning behind this? Not saying it's incorrect, just curious about your logic.

asm09fsu:
@joeljfischer the latest commit fixes the special consideration where the NSString fails to initialize. I also re-added the old conversion method as a fallback, however the likely hood of us needing it should (hopefully) be very minimal.

SmartDeviceLink-iOS/SmartDeviceLink/SDLTCPTransport.m
joeljfischer: Since this method is passing this block to a separate thread, it's in apple's recommendations that if you're doing lots of local variable creation you should `@autoreleasepool`. This is only a problem if you call it a lot like we often might. It may result in a slight performance hit to use it, but since we are already using it, we may as well clean up all the temp variables created in the block rather than just some and leaving others to be cleaned up later.

asm09fsu: Added

joeljfischer:
@asm09fsu Also see SDLHexUtilitySpec.m and make sure it's up to date or write an issue to get it up to date.

asm09fsu:
@joeljfischer I think the spec should be sufficient. The first test is actually just calling the new function directly. The second one is actually testing our less performant version.

joeljfischer:
LGTM!

PR #316, opened 2015-10-21
Fix implicit captures of self

joeljfischer:
Fixes improperly creating strongSelf variables within blocks and inadvertantly capturing self


asm09fsu:
@joeljfischer shouldn't a check of strongSelf be done after dancing from weak to strong, as a safety check?

joeljfischer:
@asm09fsu No, because it's the first thing done within the block and the block itself would be released if `self` had been released. So it's running the block from self, then strongifying the weak self which will exist. Does that make sense?

asm09fsu:
@joeljfischer I agree, I just usually see it done this way, wasn't sure if added the extra check (or not, would help or hinder us).

joeljfischer:
@asm09fsu Yeah, it's a good question, but I don't think the added check would help us.

PR #318, opened 2015-12-08
Add new "remote control" RPCs, enums, etc.

joeljfischer:
This adds new RPCs, enums, structs, and more to SDL to support remote control APIs. Remote control APIs are used to allow an app to control traditional car features such as radio and climate control.


alangonzalez:
Question... Climate control - should it have options for heated seats/heated steering wheel On/Off?  They are somewhat related if someone wants to get the cabin in the right climate.

joeljfischer:
This would be a question for @justinjdickow or @joeygrover since they created the spec this implements.

joeljfischer:
A new PR will be opened when this is ready.

PR #326, opened 2015-12-16
Fix SystemRequest RPC request and response unavailable to the developer

### LINKED ISSUE #323: SystemRequest RPCs on SDL iOS SDK missing
We had a workshop over here to work with online traffic and we saw some issues on the iOS SDK for SDL (4.0.1).
 
The file SmartDeviceLink.h is missing some imports. It should contain:
#import “SDLSystemRequest.h”
#import “SDLSystemRequestResponse.h”
 
If an iOS app wants to send SystemRequest it would need to import them by themselves.
 
The file SDLProxyListener.h should contain:
- (void)onSystemRequestResponse:(SDLSystemRequestResponse *)response;
 
If an iOS app sends a SystemRequest is cannot receive its response.

joeljfischer:
Fixes #323 

* This requires a minor version change because we are adding new apis to be public facing.


joeljfischer:
@asm09fsu @justinjdickow please review

PR #328, opened 2016-01-19
Remove unneeded IAP background code

### LINKED ISSUE #292: IAPTransport requests background processing time regardless of connection state
In the SDLIAPTransport applicationDidEnterBackground: method, it will ask for extra 'background' processing time to try to extend background execution time.

My concerns are:
1. This is done regardless of whether the app is actually connected to a headunit.
2. Incoming iAP data will wake up an app (temporarily) to process the data.
3. If an app is doing something while backgrounded, it should be the one asking for more background time, not the SDK.

https://github.com/smartdevicelink/sdl_ios/blob/release/4.0.0/SmartDeviceLink-iOS/SmartDeviceLink/SDLIAPTransport.m#L121

I suggest (at a minimum) adding a check for connection state, or removing this altogether.

joeljfischer:
Fixes #292

Needs to be tested against an IAP connection.

* Remove code to run in background during IAP sessions, which should already run in the background anyway.
* Prefix private methods for clarity in SDLIAPTransport.m


justinjdickow:
lgtm

PR #330, opened 2016-01-19
Remove code pertaining to App Launching which broke in iOS 9

### LINKED ISSUE #248: iOS 9 breaks app launching
In an effort to stop apps from scanning what other apps are installed on the phone (as a result from apps using it for advertising and marketing), Apple has made major changes to `canOpenURL` and `openURL` for iOS URL schemes.

**Note:** It appears that the `openURL` limitations are a bug. We will be able to do app launching, but it's abilities will be severely limited. We will be able to know what apps are SDL connected and launch them, but not which among iOS SDL connected apps are on the users phone.

**References:**
http://www.macstories.net/linked/ios-9-bringing-changes-to-url-schemes/
http://awkwardhare.com/post/121196006730/quick-take-on-ios-9-url-scheme-changes

joeljfischer:
* Fixes #248


justinjdickow:
lgtm after tests are removed

PR #334, opened 2016-01-22
[WIP] Replace SDLTimer's NSTimers with dispatch_source's running in the background

### LINKED ISSUE #333: Heartbeat timer fails if UI is being used
For some reason it appears that the `NSTimer` being used fails if the UI is in use (for example scrolling a tableview) while the timer is running. This prevents heartbeats from being tracked properly and ultimately causes a disconnect.

joeljfischer:
Fixes #333


asm09fsu:
Can we delete this PR?

joeljfischer:
Yes, I believe so

PR #344, opened 2016-02-02
Update MTU size to actual v3 / v4 size

### LINKED ISSUE #342: Update MTU size for v3 protocol
For the v3 protocol we need to update the mtu with the details below:
Payload of a message will be 131072
Complete message will be 131072+12 byte header

joeljfischer:
Fixes #342


joeljfischer:
@asm09fsu can you do a quick check to make sure this is correct, then i'll pull it into master

asm09fsu:
@joeljfischer Confirmed that SDLProtocol is sending 131084 bytes (request 131072 + header 12)

PR #346, opened 2016-02-03
On Session Ended, hashID should be sent for RPC Service.

### LINKED ISSUE #345: On Session Ended, hashID should be sent for RPC Service.
When ending the RPC service, the SDK needs to include the hashId received from the start service ack for service type 7.  We need to store the hash id used for each session that is established via start session ack and send in the end service request for service type 7.  This is only required to end an RPC session, other services established do not require the hash id to be ended.

asm09fsu:
Fixes #345


SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
```diff
@@ -28,6 +28,7 @@ @interface SDLProtocol () {
     dispatch_queue_t _sendQueue;
     SDLPrioritizedObjectCollection *_prioritizedCollection;
     NSMutableDictionary *_sessionIDs;
+    NSMutableDictionary *_hashIDs;
```
joeljfischer: Do you think that instead of storing multiple dictionaries with sessionIDs / hashIDs / whatever else in the future, it should just be one dict with various metadata about the service? That's what I'm starting to lean towards.

asm09fsu: This would probably be a good idea for a future enhancement. We would need to make sure it's dynamic enough to accept any future metadata. Would this be suitable for now, do you think?

joeljfischer: Either dynamic enough or private enough to change it without consequence. Probably the latter.

I'm considering the change in this security branch to track which services are encrypted, so that if you try to send an encrypted RPC on a non-encrypted service type it fails, I think this would fit in the same place. So I would say yeah, it probably fine for now, but should change sooner rather than later. Maybe as a "hotfix" separate from this.

joeljfischer: https://github.com/smartdevicelink/sdl_ios/issues/350

joeljfischer: Also, *technically* it should be `hashIds`, since "Id" is short for one word, not two. Not sure how we want to handle that since it's already done incorrectly with `sessionIDs`

asm09fsu: Just tried to best following existing naming so when we do changes it'll be consistent ;)

SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
```diff
@@ -310,7 +345,7 @@ - (void)sendRawData:(NSData *)data withServiceType:(SDLServiceType)serviceType {
 
 
 #pragma mark - SDLProtocolListener Implementation
-- (void)handleProtocolStartSessionACK:(SDLServiceType)serviceType sessionID:(Byte)sessionID version:(Byte)version {
+- (void)handleProtocolStartSessionACK:(SDLServiceType)serviceType sessionID:(Byte)sessionID hashID:(UInt32)hashID version:(Byte)version {
```
joeljfischer: Unfortunately, we're both touching the same code here. The security framework deprecates and updates most of these `SDLProtocolListener` calls. This one that you're changing would have necessitated a major version change (so that you are aware in the future). The security framework adds a method that passes the entire SessionACK header, that might require some changes to this on your part.

SmartDeviceLink-iOS/SmartDeviceLink/SDLV2ProtocolHeader.m
```diff
@@ -82,6 +82,7 @@ - (void)parse:(NSData *)data {
     UInt32 *uintPointer = (UInt32 *)data.bytes;
     self.bytesInPayload = CFSwapInt32BigToHost(uintPointer[1]); // Data is coming in in big-endian, so swap it.
     self.messageID = CFSwapInt32BigToHost(uintPointer[2]); // Data is coming in in big-endian, so swap it.
+
```
joeljfischer: Random blank line?

SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
```diff
@@ -71,6 +77,22 @@ - (UInt8)retrieveSessionIDforServiceType:(SDLServiceType)serviceType {
...
+    NSNumber* number = _hashIDs[@(serviceType)];
+    if (!number) {
+        NSString *logMessage = [NSString stringWithFormat:@"Warning: Tried to retrieve hashID for serviceType %i, but no hashID is saved for that service type.", serviceType];
+        [SDLDebugTool logInfo:logMessage withType:SDLDebugType_Protocol toOutput:SDLDebugOutput_File | SDLDebugOutput_DeviceConsole toGroup:self.debugConsoleGroupName];
+    }
+    
+    return (number ? [number unsignedIntValue] : 0);
+}
+
+- (void)removeHashIDForServiceType:(SDLServiceType)serviceType {
```
joeljfischer: Can we prefix new private methods with `sdl_`?

asm09fsu: will do.

asm09fsu:
@joeljfischer how's this latest commit?

SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocolListener.h
```diff
@@ -9,7 +9,8 @@
 @protocol SDLProtocolListener <NSObject>
 
 @optional
-- (void)handleProtocolStartSessionACK:(SDLServiceType)serviceType sessionID:(Byte)sessionID version:(Byte)version;
+- (void)handleProtocolStartSessionACK:(SDLServiceType)serviceType sessionID:(Byte)sessionID version:(Byte)version __deprecated_msg("use -handleProtocolStartSessionACK:sessionID:hashID:version: instead");
+- (void)handleProtocolStartSessionACK:(SDLServiceType)serviceType sessionID:(Byte)sessionID hashID:(UInt32)hashID version:(Byte)version;
```
joeljfischer: I'm still not happy with this because [security PR](https://github.com/smartdevicelink/sdl_ios/pull/348) is creating a different method, so in the security PR I'd have to deprecate this method. It would be better to have it follow [the security framework's changes to this protocol](https://github.com/smartdevicelink/sdl_ios/blob/feature/external_security_support/SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocolListener.h) instead.

In any case, thanks for the merge nightmare :cold_sweat:

SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
```diff
@@ -98,10 +120,21 @@ - (void)sendEndSessionWithType:(SDLServiceType)serviceType {
...
     header.serviceType = serviceType;
     header.frameData = SDLFrameData_EndSession;
-    header.sessionID = [self retrieveSessionIDforServiceType:serviceType];
+    header.sessionID = [self sdl_retrieveSessionIDforServiceType:serviceType];
+    
+    NSData* payload = nil;
+    UInt32 hashId = [self sdl_retrieveHashIDForServiceType:serviceType];
+    if (hashId) {
+        payload = [NSData dataWithBytes:&hashId length:4];
+        [self sdl_removeHashIDForServiceType:serviceType];
```
joeljfischer: Should this be done in `sendEndSession` or after receiving an ACK?

joeljfischer:
This was fixed in 4.2.0

PR #347, opened 2016-02-04
Add localization class.

kshala-ford:
The purpose of the localization class is to help out when the phone/app language differs from the SDL/head unit language. The class can load a bundle instance based on a specified language/region.

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.h
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.h
@@ -5,79 +5,105 @@
 #import <Foundation/Foundation.h>
 
 /**
- Provides a simplified way to handle localization for a specific language that may not match the system language.
+ Provides localization features fully compatible to the existing localization structure of iOS.
+
+ In addition to stock localization features it supports the fallback to other localizations which enables
+ the app to partly implement localizations for regions. This will improve maintenance for region based
+ localizations.
+ 
+ An app may contain a lproj for "en" and "en-GB" whereof "en-GB" is only a subset of "en" overwriting
+ only specific keys which doesn't match the regular "en" localization. An object of this localization class
+ will try out to to localize a key from "en-GB" first and if it fails it then tries "en".
+ 
+ Furthermore a specific language and its localization can be loaded besides the phone language.
+ 
+ The class is fully plural rule enabled using the existing strucutre of stringsdict files.
  */
 @interface SDLLocalization : NSObject
 
-/** The language code used by this instance. This code can differ from initialization regarding fallback. */
-@property (readonly) NSString *language;
-
-/** The region code used by this instance. This code can differ from initialization regarding fallback. */
-@property (readonly) NSString *region;
-
-
-/** 
- The bundle object using language and region. If region is not provided or localization for that region does not exist
- this property is set to nil.
- */
-@property (readonly) NSBundle *defaultBundle;
-
 /**
- The locale object using language and region. If region is not provided or localization for that region does not exist
- this property is set to nil.
- */
-@property (readonly) NSLocale *defaultLocale;
-
-/**
- The bundle object using language only as a fallback to the default bundle. 
- If the default bundle does not exist or it does not contain a specific localization
- then this fallback bundle is used.
- */
-@property (readonly) NSBundle *fallbackBundle;
-
-/**
- The locale object using language only as a fallback to the default locale.
- If the default bundle does not exist or it does not contain a specific localization
- then this fallback locale is used.
- */
-@property (readonly) NSLocale *fallbackLocale;
-
-/**
- Returns the default bundle. 
- Return the fallback bundle if the default bundle does not exist.
- Returns nil if both bundles do not exist.
+ Returns an array of bundles used by this localization object.
 */
-@property (readonly) NSBundle *bundle;
+@property (nonatomic, copy, readonly) NSArray *bundles;
 
 /**
- Returns the default locale.
- Return the fallback locale if the default locale does not exist.
- Returns nil if both locales do not exist.
+ Returns an array of locales used by this localization object.
  */
-@property (readonly) NSLocale *locale;
-
+@property (nonatomic, copy, readonly) NSArray *locales;
 
 /**
- Returns the default localization object representing the system language.
+ Returns an array of language ID strings used by this localization object.
+ */
+@property (nonatomic, copy, readonly) NSArray *localizations;
+
+/**
+ Returns the shared default localization object using the preferred localizations of the apps main bundle.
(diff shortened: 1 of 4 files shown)


SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.h
```diff
@@ -0,0 +1,83 @@
...
+
+#import <Foundation/Foundation.h>
+
+/**
+ Provides a simplified way to handle localization for a specific language that may not match the system language.
+ */
+@interface SDLLocalization : NSObject
+
+/** The language code used by this instance. This code can differ from initialization regarding fallback. */
+@property (readonly) NSString *language;
```
joeljfischer: These should be `nonatomic`, and `copy` is possible, `strong` if not.

SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.h
```diff
@@ -0,0 +1,83 @@
...
+
+/** The region code used by this instance. This code can differ from initialization regarding fallback. */
+@property (readonly) NSString *region;
```
joeljfischer: Can you give an example region code. And region codes as defined by whom? SDL (which I think uses an RFC)?

SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.h
kshala-ford: Will be changed.

SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.m
```diff
@@ -0,0 +1,198 @@
...
+    }
+}
+
++ (instancetype)defaultLocalization {
+    static SDLLocalization *object = nil;
+    static dispatch_once_t token;
+
+    dispatch_once(&token, ^{
+        // get the locale identifier
+        NSString *localeIdentifier = [[NSLocale currentLocale] localeIdentifier];
```
joeljfischer: You probably don't want to be using the locale to identify language. If I'm in Japan with my phone, I'm all of a sudden going to start seeing Japenese on my car radio, which is obviously undesirable. ([See this Stack answer](http://stackoverflow.com/questions/1522210/nslocale-currentlocale-always-returns-en-us-not-users-current-language) as proof that this is the incorrect call) Instead, there are two ways to approach this. ([Using this Stack Answer](http://stackoverflow.com/questions/3910244/getting-current-device-language-in-ios))

1. `NSString * language = [NSLocale preferredLanguages].firstObject;` This will give you the two letter code for the currently selected device language.
2. `[[NSBundle mainBundle] preferredLocalizations].firstObject;` This will give you the localization of the current app language.

I'm guessing that the first is what we want, since they could theoretically have localizations for SDL that they don't have for their main app, but I'd be willing to hear others' feedback on that.

SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.h
kshala-ford: Language and region code will come from SDL (see SDLLanguage). It will match ISO 639-1 and ISO 3166-1 (see https://developer.apple.com/library/ios/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html)

SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.m
```diff
@@ -0,0 +1,198 @@
...
+            fallbackLocale = [NSLocale currentLocale];
+        }
+        
+        object = [[SDLLocalization alloc] init];
+        [object setLanguage:language];
+        [object setRegion:region];
+        [object setDefaultBundle:defaultBundle];
+        [object setDefaultLocale:defaultLocale];
+        [object setFallbackBundle:fallbackBundle];
+        [object setFallbackLocale:fallbackLocale];
```
joeljfischer: Can't these calls just be `object.fallbackLocale = fallbackLocale` etc.?

SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.m
```diff
@@ -0,0 +1,198 @@
...
+    });
+    
+    return object;
+}
+
++ (instancetype)localizationForLanguage:(NSString *)language forRegion:(NSString *)region {
```
joeljfischer: Wouldn't `language` be enough? If not, I think the type signature should be `+ (instancetype)localizationForLanguage:(NSString *)language region:(NSString *)region` by convention.

kshala-ford: Region is necessary and used by existing apps to differentiate US,GB,AU english as well as FR,CA,CH french or PT,BR Portuguese (and many more). Supporting regions allows localization to be more flexible for the app developer. Otherwise developers need to give up existing region based localization (a no go in my opinion).

Signature will be changed to `localizationForLanguage: region:`

SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.m
kshala-ford: Thought this is common?

SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.m
kshala-ford: Both options are not covering what languages the app is capable of. I assume `currentLocale` does that and in addition it tells me the region which `preferred...` doesn't. It's necessary to load the correct bundle.

While testing this I have never seen the issue mentioned on StackOverflow. `currentLocale` was always returning the expected value. But I will have a look at it.

Regarding your EDIT. Language only does not fulfill the full localization feature. See other comment below to line 108.

SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.m
joeljfischer: Only in old code. Dot syntax entered for properties in 2012 I think? That's the general convention now.

SmartDeviceLink-iOS/SmartDeviceLink/SDLLocalization.m
joeljfischer: I looked at this because I know i had looked earlier at these capabilities. I'm quite certain that if I took my english based phone into japan, `currentLocale` would return Japanese.

joeljfischer:
This would need to be completely rewritten to work post-v4.3

PR #348, opened 2016-02-08
Add support for external encryption libraries

### LINKED ISSUE #345: On Session Ended, hashID should be sent for RPC Service.

joeljfischer:
Fixes #345, 370

### Risk

### Testing Plan
Unit tests must be added for new public methods including the additions to the `SDLStreamingMediaManager`, `SDLProtocol`, and `SDLProxy`.

### Summary
This PR adds the capability to hook in a security library to the SDL library. The library will take care of automatically instantiating an object based on the make passed alongside it matching the make of the vehicle connected through an RAI. The library must now also take and hold the app id to pass to the security library. The library defines a protocol that must be followed by the security library's public interface.

The `Protocol` object now has the ability to start a service with security enabled, which will enable the security (TLS) flow to begin and the `Protocol` will begin interacting with the security library and head unit to establish a secure connection.

The `StreamingMediaManager` object now has hooks to start up with encryption using the new start service with encryption methods.

##### Enchancements
* `SDLProtocol` can now set up a TLS session using an external security library
* `SDLProxy` can now take an external security library objects and initialize one when a vehilce make associated with it connects.
* `SDLStreamingMediaManager` can now set up authenticated and/or encrypted sessions of video and audio if available.

##### Bug Fixes
* `compressed` property on SDLV2ProtocolHeader is now `encrypted` in line with updated SDL specs (not developer facing)

##### Unit tests
- [x] StreamingMediaManager
- [x] SDLProxy
- [x] SDLProtocol

##### Other
- [x] Allow more than one make to be associated with a security library (e.g. Ford & Lincoln).


joeljfischer:
@mrapitis @asm09fsu Is this stable enough that it can be merged?

PR #357, opened 2016-02-18
SDLTCPTransport hangs on invalid IP Address.

### LINKED ISSUE #356: SDLTCPTransport hangs main thread on invalid IP Address.
### Bug Report
Currently when attempting to connect to a module via TCP/IP, if the ip address is incorrect, the application will hang, and the user cannot interact with the application.

##### Reproduction Steps
1. Instantiate a proxy using `SDLProxyFactory`'s tcp/ip builder `[SDLProxyFactory buildSDLProxyWithListener:self tcpIPAddress:@"badIPAddress" tcpPort:@"port"]` 
2. Application is now frozen.

##### Expected Behavior
Application to still be usable, perhaps with an error printed to the log pertaining to this incorrect IP Address.

##### Observed Behavior
Application becomes unusable.

* iOS Version: 9.2.1
* SDL iOS Version: 4.0.1
* Testing Against: Simulator & iPhone 5 & iPhone 6

asm09fsu:
Fixes #356

This PR is ready for review.

This PR makes no API changes.

### Summary
Adds the ability for us to check the provided hostname using `SystemConfiguration`'s `SCNetworkReachability` API. If the hostname is invalid, we return back -1 (error) instead of still attempting to connect, which inevitably causes the main thread to lock.

##### Bug Fixes
* Fix issue with SDLTCPTransport attempting to connect to an IP Address that is invalid.

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLTCPTransport.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLTCPTransport.m
@@ -14,6 +14,7 @@
 #import <sys/wait.h>
 #import <netinet/in.h>
 #import <netdb.h>
+#import <SystemConfiguration/SystemConfiguration.h>
 
 
 // C function forward declarations.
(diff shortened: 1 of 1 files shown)


joeljfischer:
:+1:

PR #360, opened 2016-02-25
Fix BulkData Service

### LINKED ISSUE #359: Sending binary data is not properly handled
### Bug Report
Currently sending bulk data occurs on the `RPC` service, this should be occurring on the `Bulk Data` service

##### Reproduction Steps
1. Create an RPC with some binary data
1. Send RPC using `SDLProxy sendRPC:`

##### Expected Behavior
RPC should be sent with service type `BulkData`.

##### Observed Behavior
RPC is sent with service type `RPC`

* iOS Version: N/A
* SDL iOS Version: 4.0.1
* Testing Against: N/A

TODO

joeljfischer:
Fixes #359 

### Risk

### Testing Plan
A unit test will be added to check that BulkData and RPC service messages correctly send through SDLProtocol

### Summary
This PR fixes the iOS library by allowing it to correctly send and receive hybrid service / bulk data messages on the bulk data service. This issue was not caught previously because both sides of the transaction were sending these messages incorrectly on the RPC service.

##### Bug Fixes
* SDLProtocol now sets service type to BulkData when there is binary data in the RPC message
* SDLProtocolMessage should now print correctly to the console if it is a bulk data message
* An RPC payload will now correctly convert to a dictionary if the service type is bulk data

### Tasks Remaining:
- [x] Add test case for BulkData and RPC services when sending an RPC through `SDLProtocol`


joeljfischer:
@mrapitis This is ready to test

mrapitis:
Thanks, we will plan to start testing this PR early next week, will keep you posted on results.

joeljfischer:
@mrapitis do you have updates on this?

mrapitis:
Yes, the team ran a suite of tests against the various platforms with this PR and everything checked out ok.  From a functional standpoint I think we are in good shape.  Not sure if we need someone to perform a quick review of the code though.

PR #363, opened 2016-02-28
Invalid TCP Address Error Log

### LINKED ISSUE #362: SDLTCPTransport does not give an error if unreachable IP address is used.
### Bug Report
Currently with the new bug fix to help with the issue of unreachable IP addresses returning back a failure when establishing a connection, we do not let the develop know that this is the issue why the server is not connecting. We don't give a reason at all, actually.

##### Reproduction Steps
1. Initialize a proxy using a non-existent TCP/IP address
1. Look at logs in Xcode.
1. You will see `Server Not Ready, Connection Failed` but nothing relating to the invalid IP Address.

##### Expected Behavior
An error of something like `Invalid IP address of <ip_address>. Cancelling connection.`.

##### Observed Behavior
No such log.

* iOS Version: 9.2.1
* SDL iOS Version: 4.0.2
* Testing Against: 4.0.2

asm09fsu:
Fixes #362 

### Risk

### Summary
Added an additional debug log when an invalid IP address is used when connecting over TCP/IP.

##### Enchancements
* Added an additional debug log when an invalid IP address is used when connecting over TCP/IP

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLTCPTransport.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLTCPTransport.m
@@ -132,7 +132,7 @@ int call_socket(const char *hostname, const char *port) {
     // check if hostname address is valid before we attempt to connect.
     SCNetworkReachabilityRef reachability = SCNetworkReachabilityCreateWithAddress(kCFAllocatorDefault, (const struct sockaddr *)hostname);
     if (reachability == NULL) {
-        NSString* hostnameString =  [NSString stringWithCString:hostname encoding:NSUTF8StringEncoding];
+        NSString* hostnameString =  [NSString stringWithUTF8String:hostname];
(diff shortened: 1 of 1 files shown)


SmartDeviceLink-iOS/SmartDeviceLink/SDLTCPTransport.m
```diff
@@ -132,6 +132,9 @@ int call_socket(const char *hostname, const char *port) {
     // check if hostname address is valid before we attempt to connect.
     SCNetworkReachabilityRef reachability = SCNetworkReachabilityCreateWithAddress(kCFAllocatorDefault, (const struct sockaddr *)hostname);
     if (reachability == NULL) {
+        NSString* hostnameString =  [NSString stringWithCString:hostname encoding:NSUTF8StringEncoding];
```
joeljfischer: you can just use `[NSString stringWithUTF8String:]`

asm09fsu: yeah, yeah, okay.

PR #367, opened 2016-03-03
Instead of checking the reachability object for nullability, we need …

### LINKED ISSUE #365: Reachability for TCP is broken
### Bug Report
When attempting to start a TCP connection, the Reachability check will fail even if the IP address is valid

##### Reproduction Steps
1. Start a proxy session with TCP
1. Use a valid IP address

##### Expected Behavior
The TCP code passes the reachability check

##### Observed Behavior
The TCP code does not pass the reachability check

* iOS Version: 9.2.1
* SDL iOS Version: 4.0.2
* Testing Against: SDL Core

N/A

asm09fsu:
Fixes #365 

### Risk

### Testing Plan
No way (to my knowledge) to unit test specific network connectivity.

### Summary
Using SCNetworkReachabilityCreateWithName instead of SCNetworkReachabilityCreateWithAddress and checking it's reachability flags.

##### Bug Fixes
* Fixes SDLTCPTransport to properly connect and also present error to log when incorrect IP address used.

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLTCPTransport.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLTCPTransport.m
@@ -130,14 +130,16 @@ int call_socket(const char *hostname, const char *port) {
     }
     
     // check if hostname address is valid before we attempt to connect.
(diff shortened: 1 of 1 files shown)


joeljfischer:
:-1:

PR #375, opened 2016-03-10
Check OS version in LAUNCH_APP

### LINKED ISSUE #371: Add LAUNCH_APP System Request
### Enhancement

Add the LAUNCH_APP System Request which passes a URL to be opened using the system call `openURL:`.

justinjdickow:
Fixes #371 

### Risk

### Testing Plan
@madhuy please test with iOS < 9.0 making sure launching an app that is not installed does nothing

### Summary
Checked OS version before using openURL, because it crashes without checking canOpenURL in iOS < 9.0


SmartDeviceLink-iOS/SmartDeviceLink/SDLProxy.m
```diff
@@ -362,7 +362,13 @@ - (void)handleSystemRequestLaunchApp:(SDLOnSystemRequest *)request {
         [SDLDebugTool logInfo:[NSString stringWithFormat:@"Launch App failure: invalid URL sent from module: %@", request.url] withType:SDLDebugType_RPC toOutput:SDLDebugOutput_All toGroup:self.debugConsoleGroupName];
         return;
     }
-    
+    // If system version is less than 9.0 http://stackoverflow.com/a/5337804/1370927 
+    if ([[[UIDevice currentDevice] systemVersion] compare:@"9.0" options:NSNumericSearch] == NSOrderedAscending) {
+        // Return early if we can't openURL because openURL will crash instead of fail silently in < 9.0
+        if ([[UIApplication sharedApplication] canOpenURL:URLScheme] == NO) {
```
joeljfischer: We shouldn't directly compare with BOOLs, so this should just use `!`.

PR #376, opened 2016-03-10
Remove a session id after the service is disconnected

### LINKED ISSUE #349: Session IDs don't get cleared when a service ends
Correct me if I'm wrong, but it looks like we're just storing them and not clearing them. 

https://github.com/smartdevicelink/sdl_ios/blob/master/SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m#L327

joeljfischer:
Fixes #349 

### Risk

### Testing Plan
This change will not be tested as it does not change any public API calls that can be tested

### Summary
This PR removes session ids from the `sessionId` dictionary within the `SDLProtocol` class when an `endServiceACK` is received.

##### Bug Fixes
* Properly remove session ids from tracking when their service has ended.

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLProtocol.m
@@ -62,9 +62,7 @@
 }
 
 - (void)sdl_removeSessionIdForServiceType:(SDLServiceType)serviceType {
-    if (_sessionIDs[@(serviceType)] != nil) {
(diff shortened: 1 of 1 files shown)


joeljfischer:
@justinjdickow This is changing some of your code. Please review.

joeljfischer:
@asm09fsu Since Justin's out for the week, can you double check this?

asm09fsu:
Aside from above comment, looks good.

PR #378, opened 2016-03-10
Sending application state to SDL

### LINKED ISSUE #377: Add foreground / background tracking of host app for LAUNCH_APP
### Enhancement
Add foreground / background tracking to the SDL Library to facilitate SDL Core calling System Request LAUNCH_APP on the correct connected app. The library then needs to send a notification to the head unit of its current state.

* There should be a test case to test the notification sent after moving between foreground and background.
* There should be a test case asserting that the proper notifications fire when the app moves foreground to background (if possible).

justinjdickow:
Fixes #377 

### Risk

### Testing Plan
Tested registering with SDL >= 4 (SDL Receives FULL) and backgrounding the connected app (SDL Receives BACKGROUND)

### Summary
Added a method to send the current application case to SDL in the form of an HMI Status notification

##### Enchancements
* Tells SDL the application state when registering as well as when entering foreground and background application states

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLProxy.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLProxy.m
@@ -130,6 +130,33 @@ const int POLICIES_CORRELATION_ID = 65535;
     return [self.mutableProxyListeners copy];
 }
 
+#pragma mark - Methods
+
+- (void)sendMobileHMIState {
+    UIApplicationState appState = [UIApplication sharedApplication].applicationState;
+    SDLOnHMIStatus *HMIStatusRPC = [[SDLOnHMIStatus alloc] init];
+    
+    HMIStatusRPC.audioStreamingState = [SDLAudioStreamingState NOT_AUDIBLE];
+    HMIStatusRPC.systemContext = [SDLSystemContext MAIN];
+    
+    switch (appState) {
+        case UIApplicationStateActive: {
+            HMIStatusRPC.hmiLevel = [SDLHMILevel FULL];
+        } break;
+        case UIApplicationStateBackground: // Fallthrough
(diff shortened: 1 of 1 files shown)


SmartDeviceLink-iOS/SmartDeviceLink/SDLProxy.m
```diff
@@ -295,6 +322,12 @@ - (void)handleRegisterAppInterfaceResponse:(SDLRPCResponse *)response {
     //Print Proxy Version To Console
     NSString *logMessage = [NSString stringWithFormat:@"Framework Version: %@", self.proxyVersion];
     [SDLDebugTool logInfo:logMessage withType:SDLDebugType_RPC toOutput:SDLDebugOutput_All toGroup:self.debugConsoleGroupName];
+    if ([SDLGlobals globals].protocolVersion >= 4) {
```
joeljfischer: Should this be 3 now?

justinjdickow: Not as far as I know, is there a reason why you think it should be lower now?

joeljfischer: I don't think anything is actually in 4, is it? I thought we scrapped 4 entirely and said 4 will be negotiable MTU.

justinjdickow: It should remain 4. SDL tracks these requests and v4 apps so it knows what apps it can send LAUNCH_APP and QUERY_APPS system requests: https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/commands/mobile/on_hmi_status_notification_from_mobile.cc#L81-L113

PR #379, opened 2016-03-15
Restructure of heartbeat to only respond to heartbeats with acknowledgement

### LINKED ISSUE #368: Investigate Removing Heartbeat
### Bug Report
Heartbeat has been nothing but an enormous headache on iOS. It hasn't solved any problems, but seems to have generated several (#341, #333). Furthermore, this seems unnecessary because iOS will tell us if a transport disconnects on iAP or over sockets. Heartbeat does not seem to gain us anything. We need to discuss if its removal will cause any other problems, and any advantages it offers that we are not seeing.

We must also assess any risk, and whether heartbeat gains us anything, on Android. I do not believe we should implement on iOS solely for "feature parity", but if we turn heartbeat off entirely or if we make it optional would depend on what is needed on Android.

I consider this high priority because of the unexpected major issues we are seeing as a result of the heartbeat implementation.

A POC or unit test cases would be welcome.

### Expected Behavior
The issues referenced above do not occur.

### Observed Behavior
The issues referenced above occur.

* iOS Version: 9.2.1
* SDL iOS Version: 4.0.3
* Testing Against: SDL Core, Ford TDK Gen 3

This would be welcome if anyone wishes to provide it.

@Heath-FoMoCo @khburdette @asm09fsu @tpulatha @mrapitis @justinjdickow @JackLivio @joeygrover

asm09fsu:
Fixes #368 

### Risk

### Testing Plan
Testing plan will require making sure changes relating removing heartbeat do not in-advertantly affect other components.

### Summary
Removal of in-proxy sending of heartbeats. SDL for iOS will only respond to heartbeats from Core.

##### Bug Fixes
* Heartbeats are no longer actively sent, but only ACKed. This fixes the issues we've faced regarding timers and backgrounding.

### Tasks Remaining:
- [x] Validation that changes relating to heartbeat removal do not break current messaging.
- [x] Validate that `[SDLProtocolReceivedMessageRouter dispatchControlMessage:]` still works as intended.

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLAbstractProtocol.h
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLAbstractProtocol.h
@@ -20,7 +20,7 @@
 - (void)sendEndSessionWithType:(SDLServiceType)serviceType;
 - (void)sendRPC:(SDLRPCMessage *)message;
 - (void)sendRPCRequest:(SDLRPCRequest *)rpcRequest __deprecated_msg(("Use sendRPC: instead"));
-- (void)sendHeartbeat;
+- (void)sendHeartbeat __deprecated_msg("Heartbeat is no longer used.");
 - (void)sendRawDataStream:(NSInputStream *)inputStream withServiceType:(SDLServiceType)serviceType;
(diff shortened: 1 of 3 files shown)


asm09fsu:
@joeljfischer please verify.

PR #389, opened 2016-04-22
Init rpc message with mutable dictionary

### LINKED ISSUE #388: SDL 4.1.0 not working on SYNC1 (Protocol version 1)
### Bug Report
On older SYNC1 (US only) head units apps will not be able to communicate anymore. Error "Proxy: Failed to handle protocol message -[__NSCFDictionary setObject:forKey:]: mutating method sent to immutable object'" occurs. This happens in -[SDLRPCResponse initWithDictionary:] line 31 because initWithDictionary: was called with an immutable dictionary object.

The root cause is the difference between SDLV1 and SDLV2ProtocolMessage where V2 creates a mutable dictionary while V1 creates an immutable one (see line 20 in V1 file). This immutable dictionary is passed to the proxy object into -[SDLProxy handleRPCDictionary:] where line 262 calls initWithDictionary: of the response. This call requires a mutable dictionary but a immutable one is sent to the request about the be created. No error/warning is shown in Xcode because a selector exist for "initWithDictionary:(NSDictionary *)".

Proposed solution:
Either create a mutableCopy in V1ProtocolMessage (line 20). This is not really clean because the assumption of an NSMutableDictionary object behind a NSDictionary type still remains.
Or create a mutableCopy in RPCStruct (line 16). Still an immutable object is passed through all the subclasses.
Or create a mutableCopy in SDLProxy (line 262). This is the moment from when SDL wants a mutable object.

##### Reproduction Steps
1. Install HelloSDL with SDL 4.1.0
2. Start app on the iPhone
3. Connect iPhone over USB
4. Wait up to 10 seconds until app send request to register

##### Expected Behavior
- App should receive RegisterAppInterfaceResponse
- App should be usable over SYNC1

##### Observed Behavior
- App does not receive any resposne or notification (at all).
- App does not offer any functionality over SYNC1

* iPhone 5 used
* iOS Version: 8.4.1
* SDL iOS Version: 4.1.0
* Testing Against: SYNC1 bench

HelloSDL with SDL 4.1.0 used

joeljfischer:
Fixes #388

### Risk

##### Bug Fixes
* Change an immutable dict pass to a mutable dict

### DIFF
--- a/SmartDeviceLink-iOS/SmartDeviceLink/SDLProxy.m
+++ b/SmartDeviceLink-iOS/SmartDeviceLink/SDLProxy.m
@@ -259,7 +259,7 @@ const int POLICIES_CORRELATION_ID = 65535;
     
     // From the function name, create the corresponding RPCObject and initialize it
     NSString *functionClassName = [NSString stringWithFormat:@"SDL%@", functionName];
-    SDLRPCMessage *newMessage = [[NSClassFromString(functionClassName) alloc] initWithDictionary:dict];
+    SDLRPCMessage *newMessage = [[NSClassFromString(functionClassName) alloc] initWithDictionary:[dict mutableCopy]];
     
     // Log the RPC message
     NSString *logMessage = [NSString stringWithFormat:@"%@", newMessage];


kshala-ford:
"lgtm"

PR #404, opened 2016-05-25
Shift files from SmartDeviceLink-iOS subdirectory to root

### LINKED ISSUE #403: Shift files out of SmartDeviceLink-iOS/ subdirectory
Currently, we basically have two root directories. The actual root directory, which contains some files, and the `SmartDeviceLink-iOS/` subdirectory, which is the project's root directory. This causes some oddities, inconsistencies, and errors. Most carthage builds have to explicitly specify the subdirectory, for example. Jazzy documentation generation doesn't allow for a subdirectory to be specified for source files AFAIK, because they assume you set your project up correctly. We should shift the files in the `SmartDeviceLink-iOS` directory into the root.

Unfortunately, this will cause headaches with other branches because hundreds of files are being renamed.

joeljfischer:
Fixes #403

### Risk

### Testing Plan
N/A

### Summary
This PR shifts all files from the SmartDeviceLink-iOS subdirectory to the root directory. See #403 for information on why this is necessary.

##### Other
* Move project files and folders into root git directory


joeljfischer:
@Heath-FoMoCo @asm09fsu I need to know how this might impact any build scripts you guys are using. The fix wouldn't be too bad, just changing the path to remove the subdirectory...

asm09fsu:
@joeljfischer This will impact our build scripts, but would only require a quick fix. We'll apply the fix after this is merged.

PR #405, opened 2016-05-25
Add support for Jazz Hands

jemerick:
Adds support for [Jazzy](https://github.com/realm/jazzy) to build docs using a custom theme that will later be pulled into the documentation portal.

### Installation

``` bash
[sudo] gem install jazzy
```

### Risk
This PR makes **no** API changes, just adds a build step and requires jazzy to be installed.

### Testing Plan
1. Install Jazzy
2. Build. Ensure Docs are updated to reflect any doc changes.

### Summary
Add a build step for jazzy using a custom theme and include the docs in the repo.


joeljfischer:
I've been thinking about this some more. I don't think we want jazzy to build on every SDL build, because that would extend the build times dramatically. Second, building SDL is not currently building the docs for me. I think instead there should be a scripts folder that contains a `build-documentation` script that can be run to generate the documentation easily and integrated into a release script eventually as well. This would mean that random branches wouldn't necessarily have the up to date docs, however I don't want to take the fairly quick rebuild times and extend them to > 15 seconds. That would just kill productivity. In the future, we at livio could build a commit bot that would run the documentation parser and modify commits to include it since github just released that capability.

joeljfischer:
We could also expand Travis to build and upload these for us as well?

joeljfischer:
Closed in favor of #417

PR #400, opened 2016-05-17
Video Encoder Support for Dynamic Screen Sizes

asm09fsu:
Fixes #399 

This PR is **ready** for review.

### Risk
This PR makes **minor** API changes.

### Testing Plan
We can test by starting a video session and validating that the encoder is in fact using the screen dimensions returned back from a register app interface response.

### Summary
Added support for `SDLStreamingMediaManager`'s initializer to use the screen dimensions returned from a successful RegisterAppInterface.

##### Enchancements
* Added support for `SDLStreamingMediaManager`'s initializer to use the screen dimensions returned from a successful RegisterAppInterface.

PR #406, opened 2016-05-26
Custom video encoder settings and dynamic screen sizes

### LINKED ISSUE #399: Video Encoder Should Not Use Constant Values
### Bug Report
Currently `SDLStreamingMediaManager`'s video encoder uses static values for width and height. We should be using values from the RegisterAppInterface Response.

##### Expected Behavior
Video Encoder uses values from the RegisterAppInterface Response when building encoder.

##### Observed Behavior
Video Encoder uses static values when building encoder.

* iOS Version: 9.3.2
* SDL iOS Version: 4.1.1

### LINKED ISSUE #401: Allow for Developer to Set Video Encoder Settings
### Bug Report
I propose we allow developers to override video encoder settings if they chose to do so. This is because not every project is the same, and some developers may require different video encoder settings (although the majority probably will be fine with our default settings).

* iOS Version: 9.3.2
* SDL iOS Version: 4.1.2

asm09fsu:
Fixes  #399 and #401 

### Risk

### Testing Plan
For #399 
We can test by starting a video session and validating that the encoder is in fact using the screen dimensions returned back from a register app interface response.

For #401 
A valid testing plan would make sure that we cannot change the video encoder's settings while a video stream is connected, as well as validating that settings `SDLStreamingMediaManager`'s new `videoEncoderSettings` property to nil results in using the default settings.

### Summary
For #399 
Added support for SDLStreamingMediaManager's initializer to use the screen dimensions returned from a successful RegisterAppInterface.

For #401
This PR allows a developer to specify custom video encoder settings if they believe they have found a better means of streaming with their current environment. 

##### Enhancements
* Added ability for developer-specified `VTCompressionSession` properties.
* Added support for providing SDLStreamingMediaManager with display capabilities returned from a successful RegisterAppInterface.

##### Bug Fixes
* Fixed an issue relating to setting the return error when configuring the video encoder.


SmartDeviceLink-iOS/SmartDeviceLink/SDLStreamingMediaManager.m
```diff
@@ -229,83 +251,52 @@ - (BOOL)sdl_configureVideoEncoderWithError:(NSError *__autoreleasing *)error {
...
-    CFNumberRef bitRateNumRef = CFNumberCreate(NULL, kCFNumberSInt32Type, &bitRate);
-    if (bitRateNumRef == NULL) {
-        // TODO: Log & end session
-        if (*error != nil) {
-            *error = [NSError errorWithDomain:SDLErrorDomainStreamingMediaVideo code:SDLStreamingVideoErrorConfigurationAllocationFailure userInfo:nil];
-        }
-
-        return NO;
+    
+    if (self.videoEncoderSettings == nil) {
```
joeljfischer: If this is how it's being done, then the video encoder settings dict should be null_resettable, not nullable, and the api should be redesigned so that it's initialized on `init`, maybe an `initWithEncoderSettings` if possible. For testability, as many dependencies as possible should be injected at init time.

SmartDeviceLink-iOS/SmartDeviceLink/SDLStreamingMediaManager.m
```diff
@@ -127,6 +127,28 @@ - (BOOL)sendAudioData:(NSData *)pcmAudioData {
...
+    
+    _videoEncoderSettings = videoEncoderSettings;
+}
+
+- (NSDictionary*)defaultVideoEncoderSettings {
+    static NSDictionary* defaultVideoEncoderSettings = nil;
+    if (defaultVideoEncoderSettings == nil) {
+        defaultVideoEncoderSettings = @{
+                                        (__bridge NSString*)kVTCompressionPropertyKey_ProfileLevel : (__bridge NSString*)kVTProfileLevel_H264_Baseline_AutoLevel,
+                                        (__bridge NSString*)kVTCompressionPropertyKey_RealTime : @(YES)
```
joeljfischer: `@(YES)` can be written as `@YES`

SmartDeviceLink-iOS/SmartDeviceLink/SDLStreamingMediaManager.m
asm09fsu: initWithEncoderSettings would collide with other fixes coming down the pipeline (such as dynamic encoder dimensions). The thought here is that so long as the stream isn't open, we can modify the existing `SDLStreamingMediaManager` object, instead of having to specify it via the proxy (since the developer never calls the init). We can definitely add it in the init.

joeljfischer: We need to decide how this init will work with #400 also making a change / addition to init.

asm09fsu:
@joeljfischer Issue #400 has been added to this PR as well, with no modification of the API's init methods.

SmartDeviceLink/SDLProxy.m
```diff
@@ -166,7 +169,11 @@ - (NSString *)proxyVersion {
 
 - (SDLStreamingMediaManager *)streamingMediaManager {
     if (_streamingMediaManager == nil) {
+        if (self.displayCapabilities == nil) {
+            @throw [NSException exceptionWithName:NSInvalidArgumentException reason:@"SDLStreamingMediaManager must be accessed only after a successful RegisterAppInterfaceResponse" userInfo:nil];
+        }
         _streamingMediaManager = [[SDLStreamingMediaManager alloc] initWithProtocol:self.protocol];
```
joeljfischer: Can we deprecate `initWithProtocol:` and add `initWithProtocol:displayCapabilities:`? The displayCapabilities property on SMM will have to stay nullable for now, but that would clarify the actual dependencies.

SmartDeviceLink/SDLStreamingMediaManager.m
```diff
@@ -224,36 +287,52 @@ - (BOOL)sdl_configureVideoEncoderWithError:(NSError *__autoreleasing *)error {
...
 
     // Create a compression session
-    // TODO (Joel F.)[2015-08-18]: Dimensions should be from the Head Unit
-    status = VTCompressionSessionCreate(NULL, 640, 480, kCMVideoCodecType_H264, NULL, NULL, NULL, &sdl_videoEncoderOutputCallback, (__bridge void *)self, &_compressionSession);
+    status = VTCompressionSessionCreate(NULL, self.screenSize.width, self.screenSize.height, kCMVideoCodecType_H264, NULL, NULL, NULL, &sdl_videoEncoderOutputCallback, (__bridge void *)self, &_compressionSession);
 
     if (status != noErr) {
         // TODO: Log the error
-        if (*error != nil) {
+        if (*error == nil) {
```
joeljfischer: These were correct originally. They're checking that the passed NSError address can be dereferenced, and later retrieved by the developer. See "Generating your own errors" at https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/ErrorHandling/ErrorHandling.html

SmartDeviceLink/SDLStreamingMediaManager.m
```diff
@@ -224,36 +287,52 @@ - (BOOL)sdl_configureVideoEncoderWithError:(NSError *__autoreleasing *)error {
...
+    }
+    CFRelease(supportedProperties);
+    
+    // Populate the video encoder settings from provided dictionary.
+    for (NSString* key in self.videoEncoderSettings.allKeys) {
+        id value = self.videoEncoderSettings[key];
+        
+        status = VTSessionSetProperty(self.compressionSession, (__bridge CFStringRef)key, (__bridge CFTypeRef)value);
+        if (status != noErr) {
+            if (*error == nil) {
```
joeljfischer: See above

joeljfischer:
LGTM

joeljfischer:
This was merged

PR #409, opened 2016-05-26
Improved Video Encoder Settings

### LINKED ISSUE #408: Update Video Encoder Settings for Better Quality Streaming
### Bug Report
The current video encoder settings are not ideal for streaming. I believe we can find an ideal set of properties for increasing performance of video streaming.

* SDL iOS Version: 4.1.2

asm09fsu:
Fixes #408 

### Risk

### Summary
This PR removes unnecessary video encoder properties to improve the quality of video streaming.

##### Enchancements
* Removes hindering video encoder settings


joeljfischer:
This is also adding the xctool-args stuff, which shouldn't be in this PR. They're in develop, but we can't just sneak those in like that. That would have to be a separate PR.

PR #423, opened 2016-06-15
Touch Manager

### LINKED ISSUE #402: Add Touch Manager for improved onTouchEvent Handling
### Bug Report
After dealing with using the onTouchEvent callback, I propose we design an architecture for having a touch manager to give callbacks that allow developers to know if certain touch events are happening. Currently the following are thought of as important:
- Taps
  - Single Tap
  - Double Tap
- Pinch
  - Pinch Did Start
  - Pinch Moved
  - Pinch Did End
- Pan
  - Pan Did Start
  - Pan Moved
  - Pan Did End

The point of this manager is to allow developers to use their time dealing with these touch events, instead of building logic to figure out when these touch events occur.

asm09fsu:
Fixes #402 

### Risk

### Testing Plan
Tests need to be written to make sure that we are correctly calculating pinch, pan, single and double touches.

### Summary
This PR introduces the Touch Manager, which is, simply, a wrapper around the `onOnTouchEvent` notification that we receive from Core. TM handles the logic for determining when a single or double tap has occurred, as well as panning and pinching. Some of these properties are changeable for the developer, but the main objective here is to develop a standard implementation so developers can focus on creating applications, not on logic to handle touching events. 
**The Touch Manager is only valid for applications utilizing video streaming.** 
> For information on video streaming for mobile navigation, look at our reference [here](https://smartdevicelink.com/guides/iOS/mobile-navigation/).

There are a few classes/structs/headers that have been added:
- SDLTouchManager.h/m
  - All logic for interpreting an `onOnTouchEvent` notification, and calling the `SDLTouchManagerListener`'s protocol.
- SDLTouchManagerListener.h
  - The protocol callbacks used for informing a developer of touch events.
- SDLTouch.h/m
  - Struct for storing `SDLTouchEvent` in an easier to consumer manner, as well as some functions to assist in comparisons and validity.
- SDLPinchGesture.h/m
  - Struct for storing a pinch gesture, which is comprised of `SDLTouch` structs. This also contains functions for assisting in comparisons and validity.
- CGPoint_Util.h/m
  - Basic utility functions to assist in CGPoint calculations.
- dispatch_timer.h/m
  - Wrapper around dispatch queues for using a GCD timers.

##### Enhancements
* Added a touch manager for developers to easily add touch handling for video streaming applications.

### Tasks Remaining:
- Add Documentation
  - [x] SDLTouchManager
  - [x] CGPoint_Util
  - [x] SDLTouch
  - [x] SDLPinchGesture
- Add Tests
  - [x] SDLTouchManager
  - [x] CGPoint_Util
  - [x] SDLTouch
  - [x] SDLPinchGesture
  - [x] dispatch_timer


asm09fsu:
@joeljfischer start reviewing the tests please when you get a chance.

SmartDeviceLink/SDLTouchManagerListener.h
```diff
@@ -0,0 +1,25 @@
+//
+//  SDLTouchManagerListener.h
```
joeljfischer: Listener is the Java convention I believe. It should be named `-Delegate`.

asm09fsu: If that's the case, we should be modifying all other Protocols that use `-Listener`, I was simply trying to keep it consistent 👎

joeljfischer: Oh yeah, definitely. It's just a breaking change to do so.

SmartDeviceLink/SDLPinchGesture.h
```diff
@@ -0,0 +1,31 @@
+//
```
joeljfischer: All these classes need NS_ASSUME_NONNULL_BEGIN / END

SmartDeviceLink/SDLTouchManager.h
```diff
@@ -0,0 +1,33 @@
...
+#import "SDLTouchManagerListener.h"
+
+@interface SDLTouchManager : NSObject
+
+@property (nonatomic, weak) id<SDLTouchManagerListener> touchEventListener;
+
+/*
+ *  Distance between taps.
+ */
+@property (nonatomic) CGFloat tapDistanceThreshold;
```
joeljfischer: Explicitly specify as assign(?), even though it's obvious, just for consistency. Atomicity and strong/weak/assign/copy should be on every property.

SmartDeviceLink/SDLTouchManager.m
```diff
@@ -0,0 +1,241 @@
...
+#import "SDLTouchType.h"
+
+typedef NS_ENUM(NSUInteger, SDLPerformingTouchType) {
+    SDLPerformingTouchTypeNone,
+    SDLPerformingTouchTypeSingleTouch,
+    SDLPerformingTouchTypeMultiTouch,
+    SDLPerformingTouchTypePanningTouch
+};
+
+static NSUInteger const kMaximumNumberOfTouches = 2;
```
joeljfischer: No k! k? k.

SmartDeviceLink/SDLTouchManager.m
```diff
@@ -0,0 +1,241 @@
...
+
+@interface SDLTouchManager () <SDLProxyListener>
+
+@property SDLTouch previousTouch;
+
+/*
+ * Only used for caching previous single taps for double tap detection
+ */
+@property SDLTouch singleTapTouch;
```
joeljfischer: These are all atomic?

SmartDeviceLink/SDLTouchManager.m
```diff
@@ -0,0 +1,241 @@
...
+@property dispatch_source_t singleTapTimer;
+
+@property SDLPerformingTouchType performingTouchType;
+
+@end
+
+@implementation SDLTouchManager
+
+- (instancetype)init {
+    if (self = [super init]) {
```
joeljfischer:
```objc
self = [super init];
if (!self) return nil;
...
return self;
```
is preferred to avoid right shift.

SmartDeviceLink/SDLTouchManager.m
```diff
@@ -0,0 +1,241 @@
...
+        [self sdl_handleBeginTouch:touch];
+    } else if ([notification.type isEqualToEnum:SDLTouchType.MOVE]) {
+        [self sdl_handleMovedTouch:touch];
+    } else if ([notification.type isEqualToEnum:SDLTouchType.END]) {
+        [self sdl_handleEndTouch:touch];
+    }
+}
+
+#pragma mark - Private
+- (void)sdl_handleBeginTouch:(SDLTouch)touch {
```
joeljfischer: Naming should probably be `sdl_handleTouchBegin` since that helps make the other, similar method names appear more similar to the eye even though it sounds kinda weird audibly.

joeljfischer:
We already know timers don't work properly in the background. Does the touch manager explicitly handle that case? For example, it could watch for background notifications and stop timers, watch for foreground and restart, but it would also have to notify the developer that it's not going to work in the background.

asm09fsu:
Video streaming doesn't work in the background, so I don't think we really need to be concerned with the timers.

SmartDeviceLink/SDLPinchGesture.h
```diff
@@ -0,0 +1,62 @@
...
+ *  @return SDLPinchGesture
+ *      Instance of SDLPinchGesture.
+ */
+- (instancetype)initWithFirstTouch:(SDLTouch*)firstTouch secondTouch:(SDLTouch*)secondTouch;
+
+/**
+ *  @abstract
+ *      First touch of a pinch gesture.
+ */
+@property (nonatomic, copy) SDLTouch* firstTouch;
```
joeljfischer: These aren't copyable, I don't think?

asm09fsu: You're right! This was when we were using c-structs instead of objects.

joeljfischer:
I'm seeing a few test failures after the changes

joeljfischer:
LGTM

PR #426, opened 2016-07-06
Fix not handling the TCP disconnect case

### LINKED ISSUE #385: Core disconnect crash
### Bug Report
If Core is disconnected while an app is connected, that app will crash.

##### Reproduction Steps
1. Connect to SDL Core using an app running the SDL SDK using TCP
1. Close SDL Core while the app is still connected

##### Expected Behavior
The app to close its connection gracefully

##### Observed Behavior
The app crashes

* iOS Version: 9.3.0
* SDL iOS Version: 4.1.0
* Testing Against: SDL Core 4.0.0

joeljfischer:
Fixes #385

### Risk

### Testing Plan
Currently untestable

### Summary
This PR implements a case handling a TCP debug disconnect from SDL Core. This prevents a crash when the CFSocket API sends a data packet of length 0 to indicate disconnection. The architecture of proxy makes this unideal, and if WiFi ever sent into production, we'd need a retry strategy (along with a full re-architecture, scanning for ports, etc.), but for debug, it's helpful to not crash even though the developer must manually restart the connection in-app.

##### Bug Fixes
* Fix a crash caused by a SDLCore disconnect (#385)

### DIFF
--- a/SmartDeviceLink/SDLTCPTransport.m
+++ b/SmartDeviceLink/SDLTCPTransport.m
@@ -45,7 +45,7 @@ static void TCPCallback(CFSocketRef socket, CFSocketCallBackType type, CFDataRef
 
 
 - (void)connect {
-    [SDLDebugTool logInfo:@"Init" withType:SDLDebugType_Transport_TCP];
+    [SDLDebugTool logInfo:@"TCP Transport attempt connect" withType:SDLDebugType_Transport_TCP];
 
     int sock_fd = call_socket([self.hostName UTF8String], [self.portNumber UTF8String]);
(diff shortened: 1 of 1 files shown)


asm09fsu:
Are we still worried about the issues we were seeing when we do in fact issue an `onTransportDisconnect` and the proxy shutdown happening out of order talked about in issue #385?

joeljfischer:
Yes, but since this is not production code, the developer will have to manually initiate a reconnect. We don't have a retry strategy that works for TCP.

justinjdickow:
👍

PR #449, opened 2016-09-23
[SDL-0008] iOS 6 Removal

### LINKED ISSUE #419: Drop iOS 6 support
Covers [Proposal SDL-0008](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0008-ios-7-0-minimum.md).
### Feature Request

We should seek to drop iOS 6 support. Virtually no one uses it and it would allow us to drop the NSURLConnection support and simplify a lot of code with NSURLSession instead.

### LINKED ISSUE #450: iOS 7+ allows template images
## Bug Report

Once iOS 4.3.0 releases with SDLLockScreenViewController and after we move to iOS 7+ only, we can remove half of the images by using template images instead.

asm09fsu:
Fixes #419 
Fixes #450 
Proposal [SDL-0008](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0008-ios-7-0-minimum.md).

### Risk

### Testing Plan

This version needs to be tested to confirm that it does infact no longer work with iOS 6. We also need to see if there are any other improvements (aside from `NSURLSession`) that we can use now that we are using iOS 7.
### Summary

Minimum supported iOS version increased from 6 to 7. We also removed `SDLURLSession` in favor of Apple's `NSURLSession`.
##### Breaking Changes
- Increased minimum supported version from iOS 6 to iOS 7.
- Removed `SDLURLSession` in favor of `NSURLSession`.


joeljfischer:
See https://github.com/smartdevicelink/sdl_ios/issues/450 regarding another iOS 7+ improvement.

SmartDeviceLinkTests/UtilitiesSpecs/HTTP Connection/SDLURLRequestTaskSpec.m
```diff
@@ -1,140 +0,0 @@
-#import <Quick/Quick.h>
-#import <Nimble/Nimble.h>
-#import <OHHTTPStubs/OHHTTPStubs.h>
-
-#import "SDLURLRequestTask.h"
-
-
-QuickSpecBegin(SDLURLRequestTaskSpec)
```
joeljfischer: We probably need replacement tests of some sort to make sure that the URL requests are set up properly? We don't need to test actual functionality, I don't think.

asm09fsu: We shouldn't need to do this, as we are using them for different URL request types (POST, GET). Do we really need to test to make sure that an Apple-built framework works?

PR #471, opened 2016-11-30
Updated property types to better reflect spec.

### LINKED ISSUE #469: SDLPerformInteraction's initializer sets timeout to incorrect type
### Bug Report
In 4.3, initializers were introduced for all RPCs. It seems as though `SDLPerformInteraction`'s `timeout` parameter uses `UInt16` instead of `UInt32`.

##### Expected Behavior
Compiler should not give warning when using `100000` as a timeout value.

##### Observed Behavior
Compiler gives warning

* SDL iOS Version: 4.4.1

asm09fsu:
Fixes #469 

### Risk

### Summary
This PR changes the types of parameters in RPC's initializers to match those in [Mobile_API.xml](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/interfaces/MOBILE_API.xml). There is also an initializer added to `SDLPerformInteraction` that includes all of the parameters, as this was missing.

##### Bug Fixes
* Fixed issues relating to incorrect types being used in some initializer's

### DIFF
--- a/SmartDeviceLink/SDLAddSubMenu.h
+++ b/SmartDeviceLink/SDLAddSubMenu.h
@@ -31,9 +31,9 @@
  */
 - (instancetype)initWithDictionary:(NSMutableDictionary *)dict;
 
-- (instancetype)initWithId:(UInt8)menuId menuName:(NSString *)menuName;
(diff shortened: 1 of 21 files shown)


joeljfischer:
These are major changes

PR #474, opened 2016-12-01
Allow SecurityManager's AppId to be set earlier on in lifecycle

### LINKED ISSUE #470: Security Manager should have appId set as soon as possible.
### Bug Report
Currently, there is an issue with User Experience of Mobile Navigation when utilizing security managers, in that we require a certificate download when the session is started. We believe this issue can be eliminated by providing the `appId` earlier on in the app's lifecycle (after RAI), and downloading any certificates before the TLS encryption is started.

* SDL iOS Version: 4.4.1

asm09fsu:
Fixes #470 

### Risk

### Testing Plan
This PR does not impact any existing security managers.

### Summary
On a successful RAI, the security manager's appId is set if the selector is available.

##### Enchancements
* Setting security manager's appId on successful register app interface, instead of on session start.

### DIFF
--- a/SmartDeviceLink/SDLProxy.m
+++ b/SmartDeviceLink/SDLProxy.m
@@ -386,6 +386,9 @@ const int POLICIES_CORRELATION_ID = 65535;
         _streamingMediaManager.displayCapabilties = registerResponse.displayCapabilities;
     }
     self.protocol.securityManager = [self securityManagerForMake:registerResponse.vehicleType.make];
+    if (self.protocol.securityManager) {
+        self.protocol.securityManager.appId = self.appId;
+    }
 
     if ([SDLGlobals globals].protocolVersion >= 4) {
         [self sendMobileHMIState];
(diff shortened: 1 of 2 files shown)


SmartDeviceLink/SDLSecurityType.h
```diff
@@ -12,6 +12,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @protocol SDLSecurityType <NSObject>
 
+@property (copy, nonatomic) NSString* appId;
```
joeljfischer: This also needs to be changed in https://github.com/smartdevicelink/sdl_security_ios

SmartDeviceLink/SDLProxy.m
```diff
@@ -386,6 +386,10 @@ - (void)handleRegisterAppInterfaceResponse:(SDLRPCResponse *)response {
         _streamingMediaManager.displayCapabilties = registerResponse.displayCapabilities;
     }
     self.protocol.securityManager = [self securityManagerForMake:registerResponse.vehicleType.make];
+    if (self.protocol.securityManager
+        && [self.protocol.securityManager respondsToSelector:@selector(setAppId:)]) {
```
joeljfischer: Have you already tested that this fixes the issue?

asm09fsu: yes

SmartDeviceLink/SDLSecurityType.h
asm09fsu: https://github.com/smartdevicelink/sdl_security_ios/pull/1

PR #500, opened 2017-01-19
Fix Sending Nil Request Causing Crash.

### LINKED ISSUE #499: SDLManager will crash if passed a nil request
### Bug Report
`SDLManager`, and subsequently it's backing `SDLLifecycleManager` do not check against nil requests being sent. Because of this, it causes an uncaught exception in `SDLResponseDispatcher` due to the fact that the dispatcher uses the request's correlationId as a key in it's backing hash map.

To solve this, we need to check against nil requests in the lifecycle manager. If we come across one, we should throw an error and return.

##### Reproduction Steps
1. Connect to Core.
2. Send nil request.
3. Crash

##### Expected Behavior
App does not crash when passed a nil request.

##### Observed Behavior
App crashes when passed a nil request.

* SDL iOS Version: 4.5

```objc
SDLShow* show = nil;
[self.sdlManager sendRequest:show];
```

where `sdlManager` is your `SDLManager` object.

asm09fsu:
Fixes #499 

### Risk

### Testing Plan
There is a unit test that has been added to test sending a nil request in `SDLLifecycleManagerSpec` 

### Summary
Added check for sending a nil request.

##### Bug Fixes
* Fixed issue with sending a nil request crashes due to Invalid Exception.

### Notes
Because an app should *never* send a nil request, we throw an exception. Should we instead be just firing the response handler (if nonnull)? There would be a possibility that a developer passes a nil response handler, and in that case they would never know this error was occuring if not for a throw.

### DIFF
--- a/SmartDeviceLink/SDLLifecycleManager.m
+++ b/SmartDeviceLink/SDLLifecycleManager.m
@@ -367,6 +367,15 @@ SDLLifecycleState *const SDLLifecycleStateReady = @"Ready";
 - (void)sdl_sendRequest:(SDLRPCRequest *)request withResponseHandler:(nullable SDLResponseHandler)handler {
     // We will allow things to be sent in a "SDLLifeCycleStateTransportConnected" state in the private method, but block it in the public method sendRequest:withCompletionHandler: so that the lifecycle manager can complete its setup without being bothered by developer error
 
+    // If, for some reason, the request is nil we should error out.
+    if (!request) {
+        [SDLDebugTool logInfo:@"Attempting to send a nil request, request not sent."];
+        if (handler) {
+            handler(nil, nil, [NSError sdl_lifecycle_rpcErrorWithDescription:@"Nil Request" andReason:@"Request must not be nil"]);
+        }
+        return;
+    }
+    
     // Add a correlation ID to the request
     NSNumber *corrID = [self sdl_getNextCorrelationId];
     request.correlationID = corrID;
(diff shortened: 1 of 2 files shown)


SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -367,6 +367,14 @@ - (void)sendManagerRequest:(__kindof SDLRPCRequest *)request withResponseHandler
 - (void)sdl_sendRequest:(SDLRPCRequest *)request withResponseHandler:(nullable SDLResponseHandler)handler {
     // We will allow things to be sent in a "SDLLifeCycleStateTransportConnected" state in the private method, but block it in the public method sendRequest:withCompletionHandler: so that the lifecycle manager can complete its setup without being bothered by developer error
 
+    // If, for some reason, the request is nil we should error out.
+    if (!request) {
```
joeljfischer: This should use `NSParameterAssert` instead.

joeljfischer: Not sure if we also want in production for it to fail by calling the handler and returning. The assert will only apply during DEBUG. We should probably prevent the error in production?

SmartDeviceLinkTests/DevAPISpecs/SDLLifecycleManagerSpec.m
```diff
@@ -265,6 +265,12 @@ + (void)configure:(Configuration *)configuration {
                 OCMVerify([proxyMock sendRPC:[OCMArg isKindOfClass:[SDLShow class]]]);
             });
             
+            it(@"cannot send a nil RPC", ^{
+                SDLShow *testShow = nil;
+
+                expectAction(^{ [testManager sendRequest:testShow]; }).to(raiseException().named(NSInternalInconsistencyException));
```
joeljfischer: This needs to be updated with what will happen.

PR #503, opened 2017-01-19
SDLAbstractProtocol Removal

### LINKED ISSUE #217: SDLAbstractProtocol only has one subclass and could be removed
This class could be refactored so we only have the `SDLProtocol` class, there is no need for abstraction here.

asm09fsu:
Fixes #217 

### Risk

### Testing Plan
All tests from `SDLAbstractProtocolSpec` have been added to `SDLProtocolSpec`.

### Summary
Removed `SDLAbstractProtocol` and replaced it with the already existing `SDLProtocol` as the former class was basically an empty skeleton that set 2 properties and handles 3 delegate callbacks for `SDLTransportDelegate`.

##### Breaking Changes
* Removed `SDLAbstractProtocol`.


joeljfischer:
This will need to wait for 6.0 due to `SDLProxy` not being changed in 5.0. See [this issue comment](https://github.com/smartdevicelink/sdl_evolution/issues/48#issuecomment-278344863) for more information.

PR #506, opened 2017-01-19
Fixed issue relating to getting unhandled warnings from RAI Response.

### LINKED ISSUE #505: SDLManager's startWithReadyHandler returns back success of NO for some resultCodes
### Bug Report
Currently, the success that we return back within `SDLLifecycleManager`'s `startWithReadyHandler:` functions is driven by the `resultCode` of the RegisterAppInterfaceResponse. We currently also return back a success of `YES` if the `resultCode` is equivalent to `WARNINGS` or `RESUME_FAILED`. This is an issue, however, because we have many more warnings we need to handle.

The correct way to handle the success of the RAI is to look at the response's `success` property, which, as a `BOOL`, will return `YES` or `NO` and uses the `resultCode` as further information (such as a successful registration, but warnings also). 

A prime example of one we do not handle is if we are registering with an `hmiLanguage` of `EN_US`, but the head unit is actually `EN_GB`. This will return back a `resultCode` of `WRONG_LANGUAGE`, which currently the logic with `SDLLifecycleManager` returns back a success of `NO`.

For the developer, a current workaround until this issue is resolved is to instead of checking the block's return values, instead check the `registerAppInterfaceResponse` property's `success` instead.

##### Expected Behavior
SDL should return true when the response's `success` is `YES` independent of `resultCode`.

##### Observed Behavior
SDL returns false is `success` is `YES` and `resultCode` is not `SUCCESS`, `WARNINGS`, `RESUME_FAILED`.

* SDL iOS Version: 4.5.1

##### Misc
A current workaround is below:

```objc
[self.sdlManager startWithReadyHandler:^(BOOL success, NSError * _Nullable error) {        
        if (!self.sdlManager.registerResponse.success.boolValue) {
            [SDLDebugTool logFormat:@"Error Registering: %@", strongSelf.manager.registerResponse.resultCode.value];
            return;
        }
        
        // Continued post-connection setup.
    }];
```

asm09fsu:
Fixes #505 

### Risk

### Testing Plan
@joeljfischer I may need some assistance writing a unit test for this in `SDLLifecycleManagerSpec`.

### Summary
Fixed issue with startWithReadyHandler not returning success when getting warnings.

##### Bug Fixes
* Fixed issue with startWithReadyHandler not returning success when getting warnings.

### DIFF
--- a/SmartDeviceLink/SDLError.h
+++ b/SmartDeviceLink/SDLError.h
@@ -31,7 +31,6 @@ extern SDLErrorDomain *const SDLErrorDomainFileManager;
 + (NSError *)sdl_lifecycle_managersFailedToStart;
 + (NSError *)sdl_lifecycle_startedWithBadResult:(SDLResult *)result info:(NSString *)info;
 + (NSError *)sdl_lifecycle_failedWithBadResult:(SDLResult *)result info:(NSString *)info;
-+ (NSError *)sdl_lifecycle_failedWithWarningResult:(SDLResult *)result info:(NSString *)info;
 
 #pragma mark SDLFileManager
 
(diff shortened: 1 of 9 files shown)


asm09fsu:
@joeljfischer I added in an additional error code value for warnings specifically. Can you review?

joeljfischer:
@asm09fsu the reason i hadn't done that is because it means a minor api change, and this can't be a hotfix anymore.

PR #514, opened 2017-01-28
Streaming Media Manager Updates

### LINKED ISSUE #478: Update Streaming Media Manager to handle all App & HMI States
### Bug Report
This issue is combination of #430 and #435. This is also implementing all topics discussed in [this proposal](https://github.com/smartdevicelink/sdl_evolution/pull/56).

### LINKED ISSUE #583: [SDL 0033] Streaming Media Manager 5.0
Proposal: [Streaming Media Manager 5.0](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0033-streaming-media-manager-update.md)

> This proposal is geared at revamping how the Streaming Media Manager (SMM) class manages audio/video streaming. This revamp will also include lifecycle management for these streams as the overall app/proxy lifecycle changes.

Review: smartdevicelink/sdl_evolution#103

SDLC Steering Committee Decision:

> The Steering Committee has agreed to accept this proposal, with the following text removed from the “Sending video data when iPhone is moving to background” section: “Please re-open <#App Name#>”. For now, the screen will default to black, as the committee has not yet agreed upon the text that should be displayed. Toyota will submit a separate proposal for this text once they’ve confirmed their recommendation with their appropriate teams internally.

asm09fsu:
Fixes #583 

### Risk

### Testing Plan
This will be using Unit Tests to test switching from different states.

### Summary
This PR is based off the proposal [here](https://github.com/smartdevicelink/sdl_evolution/pull/56).

##### Enchancements
* Completely rewritten `SDLStreamingMediaManager` to handle all application, HMI, and streaming states. 

### Tasks Remaining:
- [x] Add Unit tests
- [x] Remove background screen string
- [x] Add logging


joeljfischer:
This will need to be updated with the result of https://github.com/smartdevicelink/sdl_ios/pull/669 and https://github.com/smartdevicelink/sdl_ios/pull/668

SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
```diff
@@ -0,0 +1,544 @@
...
+            
+            self.backgroundingPixelBuffer = backgroundingPixelBuffer;
+        }
+    }
+    
+    [[NSNotificationCenter defaultCenter] postNotificationName:SDLVideoStreamDidStartNotification object:nil];
+}
+
+- (void)didEnterStateVideoStreamShuttingDown {
+    [self.protocol endServiceWithType:SDLServiceTypeVideo];
```
joeljfischer: Should also tear down the `videoEncoder`?

SmartDeviceLink/SDLVideoEncoder.m
```diff
@@ -0,0 +1,276 @@
...
+    _compressionSession = NULL;
+    _currentFrameNumber = 0;
+    _videoEncoderSettings = properties;
+    
+    _delegate = delegate;
+    
+    OSStatus status;
+    
+    // Create a compression session
+    status = VTCompressionSessionCreate(NULL, dimensions.width, dimensions.height, kCMVideoCodecType_H264, NULL, self.sdl_pixelBufferOptions, NULL, &sdl_videoEncoderOutputCallback, (__bridge void *)self, &_compressionSession);
```
joeljfischer: This and below is *so* much to do in the `init`, it should be shifted to a `start` method (there's already a stop, which if called, means that this object becomes useless!)

SmartDeviceLink/SDLLifecycleConfiguration.h
```diff
@@ -123,6 +124,16 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (copy, nonatomic, nullable) NSArray<Class<SDLSecurityType>> *securityManagers;
 
+/**
+ *  What encryption level video/audio streaming should be. The default is SDLStreamingEncryptionFlagAuthenticateAndEncrypt.
+ */
+@property (assign, nonatomic) SDLStreamingEncryptionFlag streamingEncryption;
```
asm09fsu: The goal for SMM is to not be interacted with directly by the developer for initialization, only for observation.

SmartDeviceLink/SDLLifecycleConfiguration.h
```diff
@@ -123,6 +124,16 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+/**
+ *  Properties to use for applications that utilitze the video encoder for streaming.
+ */
+@property (copy, nonatomic, nullable) NSDictionary<NSString *, id> *videoEncoderSettings;
```
asm09fsu: The goal for SMM is to not be interacted with directly by the developer for initialization, only for observation. We want to set these settings in the lifecycle configuration so that the LCM can handle the initialization.

SmartDeviceLink/SDLVideoEncoder.m
asm09fsu: The only reason this is "so" much to do is because settings the keys is a pain to actually do. Not sure why would would need to migrate this over as this is required to user the video encoder. `SDLVideoEncoder` will only ever be used internally, not by a developer.

SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
asm09fsu: Video encoder gets stopped once the state heads to Stopped. This only occurs if there's an error or if we receive a stop session ACK.

asm09fsu:
🙏🏻

PR #530, opened 2017-02-13
In SDLProxy.m, proxy put OnTouchEvent handle in main thread.

### LINKED ISSUE #528: Put OnTouchEvent handler out of main thread
In SDLProxy.m, proxy put OnTouchEvent handle in main thread.

It is OK when HMI list control is drawn by OpenGL,  but when app uses system list view, list scrolling animation will block main thread, so app cannot handle OnTouchEvent during scrolling.

App wants to get handle out of main thread.

* iOS Version: iOS 10.2.0
* SDL iOS Version: 4.5.3

zhengzheng:
The Touch Event should not be handled in the main thread.

Fixes #528 

### Risk

### DIFF
--- a/SmartDeviceLink/SDLProxy.m
+++ b/SmartDeviceLink/SDLProxy.m
@@ -347,6 +347,10 @@ const int POLICIES_CORRELATION_ID = 65535;
         [self handleSystemRequestResponse:newMessage];
     }
 
+    if ([functionName isEqualToString:@"OnTouchEvent"]) {
+        [self handleTouchEventResponkse:newMessage];
+        return;
+    }
     // Formulate the name of the method to call and invoke the method on the delegate(s)
     NSString *handlerName = [NSString stringWithFormat:@"on%@:", functionName];
     SEL handlerSelector = NSSelectorFromString(handlerName);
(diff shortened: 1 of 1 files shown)


asm09fsu:
@zhengzheng developers should not be using the `SDLProxyListener` delegate, but instead listening to notifications (such as `SDLDidReceiveTouchEventNotification` defined [here](https://github.com/smartdevicelink/sdl_ios/blob/master/SmartDeviceLink/SDLNotificationConstants.h).) 

Does that resolve it?

tpulatha:
For additional context, please see the touch input guide here: https://smartdevicelink.com/en/guides/iOS/mobile-navigation/touch-input/

zhengzheng:
SDLDidReceiveTouchEventNotification is posted in the main thread. I can not receive the notification while the scrollView is scrolling.

zhengzheng:
@asm09fsu this is a dead lock. the scrollView is waiting for the next touch event. SDLDidReceiveTouchEventNotification is waiting for the main thread.

joeljfischer:
Hi @zhengzheng, unfortunately, we need to do significantly more investigation around this bug. This PR would make a large impact on the library and we need assess that. Furthermore, we would prefer to evaluate options that don't single out `OnTouchEvent`. We understand the issue and are actively investigating fixes. I'm leaving this PR open for now, but even if this is the direction we go, we cannot accept this PR without you signing [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform).

PR #535, opened 2017-02-15
Update structs to support being copied

### LINKED ISSUE #523: [SDL 0011] iOS Library RPCs conform to NSCopying
Proposal: [iOS Library RPCs conform to NSCopying](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0011-ios-library-conform-to-nscopying.md)

> This proposal is to make all RPC requests, responses, and structs conform to the NSCopying protocol and for all requests containing structs to hold those properties with the copy modifier. This is a minor change, as the public interfaces for the RPCs will change to support the copy method.

Review: smartdevicelink/sdl_evolution#33

Steering Committee Decision:

> This proposal has been accepted. There was discussion around the necessity of doing so, but the eventual decision was that allowing model objects to be copied is a common practice, so even if we don't immediately begin copying these objects, we should support the ability to do so.

joeljfischer:
Fixes #523 

### Risk

### Testing Plan
n/a

### Summary
This PR merely adds the ability for RPCs to be copied. They remain not copied within any RPC class.

##### Enchancements
* RPCs may now be copied instead passed by reference if desired.

### DIFF
--- a/SmartDeviceLink/SDLAddCommand.m
+++ b/SmartDeviceLink/SDLAddCommand.m
@@ -100,6 +100,13 @@ NS_ASSUME_NONNULL_BEGIN
     return [parameters sdl_objectForName:SDLNameCommandIcon ofClass:SDLImage.class];
 }
 
+-(id)copyWithZone:(nullable NSZone *)zone {
+    SDLAddCommand *newCommand = [super copyWithZone:zone];
+    newCommand->_handler = self.handler;
+
+    return newCommand;
+}
+
 @end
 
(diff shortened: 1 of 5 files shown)


joeljfischer:
@asm09fsu I don't think that's necessary. I don't think protocols usually conform to `NSCopying` for one, and for two, `SDLRequestHandler` is going to go away in 5.0 anyway.

asm09fsu:
@joeljfischer LGTM

PR #541, opened 2017-02-20
Fix lock screen showing when HMI is backgrounded

### LINKED ISSUE #539: No optional lock screen status in background
### Bug Report
There seems to be an issue in the [SDLLockScreenStatusManager.m](https://github.com/smartdevicelink/sdl_ios/blob/master/SmartDeviceLink/SDLLockScreenStatusManager.m) whereby apps don't use the optional status when the app is in background.

##### Reproduction Steps
1. Have a non-media app which using `enabledConfiguration` 
2. Connect&register app to the HMI
3. Make sure DD is off and app knows about DD status
4. Select the app and bring it to foreground
5. Go to radio/nav (get app to background)
6. Reselect the app and bring it back to foreground

##### Expected Behavior
App should never show a lock screen.

##### Observed Behavior
App shows a lock screen at step 5
App removes lock screen again at step 6

That's what should be changed in the status manager class to allow optional for backgruond level, too.

```objc
- (SDLLockScreenStatus *)lockScreenStatus {
    if (self.hmiLevel == nil || [self.hmiLevel isEqualToEnum:[SDLHMILevel NONE]]) {
        // App is not active on the car
        return [SDLLockScreenStatus OFF];
    } else if ([self.hmiLevel isEqualToEnum:[SDLHMILevel BACKGROUND]]) {
        // App is in the background on the car
        // The lockscreen depends entirely on if the user selected the app
        if (self.userSelected) {
-            return [SDLLockScreenStatus REQUIRED];
+            // app is in background but was user selected
+            if (self.haveDriverDistractionStatus && !self.driverDistracted) {
+                // We have the distraction status, and the driver is not distracted
+                return [SDLLockScreenStatus OPTIONAL];
+            } else {
+                // We don't have the distraction status, and/or the driver is distracted
+                return [SDLLockScreenStatus REQUIRED];
+            }
        } else {
            return [SDLLockScreenStatus OFF];
        }
    } else if ([self.hmiLevel isEqualToEnum:[SDLHMILevel FULL]] || [self.hmiLevel isEqualToEnum:[SDLHMILevel LIMITED]]) {
        // App is in the foreground on the car in some manner
        if (self.haveDriverDistractionStatus && !self.driverDistracted) {
            // We have the distraction status, and the driver is not distracted
            return [SDLLockScreenStatus OPTIONAL];
        } else {
            // We don't have the distraction status, and/or the driver is distracted
            return [SDLLockScreenStatus REQUIRED];
        }
    } else {
        // This shouldn't be possible.
        return [SDLLockScreenStatus OFF];
    }
}
```

joeljfischer:
Fixes #539

### Risk

### Testing Plan
Unit tests are included.

### Summary
This fixes a bug in which the lock screen will sometimes appear when it ought not.

##### Bug Fixes
* Update `SDLLockScreenStatusManager` to use OPTIONAL status in the background when appropriate.

### DIFF
--- a/SmartDeviceLink/SDLLockScreenStatusManager.m
+++ b/SmartDeviceLink/SDLLockScreenStatusManager.m
@@ -71,10 +71,17 @@
         return [SDLLockScreenStatus OFF];
     } else if ([self.hmiLevel isEqualToEnum:[SDLHMILevel BACKGROUND]]) {
         // App is in the background on the car
-        // The lockscreen depends entirely on if the user selected the app
         if (self.userSelected) {
-            return [SDLLockScreenStatus REQUIRED];
+            // It was user selected
+            if (self.haveDriverDistractionStatus && !self.driverDistracted) {
+                // We have the distraction status, and the driver is not distracted
+                return [SDLLockScreenStatus OPTIONAL];
+            } else {
+                // We don't have the distraction status, and/or the driver is distracted
+                return [SDLLockScreenStatus REQUIRED];
+            }
         } else {
+            // It wasn't user selected
             return [SDLLockScreenStatus OFF];
         }
     } else if ([self.hmiLevel isEqualToEnum:[SDLHMILevel FULL]] || [self.hmiLevel isEqualToEnum:[SDLHMILevel LIMITED]]) {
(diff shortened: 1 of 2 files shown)


asm09fsu:
@kshala-ford will confirm that this is correct on the functionality side.

kshala-ford:
:+1: tested and works.

PR #544, opened 2017-02-22
Improve description of state machine exception throw

### LINKED ISSUE #543: Invalid state machine transitions are not descriptive enough to debug
### Bug Report
Currently if the app crashes due to an invalid state machine transition, we only received the following message
```
Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Invalid state machine transition occurred'
```

not one that describes what those states are. We cannot see the details of the thrown error (such as the states) in crash logs making it difficult to determine.

Currently we are seeing a crash in `SDLLifecycleManager` relating to invalid state machine transitions but because of this cannot determine the causation.

##### Reproduction Steps
1. Transition from one state to another invalid state
2. Observe crash log

##### Expected Behavior
The developer knows the invalid state machine

##### Observed Behavior
The developer does not know the invalid state.

* SDL iOS Version: 4.5.3

asm09fsu:
Fixes #543 

### Risk

### Summary
This PR adds in a more descriptive reason message for invalid state machine transition.

##### Enchancements
* Improved description when throwing an error for invalid state machine transition.

### DIFF
--- a/SmartDeviceLink/SDLStateMachine.m
+++ b/SmartDeviceLink/SDLStateMachine.m
@@ -62,9 +62,11 @@ SDLStateMachineTransitionFormat const SDLStateMachineTransitionFormatDidEnter = 
     }
 
     if (![self sdl_canState:self.currentState transitionToState:state]) {
+        NSString *classString = NSStringFromClass([self.target class]);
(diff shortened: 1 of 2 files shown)


asm09fsu:
There was not a spec test that actually tested the exception being thrown. This has been added.

PR #556, opened 2017-03-01
Template Updates

### LINKED ISSUE #555: Fix misspellings on PULL_REQUEST_TEMPLATE and ISSUE_TEMPLATE
### Bug Report
There are some minor misspellings within the templates.

asm09fsu:
Fixes #555 

This PR just fixes minor spelling and formatting issues with our templates

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -5,8 +5,8 @@
 
 ##### Reproduction Steps
 1. [Step 1]
-1. [Step 2]
-1. [Step 3]
+2. [Step 2]
+3. [Step 3]
 
 ##### Expected Behavior
 [Some expected behavior]
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -19,8 +19,8 @@ This PR makes **[no / minor / major]** API changes.
 ##### Breaking Changes
 * [Breaking change info]
 
-##### Enchancements
-* [Enchancement info]
+##### Enhancements
+* [Enhancement info]
 
 ##### Bug Fixes
 * [Bug Fix Info]


.github/ISSUE_TEMPLATE.md
```diff
@@ -5,8 +5,8 @@
 
 ##### Reproduction Steps
 1. [Step 1]
-1. [Step 2]
-1. [Step 3]
+2. [Step 2]
```
joeljfischer: Using `1.` in all lists is the correct form for markdown.

asm09fsu: supporting link?

joeljfischer: It renders correctly and means that if you want to insert a new item in the middle of existing items you don't have to rewrite every single number in your list.

```
Ordered lists use numbers followed by periods:

1.  Bird
2.  McHale
3.  Parish

It’s important to note that the actual numbers you use to mark the list have no effect on the HTML output Markdown produces. The HTML Markdown produces from the above list is:

<ol>
<li>Bird</li>
<li>McHale</li>
<li>Parish</li>
</ol>

If you instead wrote the list in Markdown like this:

1.  Bird
1.  McHale
1.  Parish

or even:

3. Bird
1. McHale
8. Parish

you’d get the exact same HTML output. The point is, if you want to, you can use ordinal numbers in your ordered Markdown lists, so that the numbers in your source match the numbers in your published HTML. But if you want to be lazy, you don’t have to.

If you do use lazy list numbering, however, you should still start the list with the number 1. At some point in the future, Markdown may support starting ordered lists at an arbitrary number.
```

From http://daringfireball.net/projects/markdown/syntax#list

asm09fsu: This may be true, although I feel people may not know this. I know I change it every single time I post an issue. Can we set it to 1.2.3. so that people don't question it? Doesn't seem like it negatively impacts anything.

joeljfischer: No, I'd prefer to leave it as is. For those who do use it, such as myself, it's a big timesaver. We'd switch from you changing it to 1.2.3. to me changing it to 1.1.1. Since the 1x3 method saves time and is preferred, I'd rather leave it as is.

asm09fsu: Based on your comment from daring fireball: 
```
If you instead wrote the list in Markdown like this:

1.  Bird
1.  McHale
1.  Parish

or even:

3. Bird
1. McHale
8. Parish

you’d get the exact same HTML output. The point is, if you want to, you can use ordinal numbers in your ordered Markdown lists, so that the numbers in your source match the numbers in your published HTML. But if you want to be lazy, you don’t have to.
```

I'm not sure how this is a "timesaver" as this produces the same either way. I am worried about other developers submitting issues.

PR #562, opened 2017-03-02
Prevent infinite loops in send block while by checking ostream state.

### LINKED ISSUE #522: IAPTransport Writes Output Stream from Different Thread than Scheduled Run Loop

davidswi:
Fixes #522 

This PR is ready for review.

### Risk
This PR makes no API changes.

### Testing Plan
This change *could* be tested by using a mock NSOutputStream injected into the SDLIAPTransport class. However, I tested it with a number of pre-production head units supporting the MFi protocol over BT and USB.

### Summary
This PR addresses very specifically a logic bug in the SDLIAPTransport.m sendData method. Because the NSOutputStream uses polled writes, it places the sendData while loop on a serial dispatch queue to prevent the main thread from blocking. There is a deeper flaw in this design as I've summarized in issue 522, specifically that the output stream is scheduled on the main thread but written to from the thread backing the dispatch queue, violating Apple's guideline for scheduling and accessing streams from the same thread. 

The fact that multiple send blocks can be enqueued on the dispatch queue while the stream is closed on the main thread creates a race condition. As soon as the stream closes, the NSMutableData remainder can no longer be consumed and the length will never reach zero. We have observed a number of SDL app crashes that are actually iOS force quitting the app due to the hung dispatch queue thread. It is easiest to reproduce the crash by backgrounding the app and performing a number of connect/disconnect cycles from the head unit over BT or USB.

The fix for the send block hang is to check the output stream state and exit the block as soon as it is no longer open.

##### Breaking Changes
None

##### Enhancements
None

##### Bug Fixes
Partial fix for issue 522.

### Tasks Remaining:
More PRs to come to transition from polled to scheduled writes.

### DIFF
--- a/SmartDeviceLink/SDLIAPSession.m
+++ b/SmartDeviceLink/SDLIAPSession.m
@@ -7,13 +7,20 @@
 #import "SDLStreamDelegate.h"
 #import "SDLTimer.h"
 
-NS_ASSUME_NONNULL_BEGIN
+#define IO_STREAMTHREAD_NAME            @ "com.smartdevicelink.iostream"
+#define STREAM_THREAD_WAIT_SECS         1.0
+#define STREAM_THREAD_WAIT_WRITE_SECS   5.0
 
 @interface SDLIAPSession ()
 
 @property (assign, nonatomic) BOOL isInputStreamOpen;
 @property (assign, nonatomic) BOOL isOutputStreamOpen;
-
+@property (assign, nonatomic) BOOL isDataSession;
+@property (nonatomic, copy) SessionSendHandler sendBlock;
+@property (nonatomic, strong) NSThread *ioStreamThread;
+@property (nonatomic, assign) BOOL isSendStarted;
+@property (nonatomic, strong) dispatch_semaphore_t canceledSema;
+@property (nonatomic, strong) dispatch_semaphore_t sendCompleteSema;
 @end
 
 
@@ -24,6 +31,13 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithAccessory:(EAAccessory *)accessory forProtocol:(NSString *)protocol {
     NSString *logMessage = [NSString stringWithFormat:@"SDLIAPSession initWithAccessory:%@ forProtocol:%@", accessory, protocol];
     [SDLDebugTool logInfo:logMessage];
+    
+    if ([protocol isEqualToString:@"com.smartdevicelink.prot0"]){
+        _isDataSession = NO;
+    }
+    else{
+        _isDataSession = YES;
+    }
 
     self = [super init];
     if (self) {
@@ -31,6 +45,9 @@ NS_ASSUME_NONNULL_BEGIN
         _protocol = protocol;
         _isInputStreamOpen = NO;
         _isOutputStreamOpen = NO;
+        _canceledSema = dispatch_semaphore_create(0);
+        _sendCompleteSema = dispatch_semaphore_create(0);
+        _sendBlock = nil;
     }
     return self;
 }
@@ -51,9 +68,17 @@ NS_ASSUME_NONNULL_BEGIN
 
         strongSelf.streamDelegate.streamErrorHandler = [self streamErroredHandler];
         strongSelf.streamDelegate.streamOpenHandler = [self streamOpenedHandler];
-
-        [strongSelf startStream:weakSelf.easession.outputStream];
-        [strongSelf startStream:weakSelf.easession.inputStream];
+        if (!self.isDataSession){
+            [strongSelf startStream:weakSelf.easession.outputStream];
+            [strongSelf startStream:weakSelf.easession.inputStream];
+        }
+        else{
+            strongSelf.streamDelegate.streamHasSpaceHandler = [self streamHasSpaceHandler];
+            // Start I/O event loop processing events in iAP channel
+            _ioStreamThread = [[NSThread alloc] initWithTarget:self selector:@selector(accessoryEventLoop) object:nil];
+            [_ioStreamThread setName:IO_STREAMTHREAD_NAME];
+            [_ioStreamThread start];
+        }
 
         return YES;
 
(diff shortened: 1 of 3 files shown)


davidswi:
I would also strongly urge that this change be put into a hotfix for existing apps to update to ASAP.

asm09fsu:
@davidswi as this would be a hot fix, please modify to be merged into master, not develop.

asm09fsu:
@davidswi do you have a way to reproduce the issue this is solving? I would like to see this on my end.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -7,13 +7,20 @@
 #import "SDLStreamDelegate.h"
 #import "SDLTimer.h"
 
-NS_ASSUME_NONNULL_BEGIN
+#define IO_STREAMTHREAD_NAME            @ "com.smartdevicelink.iostream"
```
joeljfischer: Note: We use constants instead of `#defines` in this project.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -7,13 +7,20 @@
 #import "SDLStreamDelegate.h"
 #import "SDLTimer.h"
 
-NS_ASSUME_NONNULL_BEGIN
```
joeljfischer: Why were out nullability checks removed here?

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -24,13 +31,23 @@ @implementation SDLIAPSession
 - (instancetype)initWithAccessory:(EAAccessory *)accessory forProtocol:(NSString *)protocol {
     NSString *logMessage = [NSString stringWithFormat:@"SDLIAPSession initWithAccessory:%@ forProtocol:%@", accessory, protocol];
     [SDLDebugTool logInfo:logMessage];
+    
+    if ([protocol isEqualToString:@"com.smartdevicelink.prot0"]){
+        _isDataSession = NO;
+    }
+    else{
```
joeljfischer: Please note the objective-c style guide [for this project](https://github.com/smartdevicelink/sdl_ios/wiki/Objective-C-Style-Guide). The else should be on the same line as the `}`.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -24,13 +31,23 @@ @implementation SDLIAPSession
 - (instancetype)initWithAccessory:(EAAccessory *)accessory forProtocol:(NSString *)protocol {
     NSString *logMessage = [NSString stringWithFormat:@"SDLIAPSession initWithAccessory:%@ forProtocol:%@", accessory, protocol];
     [SDLDebugTool logInfo:logMessage];
+    
+    if ([protocol isEqualToString:@"com.smartdevicelink.prot0"]){
```
joeljfischer: This constant is defined in `SDLIAPTransport.m`, it should probably be shared, but adding it to the `SDLIAPTransport.h` header will force this to be a minor version change.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -24,13 +31,23 @@ @implementation SDLIAPSession
...
+        _isDataSession = NO;
```
joeljfischer: This should not be before the `self = [super init]`

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -7,13 +7,20 @@
...
 
 @interface SDLIAPSession ()
 
 @property (assign, nonatomic) BOOL isInputStreamOpen;
 @property (assign, nonatomic) BOOL isOutputStreamOpen;
-
+@property (assign, nonatomic) BOOL isDataSession;
+@property (nonatomic, copy) SessionSendHandler sendBlock;
+@property (nonatomic, strong) NSThread *ioStreamThread;
+@property (nonatomic, assign) BOOL isSendStarted;
```
joeljfischer: I'm not a huge fan of having so many state variables, would a change to a state machine make more sense here? Should we have multiple subclasses or adherents to a protocol to handle the differing data session / control session cases here?

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -139,6 +273,51 @@ - (SDLStreamErrorHandler)streamErroredHandler {
...
+            SessionSendHandler sendCompletion;
+            @synchronized (strongSelf) {
+                sendCompletion = strongSelf.sendBlock;
+            }
+            
+            if (sendCompletion != nil){
+                NSError *err = nil;
+                BOOL sendComplete = sendCompletion(&err);
+                if (err != nil){
+                    NSLog(@"ERROR!!! Output stream reported %@", err);
```
joeljfischer: This would use `SDLDebugTool` along with all the other `NSLog`'s around here.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -51,9 +68,17 @@ - (BOOL)start {
...
-        [strongSelf startStream:weakSelf.easession.outputStream];
-        [strongSelf startStream:weakSelf.easession.inputStream];
+        if (!self.isDataSession){
+            [strongSelf startStream:weakSelf.easession.outputStream];
+            [strongSelf startStream:weakSelf.easession.inputStream];
+        }
+        else{
+            strongSelf.streamDelegate.streamHasSpaceHandler = [self streamHasSpaceHandler];
+            // Start I/O event loop processing events in iAP channel
+            _ioStreamThread = [[NSThread alloc] initWithTarget:self selector:@selector(accessoryEventLoop) object:nil];
```
joeljfischer: If it's not in `init` or `dealloc`, properties should be referenced via `self`, unless there's a really good reason.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -64,9 +89,118 @@ - (BOOL)start {
...
-    [self stopStream:self.easession.inputStream];
+
+    if (!self.isDataSession){
+        [self stopStream:self.easession.outputStream];
+        [self stopStream:self.easession.inputStream];
+    }
+    else{
+        [self.ioStreamThread cancel];
+        
+        long lWait = dispatch_semaphore_wait(self.canceledSema, dispatch_time(DISPATCH_TIME_NOW, STREAM_THREAD_WAIT_SECS * NSEC_PER_SEC));
```
joeljfischer: We're going to need some comments on this kind of new code to understand the thinking here.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -64,9 +89,118 @@ - (BOOL)start {
...
+    if (lWait != 0){
+        NSLog(@"ERROR! Failed to write data!!!");
+    }
+}
+
+- (void)accessoryEventLoop {
+    BOOL sendStarted;
+    SessionSendHandler sendBlock;
+    
+  @autoreleasepool {
```
joeljfischer: What's the purpose of this pool?

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -64,9 +89,118 @@ - (BOOL)start {
...
+    }
+    long lWait = dispatch_semaphore_wait(self.sendCompleteSema, dispatch_time(DISPATCH_TIME_NOW, STREAM_THREAD_WAIT_WRITE_SECS * NSEC_PER_SEC));
+    if (lWait != 0){
+        NSLog(@"ERROR! Failed to write data!!!");
+    }
+}
+
+- (void)accessoryEventLoop {
+    BOOL sendStarted;
+    SessionSendHandler sendBlock;
```
joeljfischer: These should be initialized with something, even if it's the default `NO` and `nil`.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -64,9 +89,118 @@ - (BOOL)start {
...
+        self.sendBlock = sender;
+        self.isSendStarted = NO;
+    }
+    long lWait = dispatch_semaphore_wait(self.sendCompleteSema, dispatch_time(DISPATCH_TIME_NOW, STREAM_THREAD_WAIT_WRITE_SECS * NSEC_PER_SEC));
+    if (lWait != 0){
+        NSLog(@"ERROR! Failed to write data!!!");
+    }
+}
+
+- (void)accessoryEventLoop {
```
joeljfischer: Private methods should prepend `sdl_` before the rest of the method name.

SmartDeviceLink/SDLIAPSession.m
davidswi: Hi Joel, it looks like I had too little sleep yesterday and a bit too much caffeine. I pushed some in-progress changes to my develop branch fork and these commits wound up intermingled with the single focused change that I wanted to get reviewed. I think this feedback is great and you should know that I've corrected some of the issues, but the final design will use a FIFO for managing the in-flight send fragments rather than abusing a dispatch queue.

davidswi: I think I'd like to close this PR and re-submit the original focused change against master.

joeljfischer: Sounds good!

davidswi: Also Alex asked about reproducing the specific issue - iOS terminating the app due to high CPU load in the background because of the race condition between the main thread and the send completion block on the transmit queue. The best way to reproduce is as follows:

1. Launch a background enabled SDL app such as Spotify and connect to the head unit
2. Background the app -- turn off the display as this enables iOS aggressive power management.
3. Disconnect the app from the head unit while a number of buffers are in flight to the head unit, say during a large file upload.
4. Reconnect the app to the head unit. It may be necessary to do several cycles of 3 and 4.

Result: The app will stop playing media and when it is brought to the foreground, it will disappear. The device logs will show that the app has been terminated due to excessive CPU usage in the thread backing the iAP transmit queue.

Based on how frequently you are able to reproduce the issue, I am open to making the PR a hotfix or a standard bug fix. I apologize for being a bit rushed yesterday. Sleep deprivation has a way of doing that.

davidswi: I have a somewhat related question about the ~9s delay after the accessory connect event before the control session establishment is attempted. In my testing of the background thread changes, the various crashes I've seen due to the cross-thread access to the output stream are no longer reproducible. 

However, the 9s delay causes the app to resume much more slowly than one would ideally like. There is also an issue with running timers in the background that causes the app to not resume if the control session is not established right away. If the timer is needed and if it is necessary to retry more than once, the code should be placed in a background task when the app is backgrounded. I have another PR in the hopper for the background task, but before I submit that, could we eliminate the timer?

joeljfischer: > Based on how frequently you are able to reproduce the issue, I am open to making the PR a hotfix or a standard bug fix.

So that you know "hotfix" and "bug fix" are identical, we only differentiate change types by [semantic versioning](https://semver.org).

The delay is necessary due to attempts to resume apps failing in certain cases. I haven't had to think about it in a while, so I'm not exactly sure. @asm09fsu do you remember?

davidswi: Thanks for the clarification Joel. I could see cases where the head unit is busy, say indexing the iTunes media library on the iOS device or updating the SXM guide. Could we accomplish the same result - reliable reconnection - by making the delay shorter and allowing retries, but within a background task if the app is backgrounded?

asm09fsu:
@davidswi please make sure you branch from master instead of develop, as develop is not geared towards an immediate release.

PR #564, opened 2017-03-03
Fix issue around premature stream close resulting in hangs in iAP transmit queue

### LINKED ISSUE #522: IAPTransport Writes Output Stream from Different Thread than Scheduled Run Loop

davidswi:
Fixes #522 -- hang due to stream close on main thread while loop runs on IAP transmit queue

This PR is ready for review.

### Risk
This PR makes no API changes. However, it affects *every* send over the IAP transport, so risk is moderate.

### Testing Plan
Testing has been performed using test apps against several pre-production head units running SDLCore.

### Summary
[Summary of PR changes]

##### Breaking Changes
* No breaking changes.

##### Enhancements
* Not an enhancement.

##### Bug Fixes
* Fixes hangs in the IAP transmit queue when the output stream closes on the main thread before all data in the current send block has been transmitted.

### Tasks Remaining:
Major changes to move input and output stream to background thread and perform all I/O using scheduled callbacks. Separate PR.

### DIFF
--- a/SmartDeviceLink/SDLIAPTransport.m
+++ b/SmartDeviceLink/SDLIAPTransport.m
@@ -272,8 +272,9 @@ int const streamOpenTimeoutSeconds = 2;
         NSOutputStream *ostream = self.session.easession.outputStream;
         NSMutableData *remainder = data.mutableCopy;
 
-        while (remainder.length != 0) {
-            if (ostream.streamStatus == NSStreamStatusOpen && ostream.hasSpaceAvailable) {
+        while (ostream.streamStatus == NSStreamStatusOpen &&
+               remainder.length != 0) {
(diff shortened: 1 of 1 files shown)


davidswi:
I'd like to get this reviewed against develop first. If folks agree with the change and want it merged into master as a hotfix, I'll resubmit against master without the develop changes.

asm09fsu:
@davidswi It may be easier to discuss these PRs on [slack](slack.smartdevicelink.com).

davidswi:
OK thanks, I will close and rebase the commit against master.

davidswi:
How do I get invited to your Slack?

asm09fsu:
@davidswi the slack is linked above. Or [slack.smartdevicelink.com](slack.smartdevicelink.com).

PR #565, opened 2017-03-03
Fix issue around premature stream close resulting in hangs in iAP transmit queue

### LINKED ISSUE #522: IAPTransport Writes Output Stream from Different Thread than Scheduled Run Loop

davidswi:
Fixes #522

### Risk

This PR makes **no** API changes. However, it affects every send over the IAP transport, so risk is moderate.

### Testing Plan

Testing has been performed using test apps against several pre-production head units running SDLCore.

### Summary
Fixes a hang due to stream close on main thread while loop runs on IAP transmit queue

##### Bug Fixes
* Fixes hangs in the IAP transmit queue when the output stream closes on the main thread before all data in the current send block has been transmitted.

### DIFF
--- a/SmartDeviceLink/SDLIAPTransport.m
+++ b/SmartDeviceLink/SDLIAPTransport.m
@@ -272,8 +272,9 @@ int const streamOpenTimeoutSeconds = 2;
         NSOutputStream *ostream = self.session.easession.outputStream;
         NSMutableData *remainder = data.mutableCopy;
 
-        while (remainder.length != 0) {
-            if (ostream.streamStatus == NSStreamStatusOpen && ostream.hasSpaceAvailable) {
+        while (ostream.streamStatus == NSStreamStatusOpen &&
+               remainder.length != 0) {
+            if (ostream.hasSpaceAvailable){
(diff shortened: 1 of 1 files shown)


davidswi:
I should point out that this change will not be necessary if/once the background thread stream processing changes are accepted into the SDL iOS proxy.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -272,8 +272,9 @@ - (void)sendData:(NSData *)data {
         NSOutputStream *ostream = self.session.easession.outputStream;
         NSMutableData *remainder = data.mutableCopy;
 
-        while (remainder.length != 0) {
-            if (ostream.streamStatus == NSStreamStatusOpen && ostream.hasSpaceAvailable) {
+        while (ostream.streamStatus == NSStreamStatusOpen &&
+               remainder.length != 0) {
```
joeljfischer: Style-wise, unless there's many checks, we generally prefer to leave two checks on the same line and let wrapping occur naturally if necessary.

PR #566, opened 2017-03-06
Updating Data Session I/O to be in compliance with Apple guidelines

### LINKED ISSUE #522: IAPTransport Writes Output Stream from Different Thread than Scheduled Run Loop

davidswi:
Fixes #522

This PR is ready for review.

### Risk
This PR makes no API changes.

### Testing Plan
We will perform manual and automated testing with a number of internal SDL test apps and prototype headunit hardware, covering validation and verification of correct RPC and a variety of cold and warm connect/disconnect scenarios over BT and USB.

### Summary

##### Breaking Changes
* No API breaking changes.

##### Enhancements

##### Bug Fixes

### Tasks Remaining:
Requesting SDL iOS team feedback so any implementation issues are addressed before a full internal test pass.

### DIFF
--- a/SmartDeviceLink/SDLIAPSession.h
+++ b/SmartDeviceLink/SDLIAPSession.h
@@ -10,8 +10,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-typedef void (^SessionCompletionHandler)(BOOL success);
-
 @interface SDLIAPSession : NSObject
 
 @property (nullable, strong, nonatomic) EAAccessory *accessory;
--- a/SmartDeviceLink/SDLIAPSession.h
+++ b/SmartDeviceLink/SDLIAPSession.h
@@ -10,7 +10,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-typedef void (^SessionCompletionHandler)(BOOL success);
+@class SDLStreamDelegate;
 
 @interface SDLIAPSession : NSObject
 
@@ -25,6 +25,7 @@ typedef void (^SessionCompletionHandler)(BOOL success);
 
 - (BOOL)start;
 - (void)stop;
+- (void)sendData:(NSData *)data;
 
 @end
 
--- a/SmartDeviceLink/SDLIAPSession.h
+++ b/SmartDeviceLink/SDLIAPSession.h
@@ -10,7 +10,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class SDLStreamDelegate;
+typedef void (^SessionCompletionHandler)(BOOL success);
 
 @interface SDLIAPSession : NSObject
 
--- a/SmartDeviceLink/SDLIAPSession.m
+++ b/SmartDeviceLink/SDLIAPSession.m
@@ -14,8 +14,6 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
 
 @interface SDLIAPSession ()
 
-@property (assign, nonatomic) BOOL isInputStreamOpen;
-@property (assign, nonatomic) BOOL isOutputStreamOpen;
 @property (assign, nonatomic) BOOL isDataSession;
 @property (nonatomic, strong, nullable) NSThread *ioStreamThread;
 @property (nonatomic, strong, nullable) SDLMutableDataQueue *sendDataQueue;
@@ -37,8 +35,6 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
         _isDataSession = [protocol isEqualToString:@"com.smartdevicelink.prot0"] ? NO : YES;
         _accessory = accessory;
         _protocol = protocol;
-        _isInputStreamOpen = NO;
-        _isOutputStreamOpen = NO;
         _canceledSemaphore = dispatch_semaphore_create(0);
         _sendDataQueue = [[SDLMutableDataQueue alloc] init];
     }
@@ -233,14 +229,13 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
 
         if (stream == [strongSelf.easession outputStream]) {
             [SDLDebugTool logInfo:@"Output Stream Opened"];
-            strongSelf.isOutputStreamOpen = YES;
         } else if (stream == [strongSelf.easession inputStream]) {
             [SDLDebugTool logInfo:@"Input Stream Opened"];
-            strongSelf.isInputStreamOpen = YES;
         }
 
         // When both streams are open, session initialization is complete. Let the delegate know.
-        if (strongSelf.isInputStreamOpen && strongSelf.isOutputStreamOpen) {
+        if (strongSelf.easession.inputStream.streamStatus == NSStreamStatusOpen &&
+            strongSelf.easession.outputStream.streamStatus == NSStreamStatusOpen) {
             [strongSelf.delegate onSessionInitializationCompleteForSession:weakSelf];
         }
     };
@@ -276,7 +271,6 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
 #pragma mark - Lifecycle Destruction
 
 - (void)dealloc {
-    [self.sendDataQueue flush];
     self.sendDataQueue = nil;
     self.delegate = nil;
     self.accessory = nil;
--- a/SmartDeviceLink/SDLIAPSession.m
+++ b/SmartDeviceLink/SDLIAPSession.m
@@ -19,7 +19,7 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
 @property (assign, nonatomic) BOOL isDataSession;
 @property (nonatomic, strong, nullable) NSThread *ioStreamThread;
 @property (nonatomic, strong, nullable) SDLMutableDataQueue *sendDataQueue;
-@property (nonatomic, strong, nullable) dispatch_semaphore_t canceledSema;
+@property (nonatomic, strong, nullable) dispatch_semaphore_t canceledSemaphore;
 
 @end
 
@@ -34,17 +34,12 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
 
     self = [super init];
     if (self) {
-        if ([protocol isEqualToString:@"com.smartdevicelink.prot0"]){
-            _isDataSession = NO;
-        }
-        else{
-            _isDataSession = YES;
-        }
+        _isDataSession = [protocol isEqualToString:@"com.smartdevicelink.prot0"] ? NO : YES;
         _accessory = accessory;
         _protocol = protocol;
         _isInputStreamOpen = NO;
         _isOutputStreamOpen = NO;
-        _canceledSema = dispatch_semaphore_create(0);
+        _canceledSemaphore = dispatch_semaphore_create(0);
         _sendDataQueue = [[SDLMutableDataQueue alloc] init];
     }
     return self;
@@ -54,35 +49,32 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
 #pragma mark - Public Stream Lifecycle
 
 - (BOOL)start {
-    __weak typeof(self) weakSelf = self;
-
     NSString *logMessage = [NSString stringWithFormat:@"Opening EASession withAccessory:%@ forProtocol:%@", _accessory.name, _protocol];
     [SDLDebugTool logInfo:logMessage];
-
-    if ((self.easession = [[EASession alloc] initWithAccessory:_accessory forProtocol:_protocol])) {
-        __strong typeof(self) strongSelf = weakSelf;
-
-        [SDLDebugTool logInfo:@"Created Session Object"];
-
-        strongSelf.streamDelegate.streamErrorHandler = [self streamErroredHandler];
-        strongSelf.streamDelegate.streamOpenHandler = [self streamOpenedHandler];
-        if (!self.isDataSession) {
-            [strongSelf startStream:weakSelf.easession.outputStream];
-            [strongSelf startStream:weakSelf.easession.inputStream];
-        } else {
-            strongSelf.streamDelegate.streamHasSpaceHandler = [self streamHasSpaceHandler];
-            // Start I/O event loop processing events in iAP channel
-            self.ioStreamThread = [[NSThread alloc] initWithTarget:self selector:@selector(sdl_accessoryEventLoop) object:nil];
-            [self.ioStreamThread setName:ioStreamThreadName];
-            [self.ioStreamThread start];
-        }
-
-        return YES;
-
-    } else {
+    
+    self.easession = [[EASession alloc] initWithAccessory:self.accessory forProtocol:self.protocol];
+    
+    if (!self.easession) {
         [SDLDebugTool logInfo:@"Error: Could Not Create Session Object"];
         return NO;
     }
+    
+    [SDLDebugTool logInfo:@"Created Session Object"];
+    
+    self.streamDelegate.streamErrorHandler = [self streamErroredHandler];
+    self.streamDelegate.streamOpenHandler = [self streamOpenedHandler];
+    if (!self.isDataSession) {
+        [self startStream:self.easession.outputStream];
+        [self startStream:self.easession.inputStream];
+    } else {
+        self.streamDelegate.streamHasSpaceHandler = [self streamHasSpaceHandler];
+        // Start I/O event loop processing events in iAP channel
+        self.ioStreamThread = [[NSThread alloc] initWithTarget:self selector:@selector(sdl_accessoryEventLoop) object:nil];
+        [self.ioStreamThread setName:ioStreamThreadName];
+        [self.ioStreamThread start];
+    }
+    
+    return YES;
 }
 
 - (void)stop {
@@ -93,7 +85,7 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
     } else {
         [self.ioStreamThread cancel];
         
-        long lWait = dispatch_semaphore_wait(self.canceledSema, dispatch_time(DISPATCH_TIME_NOW, streamThreadWaitSecs * NSEC_PER_SEC));
+        long lWait = dispatch_semaphore_wait(self.canceledSemaphore, dispatch_time(DISPATCH_TIME_NOW, streamThreadWaitSecs * NSEC_PER_SEC));
         if (lWait == 0) {
             [SDLDebugTool logInfo:@"Stream thread canceled"];
         } else{
@@ -107,12 +99,12 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
 
 - (void)sendData:(NSData *)data{
     // Enqueue the data for transmission on the IO thread
-    [self.sendDataQueue enqueue:data.mutableCopy];
+    [self.sendDataQueue enqueueBuffer:data.mutableCopy];
 }
 
 - (BOOL)sdl_dequeueAndWriteToOutputStream:(NSError **)error{
     NSOutputStream *ostream = self.easession.outputStream;
-    NSMutableData *remainder = [self.sendDataQueue front];
+    NSMutableData *remainder = [self.sendDataQueue frontBuffer];
     BOOL allDataWritten = NO;
     
     if (error != nil && remainder != nil && ostream.streamStatus == NSStreamStatusOpen) {
@@ -120,18 +112,16 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
         NSInteger bytesWritten = [ostream write:remainder.bytes maxLength:bytesRemaining];
         if (bytesWritten < 0) {
             *error = ostream.streamError;
-        } else {
-            if (bytesWritten == bytesRemaining) {
+        } else if (bytesWritten == bytesRemaining) {
                 // Remove the data from the queue
-                [self.sendDataQueue pop];
+                [self.sendDataQueue popBuffer];
                 allDataWritten = YES;
-            } else{
+            } else {
                 // Cleave the sent bytes from the data, the remainder will sit at the head of the queue
                 [remainder replaceBytesInRange:NSMakeRange(0, bytesWritten) withBytes:NULL length:0];
             }
-        }
     }
-    
+
     return allDataWritten;
 }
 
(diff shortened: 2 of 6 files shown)


davidswi:
A quick summary of the advantages of this implementation versus the current upstream version:
1. It eliminates the cross-thread communication that is explicitly not allowed by Apple for scheduled streams.
2. It moves from polled sends to event driven, scheduled sends allowing the CPU to be powered managed effectively on battery power. Apple cares a lot about apps using the CPU correctly and not interfering with power management.
3. By moving I/O from the main thread --- which is likely to be busy processing notifications and main queue dispatch blocks in commercial apps -- to a background thread dedicated to the IAP communication, the outbound data transfer rate should be improved. We will collect some performance metrics for some key scenarios:
(1) SDL video data transfer
(2) Large PutFile operations

SmartDeviceLink/SDLMutableDataQueue.h
```diff
@@ -0,0 +1,35 @@
+//
+//  SDLMutableDataQueue.h
+//  
+//
+//  Created by David Switzer on 3/2/17.
+//
+//
+
```
asm09fsu: Can we wrap this in `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END`?

SmartDeviceLink/SDLMutableDataQueue.h
```diff
@@ -0,0 +1,35 @@
...
+// front returns the NSMutableData * buffer at the front of the queue,
+// but does not remove it -- modeled after the STL C++ queue front method
+- (NSMutableData *)front;
+
+// pop removes the buffer at the head of the queue -- modeled after the
+// STL C++ queue pop method
+- (void)pop;
+
+// Enqueues a new NSMutableData * buffer at the back of the queue
+- (void)enqueue:(NSMutableData *)data;
```
asm09fsu: Can we do `- (void)enqueueBuffer:(NSMutableData *)buffer;` instead?

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -0,0 +1,72 @@
+//
+//  SDLMutableDataQueue.m
+//  
+//
+//  Created by David Switzer on 3/2/17.
+//
+//
+
+#import "SDLMutableDataQueue.h"
+
```
asm09fsu: Can we wrap this in `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END`?

SmartDeviceLink/SDLMutableDataQueue.h
```diff
@@ -0,0 +1,35 @@
...
+
+- (instancetype)init;
+
+// front returns the NSMutableData * buffer at the front of the queue,
+// but does not remove it -- modeled after the STL C++ queue front method
+- (NSMutableData *)front;
+
+// pop removes the buffer at the head of the queue -- modeled after the
+// STL C++ queue pop method
+- (void)pop;
```
asm09fsu: Can we change this to `- (void)popBuffer;`?

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -0,0 +1,72 @@
...
+@implementation SDLMutableDataQueue
+
+- (instancetype)init{
+    self = [super init];
+    if (self != nil) {
+        self.elements = [[NSMutableArray alloc] init];
+    }
+    
+    return self;
```
asm09fsu: This should be 
```
- (instancetype)init {
    self = [super init];
    if (!self) {
        return nil;
    }

    self.elements = [[NSMutableArray alloc] init];

    return self;
```

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+- (NSMutableData *)front{
+    NSMutableData *dataAtFront = nil;
+    
+    @synchronized (self) {
+        if (self.elements.count > 0) {
+            // The front of the queue is always at index 0
+            dataAtFront = [self.elements objectAtIndex:0];
```
asm09fsu: This can be `self.elements[0];`

SmartDeviceLink/SDLMutableDataQueue.h
```diff
@@ -0,0 +1,35 @@
...
+//
+//  Created by David Switzer on 3/2/17.
+//
+//
+
+#import <Foundation/Foundation.h>
+
+@interface SDLMutableDataQueue : NSObject
+
+- (instancetype)init;
```
asm09fsu: You can delete this.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -64,11 +86,126 @@ - (BOOL)start {
...
+        self.easession.outputStream == nil ||
+        self.easession.outputStream.streamStatus != NSStreamStatusOpen) {
+        [self.sendDataQueue flush];
+    }
+}
+
+- (void)sdl_accessoryEventLoop {
+    
+  @autoreleasepool {
+    NSAssert(self.easession, @"_session must be assigned before calling");
```
asm09fsu: looks like the spacing is a bit off here.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -64,11 +86,126 @@ - (BOOL)start {
...
+    // Open I/O streams of the iAP session
+    NSInputStream *inStream = [self.easession inputStream];
+    [inStream setDelegate:self.streamDelegate];
+    [inStream scheduleInRunLoop:[NSRunLoop currentRunLoop] forMode:NSDefaultRunLoopMode];
+    [inStream open];
+    
+    NSOutputStream *outStream = [self.easession outputStream];
+    [outStream setDelegate:self.streamDelegate];
+    [outStream scheduleInRunLoop:[NSRunLoop currentRunLoop] forMode:NSDefaultRunLoopMode];
+    [outStream open];
```
asm09fsu: These are being scheduled in the current run loop, yet in the `start` method, we use the `startStream:` method which opens them on the main run loop. Is this correct? Should the `startStream:` method be opening on the current run loop instead? If so, we can modify that function and remove this code, and use that function instead to shorten this a bit.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -64,11 +86,126 @@ - (BOOL)start {
...
+    NSMutableData *remainder = [self.sendDataQueue front];
+    BOOL allDataWritten = NO;
+    
+    if (error != nil && remainder != nil && ostream.streamStatus == NSStreamStatusOpen) {
+        NSInteger bytesRemaining = remainder.length;
+        NSInteger bytesWritten = [ostream write:remainder.bytes maxLength:bytesRemaining];
+        if (bytesWritten < 0) {
+            *error = ostream.streamError;
+        } else {
+            if (bytesWritten == bytesRemaining) {
```
asm09fsu: can we combine the else and if?

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -64,11 +86,126 @@ - (BOOL)start {
...
+    do {
+        if (self.sendDataQueue.count > 0 && !self.sendDataQueue.frontDequeued) {
+            NSError *sendErr = nil;
+            if (![self sdl_dequeueAndWriteToOutputStream:&sendErr] && sendErr != nil) {
+                [self sdl_handleOutputStreamWriteError:sendErr];
+            }
+        }
+        
+        [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode
+                                 beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.25f]];
```
asm09fsu: What is the reasoning for the 0.25f delay?

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -6,13 +6,20 @@
...
+NSTimeInterval const streamThreadWaitSecs = 1.0;
 
 @interface SDLIAPSession ()
 
 @property (assign, nonatomic) BOOL isInputStreamOpen;
 @property (assign, nonatomic) BOOL isOutputStreamOpen;
+@property (assign, nonatomic) BOOL isDataSession;
+@property (nonatomic, strong, nullable) NSThread *ioStreamThread;
+@property (nonatomic, strong, nullable) SDLMutableDataQueue *sendDataQueue;
+@property (nonatomic, strong, nullable) dispatch_semaphore_t canceledSema;
```
asm09fsu: can we write this property out? `canceledSemaphore`

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -27,10 +34,18 @@ - (instancetype)initWithAccessory:(EAAccessory *)accessory forProtocol:(NSString
 
     self = [super init];
     if (self) {
+        if ([protocol isEqualToString:@"com.smartdevicelink.prot0"]){
+            _isDataSession = NO;
+        }
+        else{
+            _isDataSession = YES;
+        }
```
asm09fsu: We could shorten this a bit `_isDataSession = [protocol isEqualToString:@"com.smartdevicelink.prot0"] ? NO : YES;`

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -51,9 +66,16 @@ - (BOOL)start {
...
+            [strongSelf startStream:weakSelf.easession.inputStream];
+        } else {
+            strongSelf.streamDelegate.streamHasSpaceHandler = [self streamHasSpaceHandler];
+            // Start I/O event loop processing events in iAP channel
+            self.ioStreamThread = [[NSThread alloc] initWithTarget:self selector:@selector(sdl_accessoryEventLoop) object:nil];
+            [self.ioStreamThread setName:ioStreamThreadName];
+            [self.ioStreamThread start];
+        }
 
         return YES;
```
asm09fsu: I don't know why this function is written this way. I know it used to be this way in the first place. Can we instead do the following?

```
- (BOOL)start {
    NSString *logMessage = [NSString stringWithFormat:@"Opening EASession withAccessory:%@ forProtocol:%@", _accessory.name, _protocol];

    self.easession = [[EASession alloc] initWithAccessory:self.accessory forProtocol:self.protocol];
    
    if (!self.easession) {
        [SDLDebugTool logInfo:@"Error: Could Not Create Session Object"];
        return NO;
    }
    
    [SDLDebugTool logInfo:@"Created Session Object"];
    
    self.streamDelegate.streamErrorHandler = [self streamErroredHandler];
    self.streamDelegate.streamOpenHandler = [self streamOpenedHandler];
    if (!self.isDataSession) {
        [self startStream:self.easession.outputStream];
        [self startStream:self.easession.inputStream];
    } else {
        self.streamDelegate.streamHasSpaceHandler = [self streamHasSpaceHandler];
        // Start I/O event loop processing events in iAP channel
        self.ioStreamThread = [[NSThread alloc] initWithTarget:self selector:@selector(sdl_accessoryEventLoop) object:nil];
        [self.ioStreamThread setName:ioStreamThreadName];
        [self.ioStreamThread start];
    }
    
    return YES;
}
```

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -139,6 +276,37 @@ - (SDLStreamErrorHandler)streamErroredHandler {
...
+                [strongSelf sdl_handleOutputStreamWriteError:sendErr];
+            }
+        }
+    };
+}
+
+#pragma mark - Lifecycle Destruction
+
+- (void)dealloc {
+    [self.sendDataQueue flush];
```
asm09fsu: Do we need to call flush?

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+- (NSUInteger)count{
+    @synchronized (self) {
+        return self.elements.count;
+    }
+}
+
+- (void)dealloc{
```
asm09fsu: Dealloc should be handled automatically.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -139,6 +276,37 @@ - (SDLStreamErrorHandler)streamErroredHandler {
...
+    self.sendDataQueue = nil;
+    self.delegate = nil;
+    self.accessory = nil;
+    self.protocol = nil;
+    self.streamDelegate = nil;
+    self.easession = nil;
+    self.ioStreamThread =  nil;
+    self.canceledSema = nil;
```
asm09fsu: don't we need to call `dispatch_release`?

asm09fsu:
Because we are now adhering to Apple's guidelines on using streams, this PR does in fact fix the issue that we put in a hot fix for around restarting the proxy: https://github.com/smartdevicelink/sdl_ios/blob/master/SmartDeviceLink/SDLLifecycleManager.m#L193

With this PR, I believe we are able to remove this `dispatch_after` block. A way to check is the following:
With current implementation (not this PR), comment out the `dispatch_after` wrapper in `SDLLifecycleManager`'s `sdl_stopManager:` function.
Then go into the example's `ProxyManager` class and add `[self startManager];` to the end of the `reset` function.

Then run the following steps in the example app:
1. Open the sdl_ios project, and open the SmartDeviceLink-iOS.xcodeproj in Xcode.
2. Deploy SmartDeviceLink-Example to an iPhone.
3. Connect to a SYNC 3 headunit.
4. On the iPhone's UI, select the "iAP" segmented control.
5. Press "Connect".
6. See app on Head Unit, under "Apps". Will be called "SDL Example App".
7. Disconnect iPhone from head unit.
8. Reconnect iPhone to head unit.
9. Wait for app to reappear on the head unit.
10. Press the same button again (should be "Disconnect").
11. Crash.

If we then use this branch instead, along with that change to the example app, and the commented out `dispatch_after` block, we see we no longer will immediately crash.

SmartDeviceLink/SDLIAPSession.m
davidswi: I wanted to respond to this comment first as it is the highest priority issue - getting the scheduling wrong! We only call startStream if we are starting the control session and it does schedule on the main run loop. However, if we are starting the data session, we do the scheduling in the beginning of the background thread function. So I believe as implemented the scheduling and access are all done on the correct thread. Nonetheless, I agree it would be cleaner to use startStream in all cases.

goelgoldi:
@theresalech This is high priority ticket for Xevo.

SmartDeviceLink/SDLIAPSession.h
```diff
@@ -10,7 +10,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class SDLStreamDelegate;
+typedef void (^SessionCompletionHandler)(BOOL success);
```
davidswi: OK

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -9,17 +9,17 @@
...
 
 NS_ASSUME_NONNULL_BEGIN
-#define IO_STREAMTHREAD_NAME            @ "com.smartdevicelink.iostream"
-#define STREAM_THREAD_WAIT_SECS         1.0
+NSString *const ioStreamThreadName  = @ "com.smartdevicelink.iostream";
+NSTimeInterval const streamThreadWaitSecs = 1.0;
 
 @interface SDLIAPSession ()
 
 @property (assign, nonatomic) BOOL isInputStreamOpen;
```
davidswi: We don't strictly need it, but we do need to check that both streams are open before calling the delegate opened method. I didn't feel a lot of pressure to change the usage of those boolean member variables, but I suppose we could check the stream states directly.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -9,17 +9,17 @@
...
 @property (assign, nonatomic) BOOL isOutputStreamOpen;
```
davidswi: Yes, see above.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -9,17 +9,17 @@
...
 @property (assign, nonatomic) BOOL isDataSession;
-@property (nonatomic, strong) NSThread *ioStreamThread;
-@property (nonatomic, strong) SDLMutableDataQueue *sendDataQueue;
-@property (nonatomic, strong) dispatch_semaphore_t canceledSema;
+@property (nonatomic, strong, nullable) NSThread *ioStreamThread;
+@property (nonatomic, strong, nullable) SDLMutableDataQueue *sendDataQueue;
+@property (nonatomic, strong, nullable) dispatch_semaphore_t canceledSema;
```
davidswi: Done.

SmartDeviceLink/SDLIAPSession.m
davidswi: Done.

SmartDeviceLink/SDLIAPSession.m
davidswi: I like that version better. Replaced mine with that.

SmartDeviceLink/SDLIAPSession.m
davidswi: Fixed.

SmartDeviceLink/SDLIAPSession.m
davidswi: We are now using startStream in all cases with [NSRunLoop currentRunLoop].

SmartDeviceLink/SDLIAPSession.m
davidswi: The reasoning is only to block in runMode for a maximum of 0.25s waiting for a stream event. It is not a delay, it is a timeout similar to a try wait or a dispatch wait with a timeout.

SmartDeviceLink/SDLIAPSession.m
davidswi: Yes, that will ensure the data queue is able to be deallocated, avoiding any memory leaks.

SmartDeviceLink/SDLIAPSession.m
davidswi: From Stack Overflow: http://stackoverflow.com/questions/19454189/do-i-ever-have-to-call-dispatch-release-under-arc
All GCD objects are managed by ARC, so you don't need to explicitly manage their memory. What is true for dispatch_queue_t also applies to all the other GCD objects.

SmartDeviceLink/SDLMutableDataQueue.m
davidswi: Done.

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -21,7 +19,7 @@ @implementation SDLMutableDataQueue
 
 - (instancetype)init{
     self = [super init];
-    if (self != nil){
+    if (self != nil) {
```
davidswi: Fixed.

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -39,7 +37,7 @@ - (NSMutableData *)front{
     NSMutableData *dataAtFront = nil;
     
     @synchronized (self) {
-        if (self.elements.count > 0){
+        if (self.elements.count > 0) {
```
davidswi: Done.

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -39,7 +37,7 @@ - (NSMutableData *)front{
...
             // The front of the queue is always at index 0
             dataAtFront = [self.elements objectAtIndex:0];
```
davidswi: Done.

SmartDeviceLink/SDLMutableDataQueue.h
davidswi: None.

SmartDeviceLink/SDLMutableDataQueue.h
davidswi: Is it a convention in the SDL iOS project to not specify the prototype for init in the public interface?

SmartDeviceLink/SDLMutableDataQueue.h
```diff
@@ -0,0 +1,35 @@
...
+- (void)enqueue:(NSMutableData *)data;
+
+// Empty the queue
+- (void)flush;
+
+@property(nonatomic, assign, readonly) NSUInteger count;
+
+// Since data can be dequeued from the front multiple times, use this
+// flag to track whether it has been dequeued once
+@property(nonatomic, assign, readonly) BOOL frontDequeued;
```
davidswi: Do we really need to specify the getter in the property declaration?

SmartDeviceLink/SDLMutableDataQueue.h
davidswi: Woops, done.

SmartDeviceLink/SDLMutableDataQueue.h
asm09fsu: Personally, I feel like this is more Apple-y, and reads better.

SmartDeviceLink/SDLIAPSession.m
asm09fsu: I don't think this is necessary, as we could write a `dealloc` method for `sendDataQueue` that removes all the objects.

SmartDeviceLink/SDLIAPSession.m
asm09fsu: I'm not saying we don't need to check the state of both, I'm just saying that this bool is the same as the above, and we only use it in one spot, and because of this I don't feel as if it is necessary.

SmartDeviceLink/SDLMutableDataQueue.h
asm09fsu: Yes, we do not write the inherited initializers in the public interface. There may be some older interfaces that still have that, but they will be removed in the future.

joeljfischer:
@davidswi can we change this to be against `master` instead of `develop`. If it's against `develop` this change cannot be released until v5.0.0 is released, which is likely to be a minimum of a month.

joeljfischer:
@davidswi @asm09fsu What do you think about splitting `SDLIAPSession` into `SDLIAPDataSession` and `SDLIAPControlSession`? That seems like it would be cleaner, as `SDLIAPSession` currently does two different things.

SmartDeviceLink/SDLIAPSession.m
davidswi: I agree. Removing.

SmartDeviceLink/SDLIAPSession.m
davidswi: OK, then I would propose putting the flush call into a dealloc method for the queue because the session could go away with multiple buffers enqueued. The buffers will be held by the NSMutableArray and therefore the array may not go away, leading to the queue not going away and a leak.

SmartDeviceLink/SDLMutableDataQueue.h
davidswi: OK, reviewing the Apple Style Guide, I agree.

davidswi:
Joel, I am happy to make a PR against master rather than develop, and I agree with splitting the concerns of the control and data sessions, particularly given the different and more complex handling for the data session.

asm09fsu:
@davidswi  Yes, let's change this to be based on master, not develop. 

@joeljfischer I agree that we should work on restructuring the classes, however I agree with David in that we should probably do it after the fact.

madhuy:
@joeljfischer @asm09fsu one more improvement we can consider is,
Now we are double buffering data  
1. sendDataQueue in SDLIAPSession.m 
2._prioritizedCollection in SDLProtocol.m

if we can optimize to buffer only once and  it will help some performance improvements .

asm09fsu:
@madhuy we can do this optimization after in a new PR.

davidswi:
Folks, we did a preliminary test run of the changes in this PR using our internal SDL test apps and headunit hardware. The headunits we are testing require a Bluetooth connection and a companion app to whitelist SDL apps. If only Bluetooth is connected, the IAP transport is also over the Bluetooth link. However, if USB is connected, the headunit will disconnect the SDL IAP session from Bluetooth and establish a new SDL session over USB. For this reason, a number of our test cases involve connecting and disconnecting USB to ensure the SDL apps are able to "hot swap" SDL sessions. 

In performing hot swaps, we observed failures where the accessory connected NSNotification would be received by both the newly created SDL proxy and transport as well as the disconnected and partially dealloc'ed proxy and transport. There appears to be a race condition where the de-registration from the notification center does not happen early enough in the disconnection/dealloc code path. I have remedied this issue by de-registering from the notification center in the disconnect handler before any other actions are performed. We have tested this change and so far have not encountered any hot swap failures.

When I open the replacement PR against master, I will also include this fix.

joeljfischer:
Closed in favor of #582

Issue #424, opened 2016-06-25
closeSession seems may cause deadlock with main thread


chenbd:
### Bug Report

[Summary]
closeSession seems may cause deadlock with main thread.
not sure this is sdl_ios bug or ios system bug.
## detail log as follows:

```
Date/Time:           2016-06-24 16:02:24.24 +0800
Launch Time:         2016-06-24 15:58:32.32 +0800
OS Version:          iOS 9.3.1 (13E238)
Report Version:      105

Exception Type:  00000020
Exception Codes: 0x000000008badf00d
Exception Note:  SIMULATED (this is NOT a crash)
Highlighted by Thread:  33

Application Specific Information:
com.mxnavi.cloudnavi failed to scene-update after 10.00s

Elapsed total CPU time (seconds): 3.820 (user 3.820, system 0.000), 19% CPU 
Elapsed application CPU time (seconds): 1.532, 8% CPU

Filtered syslog:
None found

Thread 0 name:  Dispatch queue: com.apple.main-thread
Thread 0:
0   libsystem_kernel.dylib          0x00000001819bd014 semaphore_wait_trap + 8
1   libdispatch.dylib               0x000000018189a3e8 _dispatch_semaphore_wait_slow + 244
2   libxpc.dylib                    0x0000000181ad6934 xpc_connection_send_message_with_reply_sync + 204
3   IAP                             0x000000018d199f74 IAPAppRegisterClient + 716
4   IAP                             0x000000018d19a2a8 __iap2dServerLaunched(__CFNotificationCenter*, void*, __CFString const*, void const*, __CFDictionary const*) + 124
5   CoreFoundation                  0x0000000181de0eac __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 20
6   CoreFoundation                  0x0000000181de0e10 ___CFXRegistrationPost_block_invoke + 64
7   CoreFoundation                  0x0000000181df500c __CFRUNLOOP_IS_CALLING_OUT_TO_A_BLOCK__ + 20
8   CoreFoundation                  0x0000000181df4944 __CFRunLoopDoBlocks + 308
9   CoreFoundation                  0x0000000181df2d8c __CFRunLoopRun + 1960
10  CoreFoundation                  0x0000000181d1cd10 CFRunLoopRunSpecific + 384
11  GraphicsServices                0x0000000183604088 GSEventRunModal + 180
[...]

PR #582, opened 2017-03-20
IAP transport read/writes on background thread

### LINKED ISSUE #522: IAPTransport Writes Output Stream from Different Thread than Scheduled Run Loop

davidswi:
Fixes #522, #424 

This PR is ready for review.

### Risk
This PR makes no API changes.

### Testing Plan
We performed manual testing with a number of internal SDL test apps and prototype headunit hardware, covering validation and verification of correct RPC and a variety of cold and warm connect/disconnect scenarios over BT and USB. We are adding automated testing to perform longhaul characterization of MTTF.

### Summary

In addition, race conditions caused by sending events to deallocating proxy objects have been fixed. Crashes and hangs due to these race conditions should now be reduced to rare occurrences.

##### Breaking Changes
* No API breaking changes.

##### Enhancements

##### Bug Fixes

### Tasks Remaining:
None. An additional PR for properly wrapping accessory connection in a background task will be opened shortly.

### DIFF
--- a/SmartDeviceLink/SDLMutableDataQueue.m
+++ b/SmartDeviceLink/SDLMutableDataQueue.m
@@ -51,7 +51,9 @@
 
 - (void)popBuffer {
     @synchronized (self) {
-        [self.elements removeObjectAtIndex:0];
+        if (self.elements.count) {
+            [self.elements removeObjectAtIndex:0];
+        }
     }
 }
 
--- a/SmartDeviceLink/SDLMutableDataQueue.m
+++ b/SmartDeviceLink/SDLMutableDataQueue.m
@@ -1,0 +1,75 @@
+//
+//  SDLMutableDataQueue.m
+//  
+//
+//  Created by David Switzer on 3/2/17.
+//
+//
+
+#import "SDLMutableDataQueue.h"
+
+@interface SDLMutableDataQueue()
+
+@property(nonatomic, strong) NSMutableArray *elements;
+@property(nonatomic, assign, readwrite) BOOL frontDequeued;
+
+@end
+
+@implementation SDLMutableDataQueue
+
+- (instancetype)init {
+    self = [super init];
+    if (!self) {
+        return nil;
+    }
+    
+    self.elements = [[NSMutableArray alloc] init];
+    
+    return self;
+}
+
+- (void)enqueueBuffer:(NSMutableData *)data {
+    @synchronized (self) {
+        [self.elements addObject:data];
+        self.frontDequeued = NO;
+    }
+}
+
+- (NSMutableData *)frontBuffer {
+    NSMutableData *dataAtFront = nil;
+    
+    @synchronized (self) {
+        if (self.elements.count) {
+            // The front of the queue is always at index 0
+            dataAtFront = self.elements[0];
+            self.frontDequeued = YES;
+        }
+    }
+    
+    return dataAtFront;
+}
+
+- (void)popBuffer {
+    @synchronized (self) {
+        [self.elements removeObjectAtIndex:0];
+    }
+}
+
+- (void)flush {
+    @synchronized (self) {
+        [self.elements removeAllObjects];
+    }
+}
+
+- (NSUInteger)count {
+    @synchronized (self) {
+        return self.elements.count;
+    }
+}
+
+- (void)dealloc {
+    [self flush];
+    self.elements = nil;
+}
+
+@end
--- a/SmartDeviceLink/SDLProtocol.m
+++ b/SmartDeviceLink/SDLProtocol.m
@@ -229,7 +229,10 @@ typedef NSNumber SDLServiceTypeBox;
             }
         } break;
         default: {
+            /*
             NSAssert(NO, @"Attempting to send an RPC based on an unknown version number: %@, message: %@", @([SDLGlobals globals].protocolVersion), message);
+            */
+            return NO;
         } break;
     }
 
@@ -347,11 +350,18 @@ typedef NSNumber SDLServiceTypeBox;
 }
 
 - (void)processMessages {
+    SDLProtocolHeader *header;
     NSMutableString *logMessage = [[NSMutableString alloc] init];
-    UInt8 incomingVersion = [SDLProtocolMessage determineVersion:self.receiveBuffer];
-
-    // If we have enough bytes, create the header.
-    SDLProtocolHeader *header = [SDLProtocolHeader headerForVersion:incomingVersion];
+    
+    if (self.receiveBuffer != nil && self.receiveBuffer.length > 0){
+        UInt8 incomingVersion = [SDLProtocolMessage determineVersion:self.receiveBuffer];
+        // If we have enough bytes, create the header.
+        header = [SDLProtocolHeader headerForVersion:incomingVersion];
+        if (header.version == invalidSDLProtocolVersion){
+            return;
+        }
+    }
+    
     NSUInteger headerSize = header.size;
     if (self.receiveBuffer.length >= headerSize) {
         [header parse:self.receiveBuffer];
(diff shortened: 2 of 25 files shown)


joeljfischer:
This actually fixes #522, correct? So #566 can be closed?

davidswi:
Hi Joel, yes #566 can be closed. This PR also fixes #424.

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+- (NSUInteger)count {
+    @synchronized (self) {
+        return self.elements.count;
+    }
+}
+
+- (void)dealloc {
```
joeljfischer: This shouldn't be necessary

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -0,0 +1,77 @@
...
+        return nil;
+    }
+    
+    self.elements = [[NSMutableArray alloc] init];
+    
+    return self;
+}
+
+- (void)enqueueBuffer:(NSMutableData *)data {
+    @synchronized (self) {
```
joeljfischer: I assume we're synchronizing all of this to make it thread safe. Are we expecting it to be used on multiple threads? If not, synchronization is relatively slow.

SmartDeviceLink/SDLProtocol.m
```diff
@@ -229,7 +229,10 @@ - (BOOL)sendRPC:(SDLRPCMessage *)message encrypted:(BOOL)encryption error:(NSErr
             }
         } break;
         default: {
+            /*
             NSAssert(NO, @"Attempting to send an RPC based on an unknown version number: %@, message: %@", @([SDLGlobals globals].protocolVersion), message);
```
joeljfischer: Why was this commented out? For debug purposes? Asserts only occur in DEBUG.

SmartDeviceLink/SDLProtocol.m
```diff
@@ -347,11 +350,18 @@ - (void)handleBytesFromTransport:(NSData *)receivedData {
 }
 
 - (void)processMessages {
+    SDLProtocolHeader *header;
```
joeljfischer: This should initialize explicitly with `nil` if necessary.

SmartDeviceLink/SDLProtocol.m
```diff
@@ -347,11 +350,18 @@ - (void)handleBytesFromTransport:(NSData *)receivedData {
...
     NSMutableString *logMessage = [[NSMutableString alloc] init];
-    UInt8 incomingVersion = [SDLProtocolMessage determineVersion:self.receiveBuffer];
-
-    // If we have enough bytes, create the header.
-    SDLProtocolHeader *header = [SDLProtocolHeader headerForVersion:incomingVersion];
+    
+    if (self.receiveBuffer != nil && self.receiveBuffer.length > 0){
```
joeljfischer: This should be flipped and return early.

SmartDeviceLink/SDLProtocolHeader.h
```diff
@@ -37,6 +37,8 @@ typedef NS_ENUM(UInt8, SDLFrameData) {
     SDLFrameData_ConsecutiveLastFrame = 0x00
 };
 
+UInt8 const invalidSDLProtocolVersion = 255;
```
joeljfischer: Should be capitalized. And equal UINT8_MAX instead.

SmartDeviceLink/SDLProtocolHeader.m
```diff
@@ -59,8 +58,12 @@ + (__kindof SDLProtocolHeader *)headerForVersion:(UInt8)version {
             return [[SDLV2ProtocolHeader alloc] initWithVersion:version];
         } break;
         default: {
+#ifdef DEBUG
             NSString *reason = [NSString stringWithFormat:@"The version of header that is being created is unknown: %@", @(version)];
-            @throw [NSException exceptionWithName:NSInvalidArgumentException reason:reason userInfo:@{ @"requestedVersion": @(version) }];
+            @throw [NSException exceptionWithName:NSInvalidArgumentException reason:reason userInfo:@{ @"requestedVersion" : @(version) }];
+#else
+            return [[SDLV2ProtocolHeader alloc] initWithVersion:invalidSDLProtocolVersion];
```
joeljfischer: Would it be better to return `nil`? I'm not sure this is something we want to do in this PR. It's a change that would require some testing.

SmartDeviceLink/SDLProxy.m
```diff
@@ -127,8 +124,8 @@ - (void)dispose {
 }
 
 - (void)dealloc {
-    [self destructObjects];
     [SDLDebugTool logInfo:@"SDLProxy Dealloc" withType:SDLDebugType_RPC toOutput:SDLDebugOutput_All toGroup:_debugConsoleGroupName];
+    [self destructObjects];
```
joeljfischer: Why this change?

SmartDeviceLink/SDLProxy.m
```diff
@@ -416,17 +413,19 @@ - (void)handleSystemRequest:(NSDictionary *)dict {
...
-    // Handle the various OnSystemRequest types
-    if (requestType == [SDLRequestType PROPRIETARY]) {
-        [self handleSystemRequestProprietary:systemRequest];
-    } else if (requestType == [SDLRequestType LOCK_SCREEN_ICON_URL]) {
-        [self handleSystemRequestLockScreenIconURL:systemRequest];
-    } else if (requestType == [SDLRequestType HTTP]) {
-        [self sdl_handleSystemRequestHTTP:systemRequest];
-    } else if (requestType == [SDLRequestType LAUNCH_APP]) {
-        [self sdl_handleSystemRequestLaunchApp:systemRequest];
+    if (systemRequest != nil) {
```
joeljfischer: This should never return nil. An assert would be better here. Have you seen this happen?

SmartDeviceLink/SDLProxy.m
```diff
@@ -694,9 +693,11 @@ - (void)removeDelegate:(NSObject<SDLProxyListener> *)delegate {
 }
 
 - (void)invokeMethodOnDelegates:(SEL)aSelector withObject:(id)object {
+    __weak SDLProxy *weakSelf = self;
+    
     dispatch_async(dispatch_get_main_queue(), ^{
-        @autoreleasepool {
-            for (id<SDLProxyListener> listener in self.proxyListeners) {
+        if (!_alreadyDestructed){
```
joeljfischer: This should never be necessary, that whole process, including this variable, shouldn't exist. Have you seen errors based on this?

Why was `@autoreleasepool` removed? It should not have been.

SmartDeviceLink/SDLStreamDelegate.m
```diff
@@ -89,4 +90,14 @@ - (void)stream:(NSStream *)stream handleEvent:(NSStreamEvent)eventCode {
     [SDLDebugTool logInfo:@"Stream Event End"];
 };
 
+- (void)dealloc{
```
joeljfischer: Unnecessary, handled automatically by ARC.

SmartDeviceLink/SDLURLRequestTask.m
```diff
@@ -16,7 +16,7 @@ @interface SDLURLRequestTask () <NSURLConnectionDelegate, NSURLConnectionDataDel
 
 @property (strong, nonatomic) NSURLConnection *connection;
 @property (strong, nonatomic, nullable) NSURLResponse *response;
-@property (copy, nonatomic) SDLURLConnectionRequestCompletionHandler completionHandler;
+@property (copy, nonatomic, nullable) SDLURLConnectionRequestCompletionHandler completionHandler;
```
joeljfischer: Can't be nil, it's required in the only allowable `init`.

SmartDeviceLink/SDLURLRequestTask.m
```diff
@@ -69,6 +69,7 @@ - (void)sdl_addData:(NSData *)data {
 #pragma mark - Cancel
 
 - (void)cancel {
+    self.completionHandler = nil;
```
joeljfischer: Will cause the completion handler to not be called with the error.

joeljfischer:
Closed for #597

PR #592, opened 2017-03-29
Add background task handling per Apple guidelines for connection.

### LINKED ISSUE #591: IAP Transport Background Connect Compliance
### Bug Report
If an SDL app is connecting to the head unit in the background and the connection process takes more than 10s, the app will be suspended per Apple's background task guidelines. https://developer.apple.com/library/content/documentation/iPhone/Conceptual/iPhoneOSProgrammingGuide/BackgroundExecution/BackgroundExecution.html. 

Upon an accessory connect in the background or transition to the background while the connection is in progress, the SDL proxy assumes it can run an NSTimer in SDLIAPTransport.m, and that it can performed a delayed connect selector call based on the computed retryDelay value. These assumptions are invalid when the app is backgrounded, unless it registers a background task to contain the connection logic. If the connection process completes through to the data session in less than 10s, iOS will allow the SDL app to execute based on the active accessory session. If it does not, the SDL app will be suspended indefinitely.

##### Reproduction Steps
1. Power off the head unit (not suspended) with navigation running to ensure the session resumes and consumes CPU on the head unit.
2. Background the SDL app
3. Power on the head unit from full hardware reset
4. Wait for the SDL app to connect in the background, with the iOS device either connected to the Xcode debugger or redirecting debug logs over NSLogger.

##### Expected Behavior
The SDL app should always connect to a working data session.

##### Observed Behavior
If the head unit takes more than 10s to negotiate an SDL data session with the iOS device, the SDL app is suspended indefinitely.

* iOS Version: All 9.x and above
* SDL iOS Version: 4.5.5
* Testing Against: Prototype head unit samples

Choose any SDL app, including the SmartDeviceLink Example.

davidswi:
Fixes #591 

This PR is ready for review.

### Risk
This PR makes no API changes. However, it does change background behavior and can cause simulated crashes if the background task is not cleaned up properly. 

### Testing Plan
Extensive testing has been done on this PR within Xevo using engineering sample head units.

### Summary
An app backgrounded notification handler is added to the SDLIAPTransport, along with a background task handler. If the app is in the background when it start connection to the head unit or is backgrounded during connection, a background task is started. The task is stopped when the app is foregrounded, the data session is established or the proxy is destroyed.

##### Breaking Changes
None

##### Enhancements
None

##### Bug Fixes
The SDL app should either succeed in connecting when in the background or stop trying based on the retry counter and protocol index timer timeout period.

### Tasks Remaining:
None


SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -84,6 +88,21 @@ - (void)sdl_stopEventListening {
     [SDLDebugTool logInfo:@"SDLIAPTransport Stopped Listening For Events"];
     [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
+- (void)sdl_backgroundTaskStart {
+    if (self.backgroundTaskId == UIBackgroundTaskInvalid) {
```
joeljfischer: This should go the other way, `self.backgroundTaskId != UIBackgroundTaskInvalid`, and return early.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -84,6 +88,21 @@ - (void)sdl_stopEventListening {
...
+        self.backgroundTaskId = [[UIApplication sharedApplication] beginBackgroundTaskWithName:@"SDLIAPConnectionLoop" expirationHandler:^{
+            [self sdl_backgroundTaskEnd];
+        }];
+    }
+}
+
+- (void)sdl_backgroundTaskEnd {
+    if (self.backgroundTaskId != UIBackgroundTaskInvalid) {
```
joeljfischer: This should go the other way, `self.backgroundTaskId == UIBackgroundTaskInvalid`, and return early.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -102,17 +125,29 @@ - (void)sdl_accessoryDisconnected:(NSNotification *)notification {
     EAAccessory *accessory = [notification.userInfo objectForKey:EAAccessoryKey];
     if (accessory.connectionID == self.session.accessory.connectionID) {
         self.sessionSetupInProgress = NO;
-        [self disconnect];
+        
+	[self sdl_backgroundTaskEnd];
+
+	[self disconnect];
```
joeljfischer: These lines can be together, and they appear to not be correctly aligned.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -462,6 +498,7 @@ - (void)sdl_destructObjects {
         self.controlSession = nil;
         self.session = nil;
         self.delegate = nil;
+	[self sdl_backgroundTaskEnd];
```
joeljfischer: This appears to not be aligned properly.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -102,17 +125,29 @@ - (void)sdl_accessoryDisconnected:(NSNotification *)notification {
...
 - (void)sdl_applicationWillEnterForeground:(NSNotification *)notification {
     [SDLDebugTool logInfo:@"App Foregrounded Event" withType:SDLDebugType_Transport_iAP toOutput:SDLDebugOutput_All toGroup:self.debugConsoleGroupName];
+    [self sdl_backgroundTaskEnd];
     self.retryCounter = 0;
     [self connect];
 }
 
+- (void)sdl_applicationDidEnterBackground:(NSNotification *)notification {
+    [SDLDebugTool logInfo:@"App Backgrounded Event" withType:SDLDebugType_Transport_iAP toOutput:SDLDebugOutput_All toGroup:self.debugConsoleGroupName];
+    if (self.isDelayedConnect) {
```
joeljfischer: This property doesn't appear to exist.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -34,6 +34,8 @@ @interface SDLIAPTransport () {
 @property (assign) int retryCounter;
 @property (assign) BOOL sessionSetupInProgress;
 @property (strong) SDLTimer *protocolIndexTimer;
+@property (nonatomic, assign) UIBackgroundTaskIdentifier backgroundTaskId;
+@property (assign) BOOL isDelayedConnect;
```
joeljfischer: Are both `backgroundTaskId` and `isDelayedConnect` needed? I really dislike state booleans, and any way to get rid of them and provide stateless code should be taken. They make unit testing extremely difficult. Is it possible to use the `backgroundTaskId` and check whether it's `invalid` to know if we're in a delayed connect?

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -85,13 +90,35 @@ - (void)sdl_stopEventListening {
...
+
 #pragma mark - EAAccessory Notifications
 
 - (void)sdl_accessoryConnected:(NSNotification *)notification {
     NSMutableString *logMessage = [NSMutableString stringWithFormat:@"Accessory Connected, Opening in %0.03fs", self.retryDelay];
     [SDLDebugTool logInfo:logMessage withType:SDLDebugType_Transport_iAP toOutput:SDLDebugOutput_All toGroup:self.debugConsoleGroupName];
 
     self.retryCounter = 0;
+    if ([UIApplication sharedApplication].applicationState == UIApplicationStateBackground) {
+        [self sdl_backgroundTaskStart];
```
joeljfischer: Instead of monitoring background and starting the task then, we should just wrap our connection in a background task, in other words, always start a background task before we are trying to connect, and always cancel it after we've connected / dealloc. There's no cost to starting background tasks, and it'll just automatically handle it when we go into the background. Then all of the background monitoring code can be removed.

SmartDeviceLink/SDLIAPTransport.m
davidswi: Yes, we could do that. I will remove the isDelayedConnect boolean.

SmartDeviceLink/SDLIAPTransport.m
davidswi: Sounds good, I will make that change.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -85,13 +87,35 @@ - (void)sdl_stopEventListening {
...
+        return;
+    }
+    
+    self.backgroundTaskId = [[UIApplication sharedApplication] beginBackgroundTaskWithName:backgroundTaskName expirationHandler:^{
+        [self sdl_backgroundTaskEnd];
+    }];
+}
+
+- (void)sdl_backgroundTaskEnd {
+    if (self.backgroundTaskId != UIBackgroundTaskInvalid) {
```
joeljfischer: This should be flipped to return early, like in `backgroundTaskStart`

SmartDeviceLink/SDLIAPTransport.m
joeljfischer: This shouldn't be here along with L112

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -102,13 +126,15 @@ - (void)sdl_accessoryDisconnected:(NSNotification *)notification {
...
         self.sessionSetupInProgress = NO;
+        [self sdl_backgroundTaskEnd];
         [self disconnect];
         [self.delegate onTransportDisconnected];
     }
 }
 
 - (void)sdl_applicationWillEnterForeground:(NSNotification *)notification {
     [SDLDebugTool logInfo:@"App Foregrounded Event" withType:SDLDebugType_Transport_iAP toOutput:SDLDebugOutput_All toGroup:self.debugConsoleGroupName];
+    [self sdl_backgroundTaskEnd];
```
joeljfischer: We don't need to call this here.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -247,6 +273,7 @@ - (void)onSessionInitializationCompleteForSession:(SDLIAPSession *)session {
     // Data Session Opened
     if (![controlProtocolString isEqualToString:session.protocol]) {
         self.sessionSetupInProgress = NO;
+        [self sdl_backgroundTaskEnd];
```
joeljfischer: I don't think we need to call this here.

SmartDeviceLink/SDLIAPTransport.m
davidswi: We do need to explicitly end background tasks when we are done with background processing. The expiration handler is only our last chance to end the background task before the OS force terminates the app. I've thought carefully about the cases where the task should be cleaned up. It's possible I'm missing something, so I'd love feedback on why you believe we don't need the call to occur when switching to the foreground.

SmartDeviceLink/SDLIAPTransport.m
davidswi: Per my last comment, I would greatly appreciate understanding why we would not end the background task upon successful connection of the data session.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -20,6 +20,7 @@
 NSString *const legacyProtocolString = @"com.ford.sync.prot0";
 NSString *const controlProtocolString = @"com.smartdevicelink.prot0";
 NSString *const indexedProtocolStringPrefix = @"com.smartdevicelink.prot";
+NSString *const backgroundTaskName = @"com.sdl.transport.iap.connectloop";
```
joeljfischer: Can we change this to `@"com.sdl.transport.iap.backgroundTask"`

SmartDeviceLink/SDLIAPTransport.m
joeljfischer: Based on all of my research, this is something handled by Apple when we go into the foreground. I have not seen any sample code in which the task is ended on foregrounding. The only thing I have seen happen is when the app becomes active, that the background task is started if it was ever ended. The sample code I have looked at has the background task end in the expiration handler and whenever the task is fully completed. Let me know if you've seen anything describing usage differently.

SmartDeviceLink/SDLIAPTransport.m
joeljfischer: I think you are correct, but I think the better place for this would be in the setter for `sessionSetupInProgress`. If it goes to `NO`, then the task must end. I say this to make sure you hit all cases, because I believe there may be some cases where the accessory fails to connect and we don't complete the task. Unless that is desired to attempt to connect to other accessories?

SmartDeviceLink/SDLIAPTransport.m
davidswi: http://stackoverflow.com/questions/29919573/beginbackgroundtaskwithexpirationhandler-calling-endbackgroundtask-but-not-endin

In the above article, it appears that not explicitly ending the background task when the app foregrounds can lead to the app being terminated.

SmartDeviceLink/SDLIAPTransport.m
davidswi: OK, makes sense.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -42,6 +42,8 @@ @interface SDLIAPTransport () {
 
 @implementation SDLIAPTransport
 
+@synthesize sessionSetupInProgress = _sessionSetupInProgress;
```
joeljfischer: Remove (along with comment below to fix the error it causes).

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -106,6 +108,20 @@ - (void)sdl_backgroundTaskEnd {
...
+
+- (void)setSessionSetupInProgress:(BOOL)inProgress{
+    _sessionSetupInProgress = inProgress;
+    if (!inProgress){
+        // End the background task here to catch all cases
+        [self sdl_backgroundTaskEnd];
+    }
+}
+
+- (BOOL)sessionSetupInProgress{
```
joeljfischer: Remove this method.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -106,6 +108,20 @@ - (void)sdl_backgroundTaskEnd {
     self.backgroundTaskId = UIBackgroundTaskInvalid;
 }
 
+#pragma mark session setup getter/setter
+
+- (void)setSessionSetupInProgress:(BOOL)inProgress{
```
joeljfischer: Space between `inProgress` and `{`

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -108,20 +106,6 @@ - (void)sdl_backgroundTaskEnd {
     self.backgroundTaskId = UIBackgroundTaskInvalid;
 }
 
-#pragma mark session setup getter/setter
-
-- (void)setSessionSetupInProgress:(BOOL)inProgress{
-    _sessionSetupInProgress = inProgress;
-    if (!inProgress){
-        // End the background task here to catch all cases
-        [self sdl_backgroundTaskEnd];
```
joeljfischer: Wasn't this a necessary addition based on the last review, or no?

davidswi: Yes it was, but I wasn't fully clear on what you were asking me to remove. Can you clarify?

joeljfischer: Only the `@synthesize` line and the getter had to be removed, the setter could stay as is.

davidswi: Thanks will do.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -32,8 +33,9 @@ @interface SDLIAPTransport () {
 }
 
 @property (assign) int retryCounter;
-@property (assign) BOOL sessionSetupInProgress;
 @property (strong) SDLTimer *protocolIndexTimer;
+@property (nonatomic, assign) UIBackgroundTaskIdentifier backgroundTaskId;
+@property (nonatomic, assign, setter=setSessionSetupInProgress:) BOOL sessionSetupInProgress;
```
joeljfischer: `, setter=setSessionSetupInProgress:` is superfluous, it doesn't add or change anything.

joeljfischer:
@davidswi After merging #597 to master, this has conflicts, please merge master into this branch, resolve the merge conflicts, and retest on your end. Thanks!

joeljfischer:
Did not intend to close that, whoopsie.

joeljfischer:
@davidswi 

I'm seeing repeatable crashes with the Ford Sync 3 TDK (v. 2.2, Build 16208_DEVTEST). Here's my process with the SDL example app:

1. Connect via USB
2. Connect the app with the "Connect" button
3. Open the app on the HMI
4. Use the app a bit
5. Disconnect the USB from the phone (the app UI appropriately shows that it is no longer connected)
6. Plug back in the USB
7. Press the connect button again.

This causes a hard crash in my tests. Example crash logs are attached below.

[SDL Example  5-19-17, 11-14 AM.txt](https://github.com/smartdevicelink/sdl_ios/files/1014741/SDL.Example.5-19-17.11-14.AM.txt)
[SDL Example  5-19-17, 11-24 AM.txt](https://github.com/smartdevicelink/sdl_ios/files/1014740/SDL.Example.5-19-17.11-24.AM.txt)

davidswi:
Joel, thanks for testing this. It appears that the stream delegate has been deallocated but the input stream has not been torn down. I've seen these race conditions in prior iterations of the threaded I/O changes and will need to review my notes. 

I can say that we do not see these in our codebase because I kept in a number of the deallocs we agreed to take out in the PR. What we are still seeing are intermittent crashes in the iap2d process, which indicates to me we still have race condition(s) somewhere.

This is what I would propose we do:

1. Build the SDL example app with zombies turned on so we can see which object was being called. I suspect it is the stream delegate instance.
2. I have a complete mock EAAccessoryFramework done that is able to respond to control and data session frames and packets. I think we should use the mock framework to validate all changes to the IAP transport.
3. If the iap2d crashes cannot be root caused, I would like to work onsite at Apple to resolve them.

joeljfischer:
Could you add back in the `dealloc`s you have and I'll retest and see if those still exist?

1. I agree that it is likely the stream delegate instance.
2. I agree, is that a mock framework that we can build unit tests around?
3. Sounds about right, I assume you have a point of contact? I do not, but I know Ford does.

davidswi:
2. Yes
3. I was going to start with Craig Keithley and let him redirect to Emily Schubert.

davidswi:
I think we should also discuss timing of refactor into separate control and data session classes.

joeljfischer:
This PR, assuming it's approved by June 1, will be a part of 4.6, which will have a release candidate June 1 for release June 30.

The splitting apart will probably have to happen as part of 5.0, which is an August 1 - August 31 beta to release timeframe. 5.0 code is currently on the develop branch, so I would say anytime after June 1 would be a good time to start the splitting of the IAP classes.

PR #597, opened 2017-04-02
IAP transport read/writes on background thread

### LINKED ISSUE #522: IAPTransport Writes Output Stream from Different Thread than Scheduled Run Loop

davidswi:
Fixes #522, #424 

This PR is ready for review.

### Risk
This PR makes no API changes.

### Testing Plan
We performed manual testing with a number of internal SDL test apps and prototype headunit hardware, covering validation and verification of correct RPC and a variety of cold and warm connect/disconnect scenarios over BT and USB. We are adding automated testing to perform longhaul characterization of MTTF.

### Summary

In addition, race conditions caused by sending events to deallocating proxy objects have been fixed. Crashes and hangs due to these race conditions should now be reduced to rare occurrences.

##### Breaking Changes
* No API breaking changes.

##### Enhancements

##### Bug Fixes

### Tasks Remaining:
Additional PRs for remaining sources of crashes and hangs in SDLIAPTransport will be created along with supporting issues.

### DIFF
--- a/SmartDeviceLink/SDLIAPSession.m
+++ b/SmartDeviceLink/SDLIAPSession.m
@@ -200,6 +200,7 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
 
 - (void)startStream:(NSStream *)stream {
     stream.delegate = self.streamDelegate;
+    NSAssert((self.isDataSession && [[NSThread currentThread] isEqual:self.ioStreamThread]) || [NSThread isMainThread], @"startStream is being called on the wrong thread!!!");
     [stream scheduleInRunLoop:[NSRunLoop currentRunLoop] forMode:NSDefaultRunLoopMode];
     [stream open];
 }
@@ -210,6 +211,8 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
 
     // When you disconect the cable you get a stream end event and come here but stream is already in closed state.
     // Still need to remove from run loop.
+    
+    NSAssert((self.isDataSession && [[NSThread currentThread] isEqual:self.ioStreamThread]) || [NSThread isMainThread], @"startStream is being called on the wrong thread!!!");
 
     NSUInteger status1 = stream.streamStatus;
     if (status1 != NSStreamStatusNotOpen &&
@@ -283,18 +286,6 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
             [strongSelf sdl_handleOutputStreamWriteError:sendErr];
         }
     };
-}
-
-#pragma mark - Lifecycle Destruction
-
-- (void)dealloc {
-    self.sendDataQueue = nil;
-    self.delegate = nil;
-    self.accessory = nil;
-    self.protocol = nil;
-    self.streamDelegate = nil;
-    self.easession = nil;
-    [SDLDebugTool logInfo:@"SDLIAPSession Dealloc"];
 }
 
 @end
--- a/SmartDeviceLink/SDLIAPSession.m
+++ b/SmartDeviceLink/SDLIAPSession.m
@@ -65,7 +65,7 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
         strongSelf.streamDelegate.streamErrorHandler = [self streamErroredHandler];
         strongSelf.streamDelegate.streamOpenHandler = [self streamOpenedHandler];
         if (self.isDataSession) {
-            self.streamDelegate.streamHasSpaceHandler = [self streamHasSpaceHandler];
+            self.streamDelegate.streamHasSpaceHandler = [self sdl_streamHasSpaceHandler];
             // Start I/O event loop processing events in iAP channel
             self.ioStreamThread = [[NSThread alloc] initWithTarget:self selector:@selector(sdl_accessoryEventLoop) object:nil];
             [self.ioStreamThread setName:IOStreamThreadName];
@@ -101,6 +101,10 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
         [self stopStream:self.easession.inputStream];
     }
     self.easession = nil;
+}
+
+- (BOOL)isStopped {
+    return !self.isOutputStreamOpen && !self.isInputStreamOpen;
 }
 
 # pragma mark - data send methods
@@ -140,7 +144,7 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
     if (self.easession == nil ||
         self.easession.outputStream == nil ||
         self.easession.outputStream.streamStatus != NSStreamStatusOpen) {
-        [self.sendDataQueue flush];
+        [self.sendDataQueue removeAllObjects];
     }
 }
 
@@ -184,12 +188,11 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
         return;
     }
     
-    NSLog(@"Close EASession: %tu", self.easession.accessory.connectionID);
-    NSInputStream *inStream = [self.easession inputStream];
-    NSOutputStream *outStream = [self.easession outputStream];
+    NSString *closeSessionString = [NSString stringWithFormat:@"Close EASession: %tu", self.easession.accessory.connectionID];
+    [SDLDebugTool logInfo:closeSessionString];
 
-    [self stopStream:inStream];
-    [self stopStream:outStream];
+    [self stopStream:[self.easession inputStream]];
+    [self stopStream:[self.easession outputStream]];
 }
 
 
@@ -221,8 +224,10 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
     if (status2 == NSStreamStatusClosed) {
         if (stream == [self.easession inputStream]) {
             [SDLDebugTool logInfo:@"Input Stream Closed"];
+            self.isInputStreamOpen = NO;
         } else if (stream == [self.easession outputStream]) {
             [SDLDebugTool logInfo:@"Output Stream Closed"];
+            self.isOutputStreamOpen = NO;
         }
     }
 }
@@ -262,18 +267,20 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
     };
 }
 
-- (SDLStreamHasSpaceHandler)streamHasSpaceHandler {
+- (SDLStreamHasSpaceHandler)sdl_streamHasSpaceHandler {
     __weak typeof(self) weakSelf = self;
     
     return ^(NSStream *stream) {
         __strong typeof(weakSelf) strongSelf = weakSelf;
         
-        if (strongSelf.isDataSession) {
-            NSError *sendErr = nil;
-            
-            if (![strongSelf sdl_dequeueAndWriteToOutputStream:&sendErr] && sendErr != nil) {
-                [strongSelf sdl_handleOutputStreamWriteError:sendErr];
-            }
+        if (!strongSelf.isDataSession){
+            return;
+        }
+        
+        NSError *sendErr = nil;
+        
+        if (![strongSelf sdl_dequeueAndWriteToOutputStream:&sendErr] && sendErr != nil) {
+            [strongSelf sdl_handleOutputStreamWriteError:sendErr];
         }
     };
 }
--- a/SmartDeviceLink/SDLIAPSession.m
+++ b/SmartDeviceLink/SDLIAPSession.m
@@ -212,7 +212,7 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
     // When you disconect the cable you get a stream end event and come here but stream is already in closed state.
     // Still need to remove from run loop.
     
-    NSAssert((self.isDataSession && [[NSThread currentThread] isEqual:self.ioStreamThread]) || [NSThread isMainThread], @"startStream is being called on the wrong thread!!!");
+    NSAssert((self.isDataSession && [[NSThread currentThread] isEqual:self.ioStreamThread]) || [NSThread isMainThread], @"stopStream is being called on the wrong thread!!!");
 
     NSUInteger status1 = stream.streamStatus;
     if (status1 != NSStreamStatusNotOpen &&
--- a/SmartDeviceLink/SDLIAPSession.m
+++ b/SmartDeviceLink/SDLIAPSession.m
@@ -6,12 +6,19 @@
 #import "SDLDebugTool.h"
 #import "SDLStreamDelegate.h"
 #import "SDLTimer.h"
+#import "SDLMutableDataQueue.h"
 
+NSString *const ioStreamThreadName  = @ "com.smartdevicelink.iostream";
+NSTimeInterval const streamThreadWaitSecs = 1.0;
 
 @interface SDLIAPSession ()
 
 @property (assign) BOOL isInputStreamOpen;
 @property (assign) BOOL isOutputStreamOpen;
+@property (assign, nonatomic) BOOL isDataSession;
+@property (nonatomic, strong, nullable) NSThread *ioStreamThread;
+@property (nonatomic, strong, nullable) SDLMutableDataQueue *sendDataQueue;
+@property (nonatomic, strong, nullable) dispatch_semaphore_t canceledSemaphore;
 
 @end
 
@@ -27,8 +34,11 @@
     self = [super init];
     if (self) {
         _delegate = nil;
+        _isDataSession = [protocol isEqualToString:@"com.smartdevicelink.prot0"] ? NO : YES;
         _accessory = accessory;
         _protocol = protocol;
+        _canceledSemaphore = dispatch_semaphore_create(0);
+        _sendDataQueue = [[SDLMutableDataQueue alloc] init];
         _streamDelegate = nil;
         _easession = nil;
         _isInputStreamOpen = NO;
@@ -43,19 +53,26 @@
 - (BOOL)start {
     __weak typeof(self) weakSelf = self;
 
-    NSString *logMessage = [NSString stringWithFormat:@"Opening EASession withAccessory:%@ forProtocol:%@", _accessory.name, _protocol];
+    NSString *logMessage = [NSString stringWithFormat:@"Opening EASession withAccessory:%@ forProtocol:%@", self.accessory.name, self.protocol];
     [SDLDebugTool logInfo:logMessage];
 
-    if ((self.easession = [[EASession alloc] initWithAccessory:_accessory forProtocol:_protocol])) {
+    if ((self.easession = [[EASession alloc] initWithAccessory:self.accessory forProtocol:self.protocol])) {
         __strong typeof(self) strongSelf = weakSelf;
 
         [SDLDebugTool logInfo:@"Created Session Object"];
 
         strongSelf.streamDelegate.streamErrorHandler = [self streamErroredHandler];
         strongSelf.streamDelegate.streamOpenHandler = [self streamOpenedHandler];
-
-        [strongSelf startStream:weakSelf.easession.outputStream];
-        [strongSelf startStream:weakSelf.easession.inputStream];
+        if (!self.isDataSession) {
+            [self startStream:self.easession.outputStream];
+            [self startStream:self.easession.inputStream];
+        } else {
+            self.streamDelegate.streamHasSpaceHandler = [self streamHasSpaceHandler];
+            // Start I/O event loop processing events in iAP channel
+            self.ioStreamThread = [[NSThread alloc] initWithTarget:self selector:@selector(sdl_accessoryEventLoop) object:nil];
+            [self.ioStreamThread setName:ioStreamThreadName];
+            [self.ioStreamThread start];
+        }
 
         return YES;
 
@@ -66,9 +83,109 @@
 }
 
 - (void)stop {
-    [self stopStream:self.easession.outputStream];
-    [self stopStream:self.easession.inputStream];
+
+    if (!self.isDataSession) {
+        [self stopStream:self.easession.outputStream];
+        [self stopStream:self.easession.inputStream];
+    } else {
+        [self.ioStreamThread cancel];
+        
+        long lWait = dispatch_semaphore_wait(self.canceledSemaphore, dispatch_time(DISPATCH_TIME_NOW, streamThreadWaitSecs * NSEC_PER_SEC));
+        if (lWait == 0) {
+            [SDLDebugTool logInfo:@"Stream thread canceled"];
+        } else{
+           [SDLDebugTool logInfo:@"Error: failed to cancel stream thread"];
+        }
+        self.ioStreamThread = nil;
+        self.isDataSession = NO;
+    }
     self.easession = nil;
+}
+
+- (void)sendData:(NSData *)data {
+    // Enqueue the data for transmission on the IO thread
+    [self.sendDataQueue enqueueBuffer:data.mutableCopy];
+}
+
+- (BOOL)sdl_dequeueAndWriteToOutputStream:(NSError **)error {
+    NSOutputStream *ostream = self.easession.outputStream;
+    NSMutableData *remainder = [self.sendDataQueue frontBuffer];
+    BOOL allDataWritten = NO;
+    
+    if (error != nil && remainder != nil && ostream.streamStatus == NSStreamStatusOpen) {
+        NSInteger bytesRemaining = remainder.length;
+        NSInteger bytesWritten = [ostream write:remainder.bytes maxLength:bytesRemaining];
+        if (bytesWritten < 0) {
+            *error = ostream.streamError;
+        } else if (bytesWritten == bytesRemaining) {
+                // Remove the data from the queue
+                [self.sendDataQueue popBuffer];
+                allDataWritten = YES;
+            } else {
+                // Cleave the sent bytes from the data, the remainder will sit at the head of the queue
+                [remainder replaceBytesInRange:NSMakeRange(0, bytesWritten) withBytes:NULL length:0];
+            }
+    }
+
+    return allDataWritten;
+}
+
+- (void)sdl_handleOutputStreamWriteError:(NSError *)error {
+    NSString *errString = [NSString stringWithFormat:@"Output stream error: %@", error];
+    [SDLDebugTool logInfo:errString];
+    // REVIEW: We should look at the domain and the code as a tuple and decide
+    // how to handle the error based on both values. For now, if the stream
+    // is closed, we will flush the send queue and leave it as-is otherwise
+    // so that temporary error conditions can be dealt with by retrying
+    if (self.easession == nil ||
+        self.easession.outputStream == nil ||
+        self.easession.outputStream.streamStatus != NSStreamStatusOpen) {
+        [self.sendDataQueue flush];
+    }
+}
+
+- (void)sdl_accessoryEventLoop {
+    @autoreleasepool {
+        NSAssert(self.easession, @"_session must be assigned before calling");
+
+        if (!self.easession) {
+          return;
+        }
+        
+        [self startStream:self.easession.inputStream];
+        [self startStream:self.easession.outputStream];
+
+        [SDLDebugTool logInfo:@"starting the event loop for accessory"];
+        do {
+            if (self.sendDataQueue.count > 0 && !self.sendDataQueue.frontDequeued) {
+                NSError *sendErr = nil;
+                if (![self sdl_dequeueAndWriteToOutputStream:&sendErr] && sendErr != nil) {
+                    [self sdl_handleOutputStreamWriteError:sendErr];
+                }
+            }
+            
+            [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode
+                                     beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.25f]];
+        } while (![NSThread currentThread].cancelled);
+
+        NSLog(@"closing accessory session");
+
+        // Close I/O streams of the iAP session
+        [self sdl_closeSession];
+        dispatch_semaphore_signal(self.canceledSemaphore);
+    }
+}
+
+// Must be called on accessoryEventLoop.
+- (void)sdl_closeSession {
+  if (self.easession) {
+    NSLog(@"Close EASession: %tu", self.easession.accessory.connectionID);
+    NSInputStream *inStream = [self.easession inputStream];
+    NSOutputStream *outStream = [self.easession outputStream];
+    
+    [self stopStream:inStream];
+    [self stopStream:outStream];
+  }
 }
 
 
@@ -76,7 +193,7 @@
 
 - (void)startStream:(NSStream *)stream {
     stream.delegate = self.streamDelegate;
-    [stream scheduleInRunLoop:[NSRunLoop mainRunLoop] forMode:NSDefaultRunLoopMode];
+    [stream scheduleInRunLoop:[NSRunLoop currentRunLoop] forMode:NSDefaultRunLoopMode];
     [stream open];
 }
 
@@ -93,7 +210,7 @@
         [stream close];
     }
 
-    [stream removeFromRunLoop:[NSRunLoop mainRunLoop] forMode:NSDefaultRunLoopMode];
+    [stream removeFromRunLoop:[NSRunLoop currentRunLoop] forMode:NSDefaultRunLoopMode];
     [stream setDelegate:nil];
 
     NSUInteger status2 = stream.streamStatus;
@@ -141,10 +258,26 @@
     };
 }
 
+- (SDLStreamHasSpaceHandler)streamHasSpaceHandler {
+    __weak typeof(self) weakSelf = self;
+    
+    return ^(NSStream *stream) {
+        __strong typeof(weakSelf) strongSelf = weakSelf;
+        
+        if (strongSelf.isDataSession) {
+            NSError *sendErr = nil;
+            
+            if (![strongSelf sdl_dequeueAndWriteToOutputStream:&sendErr] && sendErr != nil) {
+                [strongSelf sdl_handleOutputStreamWriteError:sendErr];
+            }
+        }
+    };
+}
 
 #pragma mark - Lifecycle Destruction
 
 - (void)dealloc {
+    self.sendDataQueue = nil;
     self.delegate = nil;
     self.accessory = nil;
     self.protocol = nil;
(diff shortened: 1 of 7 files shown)


joeljfischer:
@davidswi In the future if possible, please continue to use the same PR for the same change, so that I can track my comments and see what you've changed. It's going to be much harder to review this PR because you closed and reopened a new one and I can't easily track what's different between when I put my comments there and now.

davidswi:
Understood @joeljfischer, I have one more change related to this PR. The control session is stopped on the receive dispatch queue, but the change I made to [SDLIAPSession stopStream] assumes it is being called on the thread that scheduled the stream, which should be the main thread. I would like to commit the fix for this issue and close out the PR unless you would like me to submit that change in another PR and/or have other concerns regarding this PR.

joeljfischer:
I'm not sure what you mean by "close out". Was this PR not ready for review? I'm making minor comments to this code in a review at this moment and have not done an in depth check or testing yet. That fix should go into this branch and PR.

davidswi:
Hi @joeljfischer thanks, I will add the commit. I should have been clearer - I meant all code review issues on my side addressed.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -268,23 +304,9 @@ - (void)onSessionStreamsEnded:(SDLIAPSession *)session {
...
-                if (bytesWritten == -1) {
-                    [SDLDebugTool logInfo:[NSString stringWithFormat:@"Error: %@", [ostream streamError]] withType:SDLDebugType_Transport_iAP toOutput:SDLDebugOutput_All];
-                    break;
-                }
-
-                [remainder replaceBytesInRange:NSMakeRange(0, bytesWritten) withBytes:NULL length:0];
-            }
-        }
-    });
+    if (self.session != nil && self.session.accessory.connected){
```
joeljfischer: This should flip and return if it fails. The `{` goes a space after the `)`

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -333,9 +355,9 @@ - (SDLStreamHasBytesHandler)sdl_controlStreamHasBytesHandlerForAccessory:(EAAcce
 
             // Determine protocol string of the data session, then create that data session
             NSString *indexedProtocolString = [NSString stringWithFormat:@"%@%@", indexedProtocolStringPrefix, @(buf[0])];
-            dispatch_sync(dispatch_get_main_queue(), ^{
+            if (accessory.isConnected){
```
joeljfischer: Formatting on the `{`

Why was this added, the control stream can't send us bytes if the accessory isn't connected, right?

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -383,7 +405,7 @@ - (SDLStreamHasBytesHandler)sdl_dataStreamHasBytesHandler {
         __strong typeof(weakSelf) strongSelf = weakSelf;
 
         uint8_t buf[[SDLGlobals globals].maxMTUSize];
-        while ([istream hasBytesAvailable]) {
+        while (istream.streamStatus == NSStreamStatusOpen && istream.hasBytesAvailable) {
```
joeljfischer: Can the input stream have bytes available if we don't have an open session?

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -0,0 +1,75 @@
...
+}
+
+- (void)enqueueBuffer:(NSMutableData *)data {
+    @synchronized (self) {
+        [self.elements addObject:data];
+        self.frontDequeued = NO;
+    }
+}
+
+- (NSMutableData *)frontBuffer {
```
joeljfischer: This needs to be `nullable` if we can return `nil`.

SmartDeviceLink/SDLMutableDataQueue.h
```diff
@@ -0,0 +1,38 @@
...
+//  SDLMutableDataQueue.h
+//  
+//
+//  Created by David Switzer on 3/2/17.
+//
+//
+
+#import <Foundation/Foundation.h>
+
+@interface SDLMutableDataQueue : NSObject
```
joeljfischer: Please add some comments describing this class and why it exists.

SmartDeviceLink/SDLMutableDataQueue.h
```diff
@@ -0,0 +1,38 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+// frontBuffer returns the NSMutableData * buffer at the front of the queue,
+// but does not remove it -- modeled after the STL C++ queue front method
```
joeljfischer: Methods should use API documentation methods. In Xcode 8+, put the cursor on the method and use `cmd+option+/` to generate some helper comments.

SmartDeviceLink/SDLMutableDataQueue.h
joeljfischer: What's the purpose of this class handling `NSMutableData` instead of just `NSData`

SmartDeviceLink/SDLMutableDataQueue.h
```diff
@@ -0,0 +1,38 @@
...
+
+// popBuffer removes the buffer at the head of the queue -- modeled after the
+// STL C++ queue pop method
+- (void)popBuffer;
+
+// Enqueues a new NSMutableData * buffer at the back of the queue
+- (void)enqueueBuffer:(NSMutableData *)data;
+
+// Empty the queue
+- (void)flush;
```
joeljfischer: When I think of flush, I usually think of it returning a bunch of data. Would this be more accurate as `clear`?

SmartDeviceLink/SDLMutableDataQueue.h
```diff
@@ -0,0 +1,38 @@
...
+
+@property(nonatomic, assign, readonly) NSUInteger count;
```
joeljfischer: Needs a method comment

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -0,0 +1,75 @@
...
+//  SDLMutableDataQueue.m
+//  
+//
+//  Created by David Switzer on 3/2/17.
+//
+//
+
+#import "SDLMutableDataQueue.h"
+
+@interface SDLMutableDataQueue()
```
joeljfischer: Needs `NS_ASSUME_NONNULL`

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -0,0 +1,75 @@
...
+
+@implementation SDLMutableDataQueue
+
+- (instancetype)init {
+    self = [super init];
+    if (!self) {
+        return nil;
+    }
+    
+    self.elements = [[NSMutableArray alloc] init];
```
joeljfischer: Should be `_elements`. Can just be `[NSMutableArray array]`

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -0,0 +1,75 @@
...
+    
+    return self;
+}
+
+- (void)enqueueBuffer:(NSMutableData *)data {
+    @synchronized (self) {
```
joeljfischer: I assume we're synchronizing all of this to make it thread safe. Are we expecting it to be used on multiple threads? If not, synchronization is relatively slow.

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+- (NSUInteger)count {
+    @synchronized (self) {
+        return self.elements.count;
+    }
+}
+
+- (void)dealloc {
```
joeljfischer: This shouldn't be necessary.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -78,6 +77,8 @@ - (void)sdl_startEventListening {
                                              selector:@selector(sdl_applicationWillEnterForeground:)
                                                  name:UIApplicationWillEnterForegroundNotification
                                                object:nil];
+    
+    [[EAAccessoryManager sharedAccessoryManager] registerForLocalNotifications];
```
joeljfischer: Just want to double check that we don't need to balance this with an `unregister`. It doesn't appear to be required, but this `register` may be called several times if the app registers and unregisters.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -6,12 +6,19 @@
 #import "SDLDebugTool.h"
 #import "SDLStreamDelegate.h"
 #import "SDLTimer.h"
+#import "SDLMutableDataQueue.h"
```
davidswi: Done.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -6,12 +6,19 @@
...
 
+NSString *const ioStreamThreadName  = @ "com.smartdevicelink.iostream";
```
davidswi: Done

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -6,12 +6,19 @@
...
+NSTimeInterval const streamThreadWaitSecs = 1.0;
 
 @interface SDLIAPSession ()
 
 @property (assign) BOOL isInputStreamOpen;
 @property (assign) BOOL isOutputStreamOpen;
+@property (assign, nonatomic) BOOL isDataSession;
+@property (nonatomic, strong, nullable) NSThread *ioStreamThread;
+@property (nonatomic, strong, nullable) SDLMutableDataQueue *sendDataQueue;
+@property (nonatomic, strong, nullable) dispatch_semaphore_t canceledSemaphore;
```
davidswi: Done.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -43,19 +53,26 @@ - (instancetype)initWithAccessory:(EAAccessory *)accessory forProtocol:(NSString
 - (BOOL)start {
     __weak typeof(self) weakSelf = self;
 
-    NSString *logMessage = [NSString stringWithFormat:@"Opening EASession withAccessory:%@ forProtocol:%@", _accessory.name, _protocol];
+    NSString *logMessage = [NSString stringWithFormat:@"Opening EASession withAccessory:%@ forProtocol:%@", self.accessory.name, self.protocol];
     [SDLDebugTool logInfo:logMessage];
 
-    if ((self.easession = [[EASession alloc] initWithAccessory:_accessory forProtocol:_protocol])) {
+    if ((self.easession = [[EASession alloc] initWithAccessory:self.accessory forProtocol:self.protocol])) {
```
davidswi: Done

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -43,19 +53,26 @@ - (instancetype)initWithAccessory:(EAAccessory *)accessory forProtocol:(NSString
...
         __strong typeof(self) strongSelf = weakSelf;
 
         [SDLDebugTool logInfo:@"Created Session Object"];
 
         strongSelf.streamDelegate.streamErrorHandler = [self streamErroredHandler];
         strongSelf.streamDelegate.streamOpenHandler = [self streamOpenedHandler];
-
-        [strongSelf startStream:weakSelf.easession.outputStream];
-        [strongSelf startStream:weakSelf.easession.inputStream];
+        if (!self.isDataSession) {
```
davidswi: Done.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
 }
 
 - (void)stop {
-    [self stopStream:self.easession.outputStream];
-    [self stopStream:self.easession.inputStream];
+
```
davidswi: Done.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+    if (!self.isDataSession) {
+        [self stopStream:self.easession.outputStream];
+        [self stopStream:self.easession.inputStream];
+    } else {
+        [self.ioStreamThread cancel];
```
davidswi: Done.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+        
+        long lWait = dispatch_semaphore_wait(self.canceledSemaphore, dispatch_time(DISPATCH_TIME_NOW, streamThreadWaitSecs * NSEC_PER_SEC));
+        if (lWait == 0) {
+            [SDLDebugTool logInfo:@"Stream thread canceled"];
+        } else{
```
davidswi: Done.

joeljfischer:
@davidswi you don't have to comment on every requested change. When you push a commit that changes that line, the request will automatically be dismissed.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+        }
+        self.ioStreamThread = nil;
+        self.isDataSession = NO;
+    }
     self.easession = nil;
 }
 
+- (void)sendData:(NSData *)data {
+    // Enqueue the data for transmission on the IO thread
+    [self.sendDataQueue enqueueBuffer:data.mutableCopy];
```
davidswi: As in the original implementation, we want to consume the NSData bytes in chunks equal to the available space in the output stream. I don't know of a better way to do this unless the transport interface is changed to take a mutable NSData.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+}
+
+- (BOOL)sdl_dequeueAndWriteToOutputStream:(NSError **)error {
```
davidswi: Done.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+    NSMutableData *remainder = [self.sendDataQueue frontBuffer];
+    BOOL allDataWritten = NO;
+    
+    if (error != nil && remainder != nil && ostream.streamStatus == NSStreamStatusOpen) {
+        NSInteger bytesRemaining = remainder.length;
+        NSInteger bytesWritten = [ostream write:remainder.bytes maxLength:bytesRemaining];
+        if (bytesWritten < 0) {
+            *error = ostream.streamError;
+        } else if (bytesWritten == bytesRemaining) {
+                // Remove the data from the queue
```
davidswi: Fixed.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+            }
+    }
+
+    return allDataWritten;
+}
+
+- (void)sdl_handleOutputStreamWriteError:(NSError *)error {
+    NSString *errString = [NSString stringWithFormat:@"Output stream error: %@", error];
+    [SDLDebugTool logInfo:errString];
+    // REVIEW: We should look at the domain and the code as a tuple and decide
```
davidswi: Yes. Fixed.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+    // is closed, we will flush the send queue and leave it as-is otherwise
+    // so that temporary error conditions can be dealt with by retrying
+    if (self.easession == nil ||
+        self.easession.outputStream == nil ||
+        self.easession.outputStream.streamStatus != NSStreamStatusOpen) {
+        [self.sendDataQueue flush];
+    }
+}
+
+- (void)sdl_accessoryEventLoop {
```
davidswi: Done.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+    @autoreleasepool {
+        NSAssert(self.easession, @"_session must be assigned before calling");
+
+        if (!self.easession) {
+          return;
```
davidswi: Fixed.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+        do {
+            if (self.sendDataQueue.count > 0 && !self.sendDataQueue.frontDequeued) {
+                NSError *sendErr = nil;
+                if (![self sdl_dequeueAndWriteToOutputStream:&sendErr] && sendErr != nil) {
+                    [self sdl_handleOutputStreamWriteError:sendErr];
+                }
+            }
+            
+            [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode
+                                     beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.25f]];
```
davidswi: The principle here is to give the event loop enough time to process stream events while also allowing it to handle new enqueued data buffers in a timely manner. We're capping the run loop CPU time at 0.25s maximum before it will return control to the rest of the loop.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+        } while (![NSThread currentThread].cancelled);
+
+        NSLog(@"closing accessory session");
```
davidswi: Done.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+
+        // Close I/O streams of the iAP session
+        [self sdl_closeSession];
+        dispatch_semaphore_signal(self.canceledSemaphore);
+    }
+}
+
+// Must be called on accessoryEventLoop.
+- (void)sdl_closeSession {
+  if (self.easession) {
```
davidswi: Done. We need to be careful about NSAssert as it will crash in release app code. Is that what . we want?

davidswi: Yes, you are correct.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -141,10 +258,26 @@ - (SDLStreamErrorHandler)streamErroredHandler {
     };
 }
 
+- (SDLStreamHasSpaceHandler)streamHasSpaceHandler {
```
davidswi: Joel, I am happy to make the private methods I've added conform to the naming convention. I think it's out of the scope of this PR to rename them all. Do you agree?

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -66,17 +83,117 @@ - (BOOL)start {
...
+  }
+}
+
 
 #pragma mark - Private Stream Lifecycle Helpers
 
 - (void)startStream:(NSStream *)stream {
     stream.delegate = self.streamDelegate;
-    [stream scheduleInRunLoop:[NSRunLoop mainRunLoop] forMode:NSDefaultRunLoopMode];
+    [stream scheduleInRunLoop:[NSRunLoop currentRunLoop] forMode:NSDefaultRunLoopMode];
```
davidswi: Since this method is common to both the control and data sessions and the thread it is being called on depends on which session is being stopped, we would need to condition the assert on the session type. Is that what we want to do?

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -141,10 +258,26 @@ - (SDLStreamErrorHandler)streamErroredHandler {
...
+                [strongSelf sdl_handleOutputStreamWriteError:sendErr];
+            }
+        }
+    };
+}
 
 #pragma mark - Lifecycle Destruction
 
 - (void)dealloc {
+    self.sendDataQueue = nil;
```
davidswi: Has this been verified in develop with zombies on Instruments run? Many of the connection transition issues we have seen are due to the External Accessory Framework objects holding zombie references to partially deallocated objects. I am all for removing unnecessary code that ARC can generate itself, but it appears in the complex chain of events between the XPC communication from the app to the IAP daemon that object references are not cleaned up during transitions, leading to zombies when a new connection to the head unit is established while the old one is partially torn down.

SmartDeviceLink/SDLIAPTransport.m
davidswi: Yes, that was a bug in the merge I did from our branch. I have reinstated.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -88,19 +89,25 @@ - (void)sdl_stopEventListening {
...
 - (void)sdl_accessoryDisconnected:(NSNotification *)notification {
     [SDLDebugTool logInfo:@"Accessory Disconnected Event" withType:SDLDebugType_Transport_iAP toOutput:SDLDebugOutput_All toGroup:self.debugConsoleGroupName];
+    [self sdl_stopEventListening];
 
     // Only check for the data session, the control session is handled separately
     EAAccessory *accessory = [notification.userInfo objectForKey:EAAccessoryKey];
-    if (accessory.connectionID == self.session.accessory.connectionID) {
+    if (accessory.connectionID != self.session.accessory.connectionID) {
+         [SDLDebugTool logInfo:@"Accessory connection ID mismatch!!!" withType:SDLDebugType_Transport_iAP toOutput:SDLDebugOutput_All toGroup:self.debugConsoleGroupName];
+    }
```
davidswi: This change was made because, due to zombie object references, the EAAccessory framework was sending disconnect notifications with a different connection ID to the transport. If the disconnect sequence is not executed, these objects hang around and cause crashes shortly thereafter. I recommend based on our testing that we disconnect even if the connection ID doesn't match.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -88,19 +89,25 @@ - (void)sdl_stopEventListening {
...
+    if ([accessory.name isEqualToString:self.session.accessory.name]){
```
davidswi: The idea is to guard against another accessory disconnect, say for an MFi BT headset, causing a false positive disconnect in the SDL app. I'm happy to base this on serial number, but I do strongly recommend that if we think our accessory has gone away that we clean up regardless of the connection ID.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -117,9 +124,13 @@ - (void)sdl_applicationWillEnterForeground:(NSNotification *)notification {
 #pragma mark - Stream Lifecycle
 
 - (void)connect {
+    [self sdl_connect:nil];
+}
+
+- (void)sdl_connect:(EAAccessory *)accessory {
```
davidswi: That's not correct. We call sdl_connect: with the accessory provided in sdl_accessoryConnected.

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -140,22 +151,40 @@ - (void)disconnect {
 
 #pragma mark - Creating Session Streams
 
-- (void)sdl_establishSession {
+- (BOOL)sdl_tryConnectAccessory:(EAAccessory *)accessory {
```
davidswi: I believe this is consistent with the Apple style guidelines for APIs that attempt to perform an operation and return NO if failure occurs. An example is tryLock in NSLock.

SmartDeviceLink/SDLIAPTransport.m
davidswi: There is a race condition being guarded against here that was found during debugging. Since the handler is invoked on the receive dispatch queue, a disconnect can occur *underneath* the queue dispatching the block causing the data to be sent to the transport when it is in the process of being torn down.

SmartDeviceLink/SDLIAPTransport.m
davidswi: Yes, the same race condition can now occur with the data session because the dataStreamHasBytesHandler is executed on the IO thread and the accessory disconnect notification arrives on the main thread, causing data to be passed to the delegate while the main thread is tearing down the transport.

SmartDeviceLink/SDLMutableDataQueue.m
davidswi: Yes, enqueuing occurs on the main thread, dequeuing occurs on the IO thread.

SmartDeviceLink/SDLMutableDataQueue.m
davidswi: What if there are elements present in the queue when the owner object is torn down? I believe this will leak.

SmartDeviceLink/SDLMutableDataQueue.h
davidswi: It specifically handles NSMutableData because it feeds an NSStream with an unknown amount of space available for writing. The data at the head of the queue needs to be consumed as space is available without changing its position in the queue.

davidswi: This class is very specifically used in conjunction with the IAP data session. We could decide to move it out of utilities and make it a helper class for the IAP transport.

davidswi: I have no strong feelings either way.

SmartDeviceLink/SDLMutableDataQueue.h
davidswi: I suggest removeAllObjects, consistent with the standard NS containers.

SmartDeviceLink/SDLMutableDataQueue.m
joeljfischer: No, ARC will clean those up automatically if the reference count goes to 0.

davidswi: OK, I will remove this dealloc and I recommend that we do Instruments test runs with IAP as the transport to prove that we are not leaking objects in the transport hierarchy. Let's agree on a strategy for and ownership of long haul connect/disconnect testing over IAP to ensure we have improved stability (MTTF might be the best metric) and are not leaking.

SmartDeviceLink/SDLIAPSession.m
joeljfischer: We shouldn't be modifying this much, but I do believe it's good form to have an assert of that type if we are requiring a method to be called on a certain thread. We could do a macro here to make sure that assert is clear.

SmartDeviceLink/SDLIAPSession.m
joeljfischer: Everything I've seen from Apple says that `dealloc` post-ARC is for disposing of non instance variable resources. ivars are certainly torn down after `dealloc`, but I don't know that setting these to nil would necessarily help, because the object itself would only be partially torn down, correct? I'm just not sure this helps. I haven't seen these connection transition issues at all, so if you have verifiable, repeatable methods for testing an EA XPC issue, it would be nice if you could check that (assuming it's not a long process) or to open those steps up.

SmartDeviceLink/SDLIAPTransport.m
joeljfischer: So, what if the app is connected to a different EA Accessory (say we're connected over USB) and they're connected over Bluetooth to something else. Wouldn't that cause problems?

SmartDeviceLink/SDLIAPTransport.m
joeljfischer: Ah, I see it now. Could this be named `sdl_connectAccessory:`. It would be nice to have some API documentation for this method explaining what it does and what happens when `nil` is passed.

SmartDeviceLink/SDLIAPTransport.m
joeljfischer: I believe that's actually the exception rather than the rule. For example, in the `NSFileManager` APIs, the method `createFileAtPath:contents:attributes` returns a BOOL for success, as does `createDirectoryAtURL:blahblah:`, and many others. `tryLock` is the only one I've ever heard of that uses `try`.

SmartDeviceLink/SDLMutableDataQueue.m
joeljfischer: Can we add a comment to that effect?

SmartDeviceLink/SDLIAPTransport.m
joeljfischer: Thanks, can we add a comment explaining that?

SmartDeviceLink/SDLMutableDataQueue.h
```diff
@@ -0,0 +1,60 @@
...
+
+#import <Foundation/Foundation.h>
+
+/**
+ This queue class is used by the IAP data session to provide safe, async writing of RPC messages to the connected SDL compliant accessory. It uses NSMutableData objects because the NSOutputStream has a variable amount of space available based on iOS and accessory buffering. We want to be able to pop the buffer at the head multiiple times and consume as many bytes as there are available in the stream.
+ */
+
+@interface SDLMutableDataQueue : NSObject
+
+NS_ASSUME_NONNULL_BEGIN
```
joeljfischer: I know this is nitpicky, but for consistency's sake, this should be before the `@interface` and should end after the `@end`. Again, I know that's über-nitpicky.

SmartDeviceLink/SDLIAPTransport.m
davidswi: We are still verifying the accessory we are disconnecting from is the SDL accessory by virtue of its serial number. I believe Apple enforces non-NULL serial numbers for all shipping accessories, so the likelihood that another connected accessory being used by the app has the same serial number is very remote.

SmartDeviceLink/SDLIAPTransport.m
davidswi: Fair enough.

SmartDeviceLink/SDLIAPSession.m
davidswi: OK, I agree and have added an NSAssert that checks for this condition in the startStream and stopStream methods.

SmartDeviceLink/SDLIAPSession.m
davidswi: I think the best way to proceed is as follows:
1. I will remove the deallocs from the PR

2. Let's discuss on Slack a process for ensuring we can reproduce the failures we are seeing with connection and disconnection transition cycles. When I joined the project, we had a number of open crash tickets in our JIRA due to disconnection while SDL communication was in flight, followed by reconnection shortly thereafter. I think it's best for us to roll up the ticket repro steps and provide them to you.

3. Let's agree on a path forward based on empirical metrics such as MTTF or mean cycles to failure that allows all parties to track robustness as connections come and go.

SmartDeviceLink/SDLIAPTransport.m
davidswi: I added some comments to sdl_connectAccessory describing how we handle non-nil and nil accessory parameters.

SmartDeviceLink/SDLIAPSession.m
```diff
@@ -210,6 +211,8 @@ - (void)stopStream:(NSStream *)stream {
 
     // When you disconect the cable you get a stream end event and come here but stream is already in closed state.
     // Still need to remove from run loop.
+    
+    NSAssert((self.isDataSession && [[NSThread currentThread] isEqual:self.ioStreamThread]) || [NSThread isMainThread], @"startStream is being called on the wrong thread!!!");
```
joeljfischer: This is in `stopStream`

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -172,6 +172,9 @@ - (void)sdl_establishSessionWithAccessory:(EAAccessory *)accessory {
         // We should be attempting to connect
         self.retryCounter++;
         EAAccessory *sdlAccessory = accessory;
+        // If we are being called from sdl_connectAccessory, the EAAccessoryDidConnectNotification
+        // will contain the SDL accessory to connect to and we can connect without searching the
+        // accessory manager's connected accessory list. Otherwise, we fall through to a search.
```
joeljfischer: This can be one big comment instead of broken up (like you do below L420)

SmartDeviceLink/SDLMutableDataQueue.m
```diff
@@ -31,6 +31,9 @@ - (instancetype)init {
 }
 
 - (void)enqueueBuffer:(NSMutableData *)data {
+    // Since this method is being called from the main thread and the dequeue methods
+    // are being called from the data session stream thread, we need to put critical
+    // sections around the queue members. Use the @synchronized object level lock to do this.
```
joeljfischer: This can be one big comment

SmartDeviceLink/SDLIAPTransport.m
joeljfischer: 👍

SmartDeviceLink/SDLIAPTransport.m
joeljfischer: 👍

SmartDeviceLink/SDLIAPSession.m
joeljfischer: 👍

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -140,23 +152,45 @@ - (void)disconnect {
...
     [SDLDebugTool logInfo:@"Attempting To Connect"];
     if (self.retryCounter < createSessionRetries) {
         // We should be attempting to connect
         self.retryCounter++;
-        EAAccessory *accessory = nil;
-
+        EAAccessory *sdlAccessory = accessory;
+        // If we are being called from sdl_connectAccessory, the EAAccessoryDidConnectNotification
+        // will contain the SDL accessory to connect to and we can connect without searching the
+        // accessory manager's connected accessory list. Otherwise, we fall through to a search.
```
joeljfischer: This should be one big comment

SmartDeviceLink/SDLMutableDataQueue.m
joeljfischer: 👍

joeljfischer:
This merge was reverted due to https://github.com/smartdevicelink/sdl_ios/pull/592#issuecomment-302734996

Please create a new PR with fixes.

joeljfischer:
@davidswi

joeljfischer:
@davidswi Now I'm seeing this issue even without this PR on the master branch. The revert will be reverted again. This issue will have to be fixed before release on June 31.

davidswi:
Not surprising I'll prepare a fix.

PR #608, opened 2017-04-19
Fixed clearing of upload queue

### LINKED ISSUE #606: SDLManager becomes unusable when disconnecting during file upload
### Bug Report
When disconnecting the device during a file upload, the SDLManager gets into a state bad state, in that the ‘ready’ block is never invoked on subsequent connections.
 
It appears the root issue is that the SDLFileManager’s serial operation queue jams if uploads are active during disconnect, and so will not perform subsequent upload operations.

##### Reproduction Steps
1. Connect the device and launch the app.
2. Perform some action the initiates a file upload
3. Pull the USB cable while upload is active
4. Attempt to reconnect the device.  Observe no app icon.  As SDLManger instance lifecycle extends between HMI connections, a mobile app restart is required to recover.

##### Expected Behavior
SDLManager should remain usable after disconnect.

##### Observed Behavior
SDLManager does not complete setup process on reconnect, and so is unusable.

* iOS Version: 10.2
* SDL iOS Version: 4.5.5
* Testing Against: Sync 3 v2 TDK

After some digging, it appears that there were two things contributing to the problem:
·         SDLFileManager defines a method `willEnterStateShutdown`; SDLStateMachine only calls: `didEnter [STATE]:` on its targets.  So cancellation methods are never invoked.
·         In `willEnterStateShutdown`, SDLFileManager attempts to cancel all active uploads using `[NSOperationQueue cancelAllOperations]`.  This method does not clear the queue, just notifies each task it is canceled – tasks must still signal completion.  `SDLUploadFileOperation`‘s cancelation logic for individual tasks relies on responseHandler callbacks from the connectionManger.  As the SDLResponseDispatcher’s response map is cleared on shutdown, these callbacks are never invoked, and the queue is never cleared.

tpulatha:
Fixes #606 

### Risk

### Testing Plan
1. Connect the device and launch the app.
2. Perform some action the initiates a file upload
3. Pull the USB cable while upload is active
4. Attempt to reconnect the device. Observe no app icon.

### Summary
* Renamed `- (void)willEnterStateShutdown` to `- (void)didEnterStateShutdown` as the `SDLStateMachine` only calls `didEnter...` functions
* Added the following code to `SDLResponseDispatcher.m` to ensure that ResponseHandlers are cleared.
```objc
for (SDLRPCCorrelationId *key in self.rpcResponseHandlerMap) {
            SDLResponseHandler f = self.rpcResponseHandlerMap[key];
            f(self.rpcRequestDictionary[key], nil, [NSError sdl_lifecycle_notConnectedError]);
        }
```

##### Breaking Changes
* No breaking changes

##### Enhancements
* Fixed a bug

##### Bug Fixes
* Fixed clearing upload queues

### DIFF
--- a/SmartDeviceLink/SDLResponseDispatcher.m
+++ b/SmartDeviceLink/SDLResponseDispatcher.m
@@ -108,6 +108,10 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (void)clear {
+    for (SDLRPCCorrelationId *key in self.rpcResponseHandlerMap) {
+            SDLResponseHandler f = self.rpcResponseHandlerMap[key];
+            f(self.rpcRequestDictionary[key], nil, [NSError sdl_lifecycle_notConnectedError]);
+        }
     [self.rpcRequestDictionary removeAllObjects];
     [self.rpcResponseHandlerMap removeAllObjects];
     [self.commandHandlerMap removeAllObjects];
--- a/SmartDeviceLink/SDLResponseDispatcher.m
+++ b/SmartDeviceLink/SDLResponseDispatcher.m
@@ -108,10 +108,11 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (void)clear {
-    for (SDLRPCCorrelationId *key in self.rpcResponseHandlerMap) {
-            SDLResponseHandler f = self.rpcResponseHandlerMap[key];
-            f(self.rpcRequestDictionary[key], nil, [NSError sdl_lifecycle_notConnectedError]);
-        }
+    // When we get disconnected we have to delete all existing responseHandlers as they are not valid anymore
+    for (SDLRPCCorrelationId *correlationID in self.rpcResponseHandlerMap) {
+        SDLResponseHandler responseHandler = self.rpcResponseHandlerMap[correlationID];
+        responseHandler(self.rpcRequestDictionary[correlationID], nil, [NSError sdl_lifecycle_notConnectedError]);
+    }
     [self.rpcRequestDictionary removeAllObjects];
     [self.rpcResponseHandlerMap removeAllObjects];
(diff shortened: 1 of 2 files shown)


SmartDeviceLink/SDLResponseDispatcher.m
```diff
@@ -108,6 +108,10 @@ - (void)storeRequest:(SDLRPCRequest *)request handler:(nullable SDLResponseHandl
 }
 
 - (void)clear {
+    for (SDLRPCCorrelationId *key in self.rpcResponseHandlerMap) {
+            SDLResponseHandler f = self.rpcResponseHandlerMap[key];
+            f(self.rpcRequestDictionary[key], nil, [NSError sdl_lifecycle_notConnectedError]);
+        }
```
joeljfischer: Misaligned

SmartDeviceLink/SDLResponseDispatcher.m
```diff
@@ -108,6 +108,10 @@ - (void)storeRequest:(SDLRPCRequest *)request handler:(nullable SDLResponseHandl
 }
 
 - (void)clear {
+    for (SDLRPCCorrelationId *key in self.rpcResponseHandlerMap) {
```
joeljfischer: Please add a comment explaining that we need to call all of the completion handlers and give a reason.

SmartDeviceLink/SDLResponseDispatcher.m
```diff
@@ -108,6 +108,10 @@ - (void)storeRequest:(SDLRPCRequest *)request handler:(nullable SDLResponseHandl
...
+            SDLResponseHandler f = self.rpcResponseHandlerMap[key];
```
joeljfischer: Do not use 1 letter variable names, a better name would be `responseHandler`.

PR #610, opened 2017-04-25
Fixed providing error info if video session is not started

### LINKED ISSUE #609: If starting video session fails, error is always nil
### The problem
When [SDLStreamingMediaManager startVideoSessionWithTLS:startBlock:] fails, error is never provided in the result block.

##### Suggestion:
I expect the problem is in sdl_configureVideoEncoderWithError method of SDLStreamingMediaManager.m (lines 419, 433, 442)

it’s

> if (*error != nil)

but I think it should be

> if (error != nil)

alexanderlavrushko:
Fixes #609 

### Risk

### Testing Plan
Probably the only way to simulate a fail of VTCompressionSessionCreate (and other VideoToolbox functions) is to feed incorrect video session attributes to VideoToolbox, then we can expect a non-empty error in start video result block.

### Summary
Fixed the incorrect check of the error pointer - actually the code was always thinking that "there is no need to provide an error".

### DIFF
--- a/SmartDeviceLink/SDLStreamingMediaManager.m
+++ b/SmartDeviceLink/SDLStreamingMediaManager.m
@@ -416,7 +416,7 @@ void sdl_videoEncoderOutputCallback(void * CM_NULLABLE outputCallbackRefCon, voi
 
     if (status != noErr) {
         // TODO: Log the error
-        if (*error != nil) {
+        if (error != nil) {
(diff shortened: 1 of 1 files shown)


SmartDeviceLink/SDLStreamingMediaManager.m
```diff
@@ -416,7 +416,7 @@ - (BOOL)sdl_configureVideoEncoderWithError:(NSError *__autoreleasing *)error {
 
     if (status != noErr) {
         // TODO: Log the error
-        if (*error != nil) {
+        if (error != nil) {
```
joeljfischer: These should be `if (error != NULL)` because we are comparing a reference pointer, not an Objective-C object.

PR #615, opened 2017-05-09
Guarantee that HMI status is set for ready handler success

### LINKED ISSUE #540: SDLManager's `startWithReadyHandler` should return back after onHMIStatus has been received
### Bug Report
Currently, `SDLLifecycleManager` returns back from the `startWithReadyHandler` once we receive a register app interface response, all the managers are set up, and the app icon is set. We then fire a delegate callback for `hmiLevel:didChangeToLevel:`. With this delegate callback, we are _assuming_ that we will have an HMI State by this point. We should instead return back from the `startWithReadyHandler` only after the register app interface response in received, all the managers are set up, the app icon is set *and* the `OnHMIStatus` is received.

* SDL iOS Version: 4.3+

brettywhite:
Fixes #540

### Risk

### Testing Plan
Testing first was done manually ensuring that if for some reason the HMI status was nil, that the state would indeed transition to ready as soon as we got notification that the HMI status changed. 

Additionally, a new unit test was added ensuring that during the state `SDLLifecycleStateSettingUpHMI` that the `readyHandlerSuccess` is still set to no. A second test was added to make sure the state is changed to ready on a hmi status notification 

### Summary
First, `SDLLifecycleStatePostManagerProcessing` was renamed to `SDLLifecycleStateSettingUpAppIcon` to more appropriately let the developer know whats actually happening during that state. Second, a new state `SDLLifecycleStateSettingUpHMI` was added to go between the states of `SDLLifecycleStateSettingUpAppIcon` and `SDLLifecycleStateReady`. Its purpose it to catch rare cases where the HMI status has not been received at that point, and to not allow the `readyHandler` to be set to `YES` until we can confirm that the HMI has been setup.

##### Bug Fixes
* This fix allows us to insure that the HMI is ready when the developer receives the `YES` response from the `readyHandler`. Before, it was not guaranteed.

### DIFF
--- a/SmartDeviceLink/SDLLifecycleManager.m
+++ b/SmartDeviceLink/SDLLifecycleManager.m
@@ -285,7 +285,6 @@ SDLLifecycleState *const SDLLifecycleStateReady = @"Ready";
     }
     // We are sure to have a HMIStatus, set state to ready
     [self.lifecycleStateMachine transitionToState:SDLLifecycleStateReady];
-    
 }
 
 - (void)didEnterStateReady {
--- a/SmartDeviceLink/SDLLifecycleManager.m
+++ b/SmartDeviceLink/SDLLifecycleManager.m
@@ -49,7 +49,8 @@ SDLLifecycleState *const SDLLifecycleStateReconnecting = @"Reconnecting";
 SDLLifecycleState *const SDLLifecycleStateConnected = @"Connected";
 SDLLifecycleState *const SDLLifecycleStateRegistered = @"Registered";
 SDLLifecycleState *const SDLLifecycleStateSettingUpManagers = @"SettingUpManagers";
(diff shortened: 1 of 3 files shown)


SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -263,18 +265,29 @@ - (void)didEnterStateSettingUpManagers {
...
 
+- (void)didEnterStateSettingUpHMI{
+    // We want to make sure we've gotten a SDLOnHMIStatus notification
+    if (self.hmiLevel == nil) {
+        // If nil, return and wait until we get a notification
+        return;
+    }
+    // We are sure to have a HMIStatus, set state to ready
+    [self.lifecycleStateMachine transitionToState:SDLLifecycleStateReady];
+
```
joeljfischer: You can get rid of this extra line

PR #622, opened 2017-05-31
Add Swift example app and cleanup existing sample app

### LINKED ISSUE #620: Move Hello_SDL to this repository
### Bug Report
Instead of having separate repositories, we should move Hello_SDL's various branches (swift / obj-c examples) to this repository. We should update our Obj-C sample app to contain all the features of Hello_SDL and should create a Swift example with all the features of the Hello_SDL swift example. Then, we may delete the old repository.

We should do this because it allows us to test building the library into example apps (which lets us test for breakage in APIs, especially Swift), and it allows us to keep the example apps updated with the latest version of the library automatically; that way we do not have to update a separate repository. When a new release of the library comes out, the example apps on that commit automatically support the latest release.

This change will not be tied to any specific release, and it is merely moving and improving the existing apps.

brettywhite:
Fixes #620 

### Risk

### Testing Plan
These changes do not change the library or its tests. They are separate targets in the project.

### Summary
- Adds a new target, SmartDeviceLink-ExampleSwift
- Edits an existing target, SmartDeviceLink-Example

##### Breaking Changes
* No Breaking Changes to Library

##### Enhancements
* Adds in an example for Swift
    * Includes Vehicle Data RPC subscription / logging like the HelloSDL App
    * Includes Swiftlint
    * Includes all features of Obj-C Example App
* Cleans up some files in the existing Example App
    * Added Vehicle Data RPC subscription / logging like the HelloSDL App

##### Bug Fixes
* No Bug Fixes

### CLA


SmartDeviceLink-ExampleSwift/ConnectionContainerViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+//
+//  Copyright © 2017 smartdevicelink. All rights reserved.
+//
+
+import UIKit
+
+class ConnectionContainerViewController: UIViewController {
+
+    @IBOutlet weak var segmentedControl: UISegmentedControl!
+    var viewControllers: NSMutableArray = []
```
joeljfischer: This should use Swift's array, not `NSMutableArray`

SmartDeviceLink-ExampleSwift/ProxyManager.swift
```diff
@@ -0,0 +1,305 @@
...
+        NotificationCenter.default.addObserver(self, selector: #selector(didReceiveVehicleData(_:)), name: .SDLDidReceiveVehicleData, object: nil)
+    }
+
+    fileprivate func addPermissionManagerObservers() {
+        _ = sdlManager.permissionManager.addObserver(forRPCs: ["SubscribeVehicleData"], groupType: .allAllowed) { (_, _) in
+        }
+    }
+
+    fileprivate func prepareRemoteSystem(overwrite: Bool = false, completionHandler: @escaping (Void) -> (Void)) {
+
```
joeljfischer: There seems to be some empty lines at the top of a few functions and classes which shouldn't be there.

joeljfischer:
Re-target to the 5.0 library

PR #631, opened 2017-06-14
Fix: crash when using example app with iAP

### LINKED ISSUE #625: Crash issue on USB disconnection
### Bug Report
Using the SDL example app, on the 2nd USB disconnection, the app will crash.

##### Reproduction Steps
1. Connect via USB
2. Connect the app with the "Connect" button
3. Open the app on the HMI
4. Use the app a bit
5. Disconnect the USB from the phone (the app UI appropriately shows that it is no longer connected)
6. Plug back in the USB
7. Press the connect button again.

##### Expected Behavior
The app does not crash

##### Observed Behavior
The app crashes

* iOS Version: 10.3.2
* SDL iOS Version: 4.5.5 / 4.6.0
* Testing Against: Ford Sync 3 TDK (v. 2.2, Build 16208_DEVTEST)

[SDL Example  5-19-17, 11-14 AM.txt](https://github.com/smartdevicelink/sdl_ios/files/1014741/SDL.Example.5-19-17.11-14.AM.txt)
[SDL Example  5-19-17, 11-24 AM.txt](https://github.com/smartdevicelink/sdl_ios/files/1014740/SDL.Example.5-19-17.11-24.AM.txt)

brettywhite:
Fixes #625 

### Risk

### Testing Plan
This was tested extensively on 2 TDKs, on an older test version and on the newest version. I also tested this bug with previous versions of the SDL library and it was present back to 4.5.3, at which time i decided that was far back enough. 

### Summary
The issue was specifically in our example application that is baked into the library. The issue was on a forced USB disconnection and the subsequent reconnection, iAP was calling a reconnect event while  the `proxyManager`also tried to connect if the user pressed the connect button.

There were then 2 instances of the `sdlManager`. The fix was to properly check and then remove the reference when the `proxyManager` reset. 

It is also worth noting that this would almost never happen in a production application. Users are not given an option to force connect to SDL like in this instance. If we can think of a time where this could happen, it would either need more investigation or require added developer documentation to show how to remove references appropriately. 

##### Breaking Changes
*  No breaking changes
*  Edited `proxyManager` in the example app

### CLA

### DIFF
--- a/SmartDeviceLink_Example/Classes/ProxyManager.m
+++ b/SmartDeviceLink_Example/Classes/ProxyManager.m
@@ -71,7 +71,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)startIAP {
     [self sdlex_updateProxyState:ProxyStateSearchingForConnection];
     // Check for previous instance of sdlManager
-    if (!self.sdlManager){
+    if (self.sdlManager) { return; }
     SDLLifecycleConfiguration *lifecycleConfig = [self.class setLifecycleConfigurationPropertiesOnConfiguration:[SDLLifecycleConfiguration defaultConfigurationWithAppName:SDLAppName appId:SDLAppId]];
     
     // Assume this is production and disable logging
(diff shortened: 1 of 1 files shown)


SmartDeviceLink_Example/Classes/ProxyManager.m
```diff
@@ -70,6 +70,8 @@ - (instancetype)init {
 
 - (void)startIAP {
     [self sdlex_updateProxyState:ProxyStateSearchingForConnection];
+    // Check for previous instance of sdlManager
+    if (!self.sdlManager){
```
NicoleYarroch: I would probably do 
```
if (self.sdlManager) { return }
// Then you don't have to indent all the code
```

Issue #637, opened 2017-06-29
stopStream NSAssert crashes on head unit shutdown


brettywhite:
### Bug Report
An NSAssert statement is triggered during the `stopStream` method during shutdown of SDL enabled head unit. This is a problem on the newest 4.6 version.

##### Reproduction Steps
1. connect sdl enabled app to head unit with SDL 4.6, put into foreground
2. power down the tdk (1. key off 2. key out 3. door open)
3. observe crash in app as tdk powers off

##### Expected Behavior
The application does not crash

##### Observed Behavior
The application crashes

##### OS & Version Information
* iOS Version: 10.3.2
* SDL iOS Version: 4.6.0-rc.2
* Testing Against: Ford TDK Build 17144_DEVTEST

##### Link to Code
[See code here](https://github.com/smartdevicelink/sdl_ios/blob/master/SmartDeviceLink/SDLIAPSession.m#L215)

PR #647, opened 2017-07-18
Fix deadlock on control session disconnection

### LINKED ISSUE #643: SDL iOS Proxy deadlocks when attempting to stop off the main thread
### Bug Report
When the IO stream thread was added to handle reads and writes outside the main run loop without the iAP transmit dispatch queue workaround, the `SDLIAPSession:stop:` method was modified to check for the data session, cancel the stream thread and wait on a `dispatch_semaphore`. This code assumed stop would *always* be called from the main thread. However, when there is a `NSStreamEventEnd` or `NSStreamEventError` event sent to the `SDLStreamDelegate:stream:handleEvent` method, the `stop` method is called before the `SDLIAPTransport:retry_establishSession:` is called, and the events are signaled on the stream thread, causing the stream thread to be canceled and the `dispatch_semaphore_wait()` call being made on the stream thread. The wait call fails because it is on the same thread as the signal call. This will deadlock for 1s causing the stream thread cancel to fail. 

The OS will also post the `EAAccessoryDidDisconnect` notification on the main thread. This will always happen, but it is not deterministic as to whether the notification will occur before or after the stream end event is triggered. The data session should only be cleaned up once and the safest way to do that is from the `EAAccessoryDidDisconnect` notification handler. 

We could dispatch the session stop call to the main thread from the `SDLStreamDelegate` handler, but this does not address the design principle of only one data session cleanup. I believe it is safe to assume the stream will only end on an accessory disconnect. From the Apple stream documentation at https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/NetworkingTopics/Articles/UsingSocketsandSocketStreams.html

If the other end of the connection closes the connection:

Your connection `delegate`’s `stream:handleEvent:` method is called with `streamEvent` set to `NSStreamEventHasBytesAvailable`. When you read from that stream, you get a length of zero (0).
Your connection delegate’s stream:handleEvent: method is called with `streamEvent` set to `NSStreamEventEndEncountered`.
When either of these two events occurs, the delegate method is responsible for detecting the end-of-file condition and cleaning up.

For the stream error handler, we do need to stop the session. I believe the correct approach here is to check in the `SDLIAPSession:stop:` method whether we are being called on the stream thread and if so, we can directly close the streams and end our thread.

##### Reproduction Steps
1. Launch SDL app and connect to SDL compatible accessory
2. Disconnect during SDL communication. Reconnect within a few seconds.
3. Repeat 2 until communication fails.

##### Expected Behavior
Communication should not fail.

##### Observed Behavior
The iOS console log shows Error: failed to cancel stream thread, followed by a crash due to the stream delegate calling a dealloc'ed object or a crash in the iap2d process.

* iOS Version: Reproduces on 8.x, 9.x and 10.x. Most easily reproduced on 10.x due to automatic app connection to compatible accessories.
* SDL iOS Version: Master, 4.6RC2
* Testing Against: Internal head unit engineering samples.

davidswi:
Fixes #643, #637 

This PR is ready for review.

### Risk

### Testing Plan
This PR has been tested on multiple head units by doing repeated disconnect/reconnect USB and BT cycles.

### Summary
When the iOS device is disconnected from the head unit, iOS generates stream end events for the input and output streams for the active `EASession`, along with the `EAAccessoryDidDisconnect` notification. Which order these events arrive in is not deterministic, and if the stream events arrive before the `EAAccessoryDidDisconnect` notification, they will occur on the stream run loop. For the data session, this means the `ioStreamThread` run loop and since `[session stop]` is called, it will wait on the session `canceledSemaphore`, which will deadlock for 1s because the stop method assumes it is being called from a thread other than the stream run loop thread. After the semaphore wait times out, the cleanup code in the stop method will wipe out the references to the data session and it will not close its streams before being deallocated. The `SDLStreamDelegate` instance will also go away, but since the streams are still open and scheduled they may generate additional events that will cause the `NSStream:streamEventTrigger:` method to crash with a call to the deallocated delegate.

This is not an issue for the control session because it is scheduled on the main thread. In this PR, I am leaving the control session event handlers alone. For the data session handlers, both now call `[session stop]` on the main queue synchronously before the session delegate and session are nil'ed.

##### Bug Fixes
* Fixes rare cases when attempting to stop the IAP transport can lead to a crash (#643, #637?)

### CLA

### DIFF
--- a/SmartDeviceLink/SDLIAPSession.m
+++ b/SmartDeviceLink/SDLIAPSession.m
@@ -84,6 +84,9 @@ NSTimeInterval const streamThreadWaitSecs = 1.0;
 }
 
 - (void)stop {
+    // This method must be called on the main thread
+    NSAssert([NSThread isMainThread], @"stop must be called on the main thread");
+    
     if (self.isDataSession) {
         [self.ioStreamThread cancel];
 
--- a/SmartDeviceLink/SDLIAPTransport.m
+++ b/SmartDeviceLink/SDLIAPTransport.m
@@ -389,12 +389,10 @@ int const streamOpenTimeoutSeconds = 2;
 #pragma mark Data Stream
 
 - (SDLStreamEndHandler)sdl_dataStreamEndedHandler {
-    
     __weak typeof(self) weakSelf = self;
 
     return ^(NSStream *stream) {
         __strong typeof(weakSelf) strongSelf = weakSelf;
-        
         
         [SDLDebugTool logInfo:@"Data Stream Event End"];
         if (strongSelf.session != nil) {
--- a/SmartDeviceLink/SDLIAPTransport.m
+++ b/SmartDeviceLink/SDLIAPTransport.m
@@ -395,14 +395,10 @@ int const streamOpenTimeoutSeconds = 2;
         __strong typeof(weakSelf) strongSelf = weakSelf;
 
         [SDLDebugTool logInfo:@"Data Stream Event End"];
-        [strongSelf.session stop];
-        strongSelf.session.streamDelegate = nil;
-
-        if (![legacyProtocolString isEqualToString:strongSelf.session.protocol]) {
-            [strongSelf sdl_retryEstablishSession];
-        }
-
-        strongSelf.session = nil;
+        // Per the Apple documentation, stream end is received when the accessory disconnects
+        // As we have observed, sometimes the EAAccessoryDidDisconnect notification is received before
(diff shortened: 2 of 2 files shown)


joeljfischer:
@davidswi In order to get fuller testing on this and #649, I'd like to bump them to the August 31 release v4.7.0. That release is currently housed on the develop branch. I shifted the base branch via the github interface; could you please do any re-testing you need to do against that branch?

brettywhite:
This does appear to fix #637 🎉

PR #649, opened 2017-07-18
Fix IAP control session crash on proxy dispose

### LINKED ISSUE #648: On proxy dispose, control session can be left running leading to crash
### Bug Report
When the Lifecycle Manager disposes the proxy directly with the IAP transport connected, the disconnect method will test for and stop the data session, but if there is a control session open waiting for the data session index byte, it will not be stopped, even though the stream delegate and all other objects associated with and including the transport, will be deallocated. If the accessory sends the control session byte after deallocation is complete, the EASession input stream will trigger a has bytes available event to the deallocated delegate and a crash will occur. 

##### Reproduction Steps
1. Connect SDL app to HU
2. Repeatedly connect/disconnect while app is going through initialization process

##### Expected Behavior
The app will cycle through initialization repeatedly

##### Observed Behavior
The app crashes eventually

* iOS Version: 8.x, 9.x, 10.x
* SDL iOS Version: 4.6RC2
* Testing Against: Prototype engineering head unit samples

Reproducible with the SDL example app if it is modified to always connect over iAP.

davidswi:
Fixes #648 

This PR is ready for review.

### Risk

### Testing Plan
This fix has been tested with multiple head units through disconnect/reconnect cycles.

### Summary
When the app calls the LifecycleManager stop API, the SDLProxy will be disposed and it will call `[transport disconnect]` as part of the dispose process.

##### Bug Fixes
* Fix IAP control session being left open in rare cases, leading to a crash (#648)

### CLA

### DIFF
--- a/SmartDeviceLink/SDLIAPTransport.m
+++ b/SmartDeviceLink/SDLIAPTransport.m
@@ -141,10 +141,16 @@ int const streamOpenTimeoutSeconds = 2;
     // Stop event listening here so that even if the transport is disconnected by the proxy
     // we unregister for accessory local notifications
     [self sdl_stopEventListening];
-    // Only disconnect the data session, the control session does not stay open and is handled separately
-    if (self.session != nil) {
-        [self.session stop];
-        self.session = nil;
+    if (self.controlSession) {
+        [self.controlSession.streamDelegate clearHandlers];
+        self.controlSession.streamDelegate = nil;
+        [self.controlSession stop];
+    } else {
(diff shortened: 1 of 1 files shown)


joeljfischer:
@davidswi In order to get fuller testing on this and #647, I'd like to bump them to the August 31 release v4.7.0. That release is currently housed on the develop branch. I shifted the base branch via the github interface; could you please do any re-testing you need to do against that branch?

PR #747, opened 2017-09-20
Fix Tests in Xcode 9

### LINKED ISSUE #742: Only disable LifecycleManagerSpec tests that crash due to libs
### Bug Report
Currently we disable the entire `SDLLifecycleManagerSpec` tests, but it's possible to only disable the few tests that are broken. When this is done, a few tests are failing for legitimate reasons, we need to investigate.

##### Reproduction Steps
1. Change the top level `xdescribe` to `describe`
2. Change L95 `it` to `xit`
3. Comment out L188

##### Expected Behavior
Tests pass

##### Observed Behavior
Tests fail

joeljfischer:
Fixes #742 

### Risk

### Testing Plan
This fixes tests so... 🤷‍♂️ 

### Summary
Fixes a number of tests that are currently broken and re-enables most lifecycle manager tests

##### Tests
* Fixes a number of tests

### CLA


NicoleYarroch:
I think there should be a warning that `SDLLockScreenViewControllerSnapshotTests` are going to fail if tested with any simulator that is not a iPhone 7.

joeljfischer:
@NicoleYarroch I don't believe it's possible to detect which simulator you're running at runtime...

cromotron:
@joeljfischer @NicoleYarroch  Actually you can detect the simulator type.

getenv("SIMULATOR_MODEL_IDENTIFIER")
iPhone8,1 -> iPhone 6s
iPhone8,2 ->  iPhone 6s Plus
iPhone8,4 -> iPhone SE
iPhone9,1 ->  iPhone 7
iPhone9,2 -> iPhone 7 Plus

Here is a link to the entire list https://www.theiphonewiki.com/wiki/Models

joeljfischer:
@NicoleYarroch Short answer, no there is no good way to present a warning if they're not using the correct simulator. We can show a log, but that's not super helpful I don't think.

PR #752, opened 2017-09-21
WIP: Issue 749 app crashes when usb cord is pulled

### LINKED ISSUE #679: Remove all deprecated and unneeded code for v5.0
### Bug Report
For v5.0, we should remove all deprecated methods and classes (other than the `SDLProxy` and `SDLProtocol` layers).

### LINKED ISSUE #749: App crashes when USB cord is pulled immediately after accessory setup
### Bug Report
When app is connected to a SDL accessory via a iAP connection, the app crashes when USB cord is pulled right after the icon shows up on the screen, but before the app launches on the SDL HMI. This bug is an edge case is is difficult to reproduce. 

There are three issues that may be at play
1. Notifications are not being unsubscribed to properly in the `SDLIAPTransport` class. As a result, the app crashes
1. A `NSException` in the `SDLStateMachine` is force quitting the app because of invalid state transitions. The ones I have found so far are:
    * old state: **Reconnecting**, new state: **SettingUpHMI**
    * old state: **Ready**, new state: **Started**
    * old state: **Shutdown**, new state: **StartupError**
1. `NSStream`s in the `SDLIAPTransport` class are not being closed properly before dealloc is being called. 
    * This is because the `SDLStreamDelegate` is being set to `nil` before the input and output streams are are closed and released.
  
##### Reproduction Steps
1. Use a iAP connection to connect the SDL Example app to a TDK.
1. Launch the app on SDL HMI.
1. Wait until the app has launched and the app has completed setup on the SDL HMI.
1. The app should be in the foreground, pull the USB cord.
1.  The app should be in the foreground, reconnect the USB cord. As soon as the app-icon appears on the HMI, but before the app launches, pull the USB cord.
1. The app should crash. (It might take several tries for this to happen - repeat 4 and 5 as necessary)

##### Expected Behavior
The app should never crash

##### Observed Behavior
The app crashes

* iOS Version: iOS 10 and 11
* SDL iOS Version: 5.0.0 and 4.7
* Testing Against: The SDL_Example app and a TDK (SYNC 3.0)

NicoleYarroch:
Fixes #749 
Fixes #679 

This PR is **not ready** for review.

### Risk

### Testing Plan
None written yet

### Summary
1. Notifications were not being unsubscribed properly in the `SDLIAPTransport` class, causing the app to crash when the USB cord was pulled.
1. The lifecycle state machine was not taking into account that the SDL accessory could be disconnected after the app icon was setup but before the HMI was setup.
1. The file manager state machine was not taking into account that the accessory could be disconnected before the list files request was returned. 

### Tasks Remaining:
- [ ] Add test cases for modified lifecycle state machine transitions 
- [ ] Add test cases for modified file manager state machine transitions

### CLA


joeljfischer:
@NicoleYarroch The IAPTransport notifications were being unsubscribed correctly. You should not have to do each individually.

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -301,11 +336,19 @@ - (void)didEnterStateSettingUpHMI {
...
+    if ([self sdl_didAccessoryDisconnect]) {
+        return;
+    }
+
     // We are sure to have a HMIStatus, set state to ready
     [self.lifecycleStateMachine transitionToState:SDLLifecycleStateReady];
 }
 
 - (void)didEnterStateReady {
+    if ([self sdl_didAccessoryDisconnect]) {
```
joeljfischer: Are we certain having these everywhere is necessary? We are getting a transport disconnect notification (transportDidDisconnect) if the accessory disconnected.

joeljfischer:
This was closed due to targeting a deleted branch. If this still needs to exist, re-open against master or develop

PR #775, opened 2017-10-23
Video streaming does not work when app foregrounded

### LINKED ISSUE #774: Video streaming does not start on reconnect if app is foregrounded
### Bug Report
This is an edge case where video streaming does not start on reconnect if the app was backgrounded and then foregrounded before the iPhone is reconnected to Core. This is happening because the video setup in the `SDLStreamingMediaLifecycleManager` is occurring when the app is foregrounded but before it has connected to Core.

##### Reproduction Steps
1. Connect app to Core and start video
1. Put app in background by pressing the home button
1. Pull the USB cord from the phone
1. Foreground the app
1. Plug in the USB cord to the phone
1. Restart video if necessary

##### Expected Behavior
Video will start playing

##### Observed Behavior
Video will not start playing

* iOS Version: iOS 11
* SDL iOS Version: 5.0.0
* Testing Against: SDL_Video_Ex and a TDK (SYNC 3.0)

NicoleYarroch:
Fixes #774 

### Risk

### Testing Plan
Tested using smoke tests created for 5.0.0

### Summary
Fixed a bug where video streaming did not start on reconnect if the app was backgrounded and then foregrounded before the iPhone was reconnected to Core. This happened because the video setup RPCs in the `SDLStreamingMediaLifecycleManager` were being sent before the device was connected to Core. 

### CLA

### DIFF
--- a/SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
+++ b/SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
@@ -269,10 +269,6 @@ typedef void(^SDLVideoCapabilityResponseHandler)(SDLVideoStreamingCapability *_N
 // We should be waiting to start any OpenGL drawing until UIApplicationDidBecomeActive is called.
 - (void)didEnterStateAppActive {
     SDLLogD(@"App became active");
-    if (!self.protocol) { return; }
-
-    [self sdl_startVideoSession];
-    [self sdl_startAudioSession];
 }
 
 #pragma mark Video Streaming


joeljfischer:
@NicoleYarroch Won't removing this break the case of restarting the sessions if the app is connected to Core and the streaming app is backgrounded then foregrounded?

NicoleYarroch:
Yes. The problem is that we need a way of knowing if the phone is connected to Core. The `isStreamingSupported` is not reset when the cord is pulled, so when the app is foregrounded,  the `isStreamingSupported` check passes in the `sdl_startVideoSession` method of the `SDLStreamingMediaLifecycleManager`.

PR #799, opened 2017-11-28
CarWindow - Automatic Video Streaming

### LINKED ISSUE #794: [SDL 0111] SDLCarWindow
Proposal: [SDLCarWindow](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0111-ios-carwindow.md)

> This proposal seeks to provide an interface for video streaming developers by which they may set their phone UI window and have it automatically streamed.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/345

joeljfischer:
Fixes #794 

### Risk

### Testing Plan
Unit tests will be added.

### Summary
This PR adds a system for automatically video streaming whatever is on the screen. This is a less efficient method than using pure OpenGL, but able to stream anything.

##### Enhancements
* Add automatic video streaming via screenshotting.
* Update the Lock Screen to use a UIWindow on top of any other UI.
* `SDLStreamingMediaManager` now manages framerate more strictly using a `CADisplayLink` and uses this to sync touch manager and car window updates to the framerate.
* Update `SDLTouchManager` to sync callbacks with the framerate, this fixes many issues with drawing at inopportune times. There is a BOOL to retain old behavior.

### Tasks Remaining:
- [x] Integration to `SDLStreamingMediaManager`
- [x] Unit Tests
- [x] Documentation
- [x] Logs
- [x] Configurable screen updates

### CLA


SmartDeviceLink/SDLCarWindow.m
```diff
@@ -0,0 +1,257 @@
...
+        
+        UIGraphicsBeginImageContextWithOptions(bounds.size, YES, 1.0f);
+        [self.rootViewController.view drawViewHierarchyInRect:bounds afterScreenUpdates:YES];
+        UIImage *screenshot = UIGraphicsGetImageFromCurrentImageContext();
+        UIGraphicsEndImageContext();
+        
+        CGImageRef imageRef = screenshot.CGImage;
+        
+        // We use MD5 Hashes to determine if we are sending the same frame over and over. If so, we will only send 30.
+        NSString *currentMd5Hash = [self.class sdl_md5HashForImageRef:imageRef];
```
Toyota-BSalahat: I think this is a carry-over from the Ford prototype. Was there any testing to see if the md5 hashing limits the framerate at all?

SmartDeviceLink/SDLCarWindow.m
```diff
@@ -0,0 +1,257 @@
...
+        self.sameFrameCounter = 0;
+        
+        // And start up the displayLink
+        self.displayLink = [CADisplayLink displayLinkWithTarget:self selector:@selector(sdl_sendFrame:)];
+        if (SDL_SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"10")) {
+            self.displayLink.preferredFramesPerSecond = self.targetFramerate;
+        } else {
+            self.displayLink.frameInterval = 60/self.targetFramerate;
+        }
+        [self.displayLink addToRunLoop:[NSRunLoop currentRunLoop] forMode:NSRunLoopCommonModes];
```
Toyota-BSalahat: Would there be any benefit to putting the displaylink on its own thread, similar to how it is in the iap transport class?

SmartDeviceLink/SDLFocusableItemLocator.h
```diff
@@ -16,6 +16,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, assign) BOOL enableHapticDataRequests;
 
+/**
+ The projection view controller associated with the Haptic Manager
+ */
+@property (nonatomic, weak) UIViewController *viewController;
```
Toyota-BSalahat: Was there a benefit to moving from UIWindow to UIViewController?

SmartDeviceLink/SDLLockScreenPresenter.m
```diff
@@ -8,48 +8,101 @@
...
 @implementation SDLLockScreenPresenter
 
+- (instancetype)init {
+    self = [super init];
+    if (!self) { return nil; }
+
+    CGRect screenFrame = [[UIScreen mainScreen] bounds];
+    _lockWindow = [[UIWindow alloc] initWithFrame:screenFrame];
+    _screenshotViewController = [[SDLScreenshotViewController alloc] init];
+    _lockWindow.rootViewController = _screenshotViewController;
```
Toyota-BSalahat: Do I understand correctly that the lockscreen is now window-based, instead of trying to find the top-level VC and presenting over it (we were working on a proposal for this change)?

SmartDeviceLink/SDLScreenshotViewController.h
```diff
@@ -0,0 +1,22 @@
...
+//  SDLScreenShotViewController.h
+//
+//  Created by Muller, Alexander (A.) on 2/6/17.
+//  Copyright © 2017 Ford Motor Company. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+/**
+ This class interacts with `SDLLockScreenPresenter`. It loads a screenshot of the app window and presents it on itself in an image view. This view controller is then presented on the lock window, which then presents the lock view controller over it.
```
Toyota-BSalahat: I'm not sure I follow what the goal of this class is. Why does a screen capture of the app need to be behind the lock screen?

SmartDeviceLink/SDLCarWindow.m
joeljfischer: Yes, when the application is static on the screen (no changes at all, which, admittedly, is not very often) it cuts the CPU usage in half.

SmartDeviceLink/SDLCarWindow.m
joeljfischer: No, I don't believe so. The screenshotting must be done on the main thread, and the displaylink is simply calling the method synced with the framerate, so it would likely be detrimental to do so.

SmartDeviceLink/SDLFocusableItemLocator.h
joeljfischer: No, but because `CarWindow` needs a `UIViewController`, it's easier to have developers set just the VC than the `UIWindow` and the VC both.

SmartDeviceLink/SDLLockScreenPresenter.m
joeljfischer: That is correct, this was included in the CarWindow proposal.

SmartDeviceLink/SDLScreenshotViewController.h
joeljfischer: So, what happens is, the lock window has this VC as its root. This VC takes a screenshot of the app's window before the lock window is displayed. Then, when the lock window is displayed, it appears that nothing changed in the UI, when in fact a new window has been displayed over the app window. Then, the lock screen VC modally presents on top of this VC.

SmartDeviceLink/SDLCarWindow.m
Toyota-BSalahat: Understood. I was just curious if when the application is trying to stream at 30+ fps, if hashing each frame began to become a bottleneck.

SmartDeviceLink/SDLScreenshotViewController.h
Toyota-BSalahat: Got it. Is there any advantage to doing this legwork vs just setting the window background as transparent? When we migrated the lock-screen to a UIWindow base we were able to achieve a similar effect without needing the additional logic. I am not as sure about iOS 7, but at least since iOS 8 this should work.

SmartDeviceLink/SDLCarWindow.m
joeljfischer: No, definitely not, virtually all of the CPU is used in the screenshotting.

SmartDeviceLink/SDLScreenshotViewController.h
joeljfischer: So the `UIWindow` is transparent, but it would still need a `rootViewController` to modally present the lock screen over, right? The rootVC could be transparent as well I suppose.

joeljfischer: I've attempted to modify the code to work the way you describe, but the backing window or view controller doesn't seem to be transparent. Creating the window this way,

```
_lockWindow = [[UIWindow alloc] initWithFrame:screenFrame];
_lockWindow.backgroundColor = [UIColor clearColor];
_lockWindow.opaque = NO;

UIViewController *rootController = [[UIViewController alloc] init];
rootController.view.backgroundColor = [UIColor clearColor];
rootController.view.opaque = NO;
_lockWindow.rootViewController = rootController;
```

NicoleYarroch:
The lock screen takes a long time to dismiss and is very jerky. It looks like the UI does not update while the lock screen is dismissing.

NicoleYarroch:
Why does the view resize once streaming has started? If I rotate the device from portrait -> landscape -> portrait, the app crashes.

joeljfischer:
@NicoleYarroch 

> The lock screen takes a long time to dismiss and is very jerky. It looks like the UI does not update while the lock screen is dismissing.

This is due to the screenshotting that takes place on the main queue, alongside the presentation / dismissal that also takes place on the main queue. I haven't thought of a possible / good solution to this yet. If you are not streaming, the presentation / dismissal should remain normal.

> Why does the view resize once streaming has started? 

This is the core of what CarWindow does, it literally screenshots and streams the view to the head unit, so, the view has to be resized or it will look like an iphone screen on the head unit.

> If I rotate the device from portrait -> landscape -> portrait, the app crashes.

I need more information from you about this, "crashes" doesn't help me much. But I do know that we will have to add documentation telling apps to stop allowing orientation changes when streaming, as we cannot do so for them.

Toyota-BSalahat:
> If I rotate the device from portrait -> landscape -> portrait, the app crashes.
>>I need more information from you about this, "crashes" doesn't help me much. But I do know that we will have to add documentation telling apps to stop allowing orientation changes when streaming, as we cannot do so for them.

Would it be possible to have a VPM UIViewController subclass that requests only landscape mode, and require that type as the projection view controller, instead of just a base UIViewController class? The overhead would be trivial for app implementers, and would give us an additional layer of control for things like this.

Toyota-BSalahat:
> I don't think so, for example, some implementers are probably just using the ordinary landscape orientation of their app and I don't think we'd want to force them to subclass a VC we define. There are few enough VPM apps that we should be able to document that they ought to prevent rotation while connected.

I can see both sides of this. On one hand, this is the path of least resistance for app developers, but on the other hand I am not sure we should encourage app developers to just project their normal interface to the head unit, since they really should tailor the VPM interface to the in-car experience.

Regardless, having a subclass won't prevent them from doing so, but may steer them away from just reusing their phone-centric UI.

The high level discussion on this may need to go to a wider audience.

NicoleYarroch:
> If I rotate the device from portrait -> landscape -> portrait, the app crashes.
>> I need more information from you about this, "crashes" doesn't help me much. But I do know that we will have to add documentation telling apps to stop allowing orientation changes when streaming, as we cannot do so for them.

After more investigation, this only happens if the `StreamingMediaManager`s `rootviewController` is also the `UIWindow`'s `rootviewController`. Since the width/height have changed on rotation, the `sdl_pixelBufferForImageRef` fails at `CGContextDrawImage` in the `SDLCarWindow` class.

SmartDeviceLink/SDLCarWindowViewController.h
```diff
@@ -0,0 +1,15 @@
...
+//  Created by Joel Fischer on 12/12/17.
+//  Copyright © 2017 smartdevicelink. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+@interface SDLCarWindowViewController : UIViewController
+
+@property (nonatomic, assign) UIInterfaceOrientation supportedOrientation;
+
```
NicoleYarroch: I tested the `SDLCarWindowViewController` with an app that supports all device orientations but with a view controller that only supported `landscapeRight`. This did not work because the view always loads initially in portrait mode must be rotated by the user in the `viewWillAppear` method. Once the view was rotated, the content disappeared off the screen and nothing could be streamed because the the screenshot returned a `nil` value. The only way I could get it to work was if the app only supports `landscapeRight` mode.

SmartDeviceLink/SDLStreamingMediaConfiguration.h
```diff
@@ -47,7 +47,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+ Activates the haptic view parser and CarWindow systems when set. This library will also use that information to attempt to return the touched view to you in `SDLTouchManagerDelegate`.
+
+ @note If you wish to alter this rootViewController while streaming via CarWindow, you must set a new `rootViewController` on `SDLStreamingMediaManager` and this will update both the haptic view parser and CarWindow.
+
+ @warning Apps using views outside of the `UIView` heirarchy (such as OpenGL) are currently unsupported. If you app uses partial views in the heirarchy, only those views will be discovered. Your OpenGL views will not be discoverable to a haptic interface head unit and you will have to manually make these views discoverable via the `SDLSendHapticData` RPC request.
+
+ @warning This is a weak property and it's therefore your job to hold a strong reference to this view controller.
+ */
+@property (strong, nonatomic, nullable) UIViewController *rootViewController;
```
NicoleYarroch: Why are we forcing user's to init a `rootViewController` in the `SDLStreamingMediaConfiguration`? Since it is getting initialized in the `AppDelegate` before the phone's `UIWindow` `rootViewController` is setup, I am forced to set a dummy `rootViewController ` and then set the actual view controller that I want to stream later.

SmartDeviceLink/SDLScreenshotViewController.m
```diff
@@ -0,0 +1,48 @@
...
+
+#import "SDLScreenshotViewController.h"
+
+@interface SDLScreenshotViewController ()
+
+@property (nonatomic, strong) UIImageView *imageView;
+
+@end
+
+@implementation SDLScreenshotViewController
```
NicoleYarroch: I am getting this warning: 
> Unbalanced calls to begin/end appearance transitions for <SDLScreenshotViewController: 0x157e1fc30>.

SmartDeviceLink/SDLStreamingMediaConfiguration.h
joeljfischer: After investigating about making it friendlier, it is not possible without a major version change due to the need to continue to support the `window` property in `SDLStreamingMediaConfiguration`.

SmartDeviceLink/SDLCarWindowViewController.h
joeljfischer: So should we lock the orientation to portrait?

SmartDeviceLink/SDLCarWindow.m
```diff
@@ -0,0 +1,213 @@
...
+
+- (void)sdl_sendFrame:(CADisplayLink *)displayLink {
+    if (!self.streamManager.isVideoConnected || self.streamManager.isVideoStreamingPaused) {
+        return;
+    }
+
+    if (self.isLockScreenMoving) {
+        SDLLogD(@"Paused CarWindow, lock screen moving");
+        return;
+    }
```
NicoleYarroch: The reason the `isLockScreenMoving` is not working is because of a race condition that occurs when two driver distraction notifications are triggered right after each other. For example, if a `DD_OFF` arrives during the middle of the lock screen being presented, `isLockScreenMoving` will be set to `YES`, `YES`, `NO`, `NO` instead of the expected `YES`, `NO`, `YES`, `NO`.

Instead we should check if the lock screen is being presented or dismissed: 
```
@property (assign, nonatomic, getter=isLockScreenPresenting) BOOL lockScreenPresenting;
@property (assign, nonatomic, getter=isLockScreenDismissing) BOOL lockScreenBeingDismissed;

- (void)sdl_sendFrame:(CADisplayLink *)displayLink {
    ...
    if (self.isLockScreenPresenting || self.isLockScreenDismissing) {
        SDLLogD(@"Paused CarWindow, lock screen moving");
        return;
    }
    ...
}

- (void)sdl_willPresentLockScreenViewController:(NSNotification *)notification {
    self.lockScreenPresenting = YES;
}

- (void)sdl_didPresentLockScreenViewController:(NSNotification *)notification {
    self.lockScreenPresenting = NO;
}

- (void)sdl_willDismissLockScreenViewController:(NSNotification *)notification {
    self.lockScreenBeingDismissed = YES;
}

- (void)sdl_didDismissLockScreenViewController:(NSNotification *)notification {
    self.lockScreenBeingDismissed = NO;
}
```

SmartDeviceLink/SDLStreamingMediaLifecycleManager.h
```diff
@@ -64,6 +65,10 @@ extern SDLAudioStreamState *const SDLAudioStreamStateShuttingDown;
  */
 @property (nonatomic, strong, readonly) SDLTouchManager *touchManager;
 
+@property (nonatomic, strong) UIViewController *rootViewController;
```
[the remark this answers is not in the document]
joeljfischer: It actually needs to be on `SDLStreamingMediaManager` as the lifecycle is private. This is fixed.

SmartDeviceLink/SDLCarWindow.m
```diff
@@ -0,0 +1,208 @@
...
+
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_didReceiveVideoStreamStarted:) name:SDLVideoStreamDidStartNotification object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_didReceiveVideoStreamStopped:) name:SDLVideoStreamDidStopNotification object:nil];
+
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_willPresentLockScreenViewController:) name:SDLLockScreenManagerWillPresentLockScreenViewController object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_willDismissLockScreenViewController:) name:SDLLockScreenManagerWillDismissLockScreenViewController object:nil];
+
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_willPresentLockScreenViewController:) name:SDLLockScreenManagerDidPresentLockScreenViewController object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_willDismissLockScreenViewController:) name:SDLLockScreenManagerDidDismissLockScreenViewController object:nil];
+
```
NicoleYarroch: The notifications are calling the wrong selector. It should be:

```
    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_willPresentLockScreenViewController:) name:SDLLockScreenManagerWillPresentLockScreenViewController object:nil];
    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_willDismissLockScreenViewController:) name:SDLLockScreenManagerWillDismissLockScreenViewController object:nil];

    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_didPresentLockScreenViewController:) name:SDLLockScreenManagerDidPresentLockScreenViewController object:nil];
    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_didDismissLockScreenViewController:) name:SDLLockScreenManagerDidDismissLockScreenViewController object:nil];
```

NicoleYarroch:
Presenting the lock screen controller when using a TCP connection causes the app to crash with the following error:

> Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Application tried to present modally an active controller <SDLScreenshotViewController: 0x7fad95d068b0>.'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000011271a12b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x000000010ea50f41 objc_exception_throw + 48
	2   UIKit                               0x000000010fbb40b7 -[UIViewController _presentViewController:withAnimationController:completion:] + 5168
	3   UIKit                               0x000000010fbb69a9 __63-[UIViewController _presentViewController:animated:completion:]_block_invoke + 99
	4   UIKit                               0x000000011076a811 -[_UIViewControllerTransitionCoordinator _applyBlocks:releaseBlocks:] + 294
	5   UIKit                               0x00000001107663c0 -[_UIViewControllerTransitionContext _runAlongsideCompletions] + 155
	6   UIKit                               0x0000000110766098 -[_UIViewControllerTransitionContext completeTransition:] + 118
	7   UIKit                               0x000000010fb6adba -[UITransitionView notifyDidCompleteTransition:] + 251
	8   UIKit                               0x000000010fb6aa31 -[UITransitionView _didCompleteTransition:] + 1397
	9   UIKit                               0x000000010fb6d0c8 -[UITransitionView _transitionDidStop:finished:] + 104
	10  UIKit                               0x000000010fa8e6c4 -[UIViewAnimationState sendDelegateAnimationDidStop:finished:] + 343
	11  UIKit                               0x000000010fa8ed23 -[UIViewAnimationState animationDidStop:finished:] + 293
	12  UIKit                               0x000000010fa8edd7 -[UIViewAnimationState animationDidStop:finished:] + 473
	13  QuartzCore                          0x00000001154996bd _ZN2CA5Layer23run_animation_callbacksEPv + 323
	14  libdispatch.dylib                   0x0000000113eea33d _dispatch_client_callout + 8
	15  libdispatch.dylib                   0x0000000113ef55f9 _dispatch_main_queue_callback_4CF + 628
	16  CoreFoundation                      0x00000001126dce39 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
	17  CoreFoundation                      0x00000001126a1462 __CFRunLoopRun + 2402
	18  CoreFoundation                      0x00000001126a0889 CFRunLoopRunSpecific + 409
	19  GraphicsServices                    0x0000000117ffd9c6 GSEventRunModal + 62
	20  UIKit                               0x000000010f9f85d6 UIApplicationMain + 159
	21  Trippy                              0x000000010d8655e7 main + 55
	22  libdyld.dylib                       0x0000000113f66d81 start + 1
	23  ???                                 0x0000000000000001 0x0 + 1
)

This happens with a TCP connection only. This also only fails the **second** time the lock screen is presented.

SmartDeviceLink/SDLCarWindow.m
joeljfischer: Fixed. Thanks autocomplete.

SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
```diff
@@ -784,7 +824,17 @@ - (void)sdl_sendVideoStartService {
...
+#pragma mark Setters / Getters
+
+- (void)setRootViewController:(UIViewController *)rootViewController {
+    if (self.focusableItemManager != nil) {
+        self.focusableItemManager.viewController = rootViewController;
+    }
+
+    if (self.carWindow != nil) {
+        self.carWindow.rootViewController = rootViewController;
+    }
```
NicoleYarroch: This only works for the first `rootViewController`. Setting a new `rootViewController` crashes the app because the frame of the new `rootViewController` is the same as the dimensions of the device screen instead of the SDL accessory screen size. The frame needs to be reset again every time a `rootViewController` is set.

```
self.rootViewController.view.frame = CGRectMake(0, 0, self.streamManager.screenSize.width, self.streamManager.screenSize.height);
self.rootViewController.view.bounds = self.rootViewController.view.frame;
```

SmartDeviceLink/SDLLockScreenViewController.m
```diff
@@ -39,6 +39,10 @@ - (BOOL)shouldAutorotate {
     return NO;
 }
 
+- (UIInterfaceOrientationMask)supportedInterfaceOrientations {
+    return UIInterfaceOrientationMaskPortrait;
+}
+
```
NicoleYarroch: Sometimes the app crashes right after connection to the SDL accessory due to the background color of the lock screen not being set on the main thread. The crash occurs in the `setBackgroundColor:backgroundColor` function.

SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
joeljfischer: Should be fixed

SmartDeviceLink/SDLLockScreenViewController.m
joeljfischer: Should be fixed

SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
```diff
@@ -393,6 +408,23 @@ - (void)didEnterStateVideoStreamReady {
...
+            NSInteger targetFramerate = ((NSNumber *)self.videoEncoderSettings[(__bridge NSString *)kVTCompressionPropertyKey_ExpectedFrameRate]).integerValue;
+            SDLLogD(@"Initializing CADisplayLink with framerate: %ld", (long)targetFramerate);
+            self.displayLink = [CADisplayLink displayLinkWithTarget:self selector:@selector(sdl_displayLinkFired:)];
+            if (SDL_SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"10")) {
+                self.displayLink.preferredFramesPerSecond = targetFramerate;
+            } else {
+                self.displayLink.frameInterval = (60 / targetFramerate);
+            }
+            [self.displayLink addToRunLoop:[NSRunLoop currentRunLoop] forMode:NSRunLoopCommonModes];
+        });
```
NicoleYarroch: I am having issues with the video frame going blurry every few seconds when nothing on the screen is changing. For example if I have an image on the SDL screen, it is relatively crisp but a second later, it goes blurry, and then back to being sharp again.

joeljfischer: Interesting. I assume that this is related to new restrictions that were added on the amount of data allowed to be sent per 5 seconds. This limit may have been too low.

SmartDeviceLink/SDLCarWindow.m
```diff
@@ -0,0 +1,213 @@
...
+    UIGraphicsBeginImageContextWithOptions(bounds.size, YES, 1.0f);
+    [self.rootViewController.view drawViewHierarchyInRect:bounds afterScreenUpdates:self.drawsAfterScreenUpdates];
+    UIImage *screenshot = UIGraphicsGetImageFromCurrentImageContext();
+    UIGraphicsEndImageContext();
+
+    CGImageRef imageRef = screenshot.CGImage;
+    CVPixelBufferRef pixelBuffer = [self.class sdl_pixelBufferForImageRef:imageRef usingPool:self.streamManager.pixelBufferPool];
+    [self.streamManager sendVideoData:pixelBuffer];
+    CVPixelBufferRelease(pixelBuffer);
+}
```
NicoleYarroch: Gestures on the phone screen are delayed ~3-5 seconds when using `drawViewHierarchyInRect`.  It appears to be a known bug: https://stackoverflow.com/questions/23157653/drawviewhierarchyinrectafterscreenupdates-delays-other-animations/25704861. 

I tried implementing the suggested answer and gestures were recognized instantaneously. 

The trade-off is that the screen shot may not capture everything accurately (i.e. blurs and what not)

SmartDeviceLink/SDLCarWindow.m
```diff
@@ -0,0 +1,213 @@
...
+
+    if (self.isLockScreenPresenting || self.isLockScreenDismissing) {
+        SDLLogD(@"Paused CarWindow, lock screen moving");
+        return;
+    }
+
+    CGRect bounds = self.rootViewController.view.bounds;
+
+    UIGraphicsBeginImageContextWithOptions(bounds.size, YES, 1.0f);
+    [self.rootViewController.view drawViewHierarchyInRect:bounds afterScreenUpdates:self.drawsAfterScreenUpdates];
```
NicoleYarroch: I think the captured video could be much sharper. Text is hard to read because of a little halo around each letters' edge.

joeljfischer: The problem is that bandwidth is very restricted, which means we have to make a tradeoff between framerate and sharpness, and we have to err to framerate right now. It's about as good as it currently gets.

SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
NicoleYarroch: It is very noticeable as images and text seem to flash and pulse every few seconds.

NicoleYarroch: I tested the **up the max data rate limit / 5 second** change and it did not fix the flickering.

joeljfischer: Removed the data limit

SmartDeviceLink/SDLCarWindowViewController.h
NicoleYarroch: I think setting the frame so part of the view controller is missing because it is off the screen is confusing from a UX standpoint. I think it would be better to force the orientation to landscape and then recalculate the `rootViewController`'s bounds so it matches the SDL accessory screen size but the entire view still fits on the screen. One negative is that this will complicate the `SDLTouchManager` because it means that it must scale touch points when the car window feature is being utilized.

SmartDeviceLink/SDLCarWindowViewController.h
```diff
@@ -0,0 +1,15 @@
+//
+//  SDLCarWindowViewController.h
+//  SmartDeviceLink
+//
+//  Created by Joel Fischer on 12/12/17.
+//  Copyright © 2017 smartdevicelink. All rights reserved.
+//
+
```
NicoleYarroch: I think we should note in the documentation that orientation locking will not work if the `viewController` is embedded in a `UINavigationController` or a `UITabBarController`.

SmartDeviceLink/SDLCarWindow.m
NicoleYarroch: This is no longer an issue because the hard data limit was removed.

SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
NicoleYarroch: Removing the hard data limit fixes the flickering issue.

SmartDeviceLink/SDLLockScreenViewController.m
NicoleYarroch: This seems to be fixed. I haven't had an issue since the fix was pushed.

SmartDeviceLink/SDLCarWindowViewController.h
joeljfischer: I don't think that will work. Accessory sizes could be in portrait (e.g. if someone made a screen like Tesla's) and could be of all sorts of resolutions. It would be more helpful to always have the lock screen displayed.

SmartDeviceLink/SDLCarWindow.m
```diff
@@ -0,0 +1,216 @@
...
+    self.lockScreenBeingDismissed = NO;
+}
+
+#pragma mark - SDLNavigationLifecycleManager Notifications
+- (void)sdl_didReceiveVideoStreamStarted:(NSNotification *)notification {
+    dispatch_async(dispatch_get_main_queue(), ^{
+        // If the video stream has started, we want to resize the streamingViewController to the size from the RegisterAppInterface
+        self.rootViewController.view.frame = CGRectMake(0, 0, self.streamManager.screenSize.width, self.streamManager.screenSize.height);
+        self.rootViewController.view.bounds = self.rootViewController.view.frame;
+
```
NicoleYarroch: The `rootViewController` is not resized when the USB cord is pulled before video is streamed.

1. Open the app on the phone and then connect the phone to the TDK. The `rootViewController` will resize.
2. Do **not** open the app on the TDK.
3. Pull the USB cord. 

The `rootViewController`'s frame revert back to its original frame but it stays the same.

SmartDeviceLink/SDLLockScreenPresenter.m
```diff
@@ -8,61 +8,108 @@
...
-            presented = (self.viewController.isViewLoaded && (self.viewController.view.window || self.viewController.isBeingPresented));
+            isPresented = [self sdl_presented];
         });
-    } else {
-        presented = (self.viewController.isViewLoaded && (self.viewController.view.window || self.viewController.isBeingPresented));
     }
-
-    return presented;
+    
+    return self;
```
NicoleYarroch: The lock screen is no longer working because the wrong value is being returned for `presented`. If `return self` is changed to `return isPresented` the lock screen works as expected.

SmartDeviceLink/SDLCarWindowViewController.h
joeljfischer: Done

SmartDeviceLink/SDLCarWindowViewController.h
NicoleYarroch: We could just resize the `rootViewController` to best fit whichever the orientation the developer has chosen to lock the device. 

I think showing the lock screen all the time should not be the only option available to the developer. If a user wants to use the app on the phone while the car is parked we are inconveniencing them by making them pull and reconnect the USB cable multiple times.

SmartDeviceLink/SDLLockScreenPresenter.m
joeljfischer: Fixed

SmartDeviceLink/SDLCarWindowViewController.h
joeljfischer: I want to note that this is _only_ an issue if the dev uses the root of the UIWindow to stream. If they use an off-screen VC it will work fine. That's the best practice. If the dev wants to stream the same VC being displayed on the phone, this _will never_ look right or be usable to the user. Developers shouldn't have to (and cannot) try and lock their app in an orientation based on the size of the connected screen.

SmartDeviceLink/SDLCarWindow.m
NicoleYarroch: The `rootViewController` is not being resized correctly after reconnecting to the head unit if the app on the device is in the background.

1. Connect phone to TDK
1. Open app on TDK
1. Disconnect phone from TDK by pulling the USB cord
1. Put app on phone in the background
1. Reconnect the phone to the TDK using the USB cord
1. The dialog box with the spinner should show up (waiting for video)
1. Bring the app on the phone to the foreground

The `rootViewController`'s frame is not being set to the dimensions of the TDK screen so the video stream does not fill the entire screen.

SmartDeviceLink/SDLCarWindowViewController.h
NicoleYarroch: Good to know. The proposal is not very clear about how the car window API is supposed to be used so I assumed that it was meant to support both off-screen and visible screens.

SmartDeviceLink/SDLCarWindow.m
joeljfischer: I've added an option for layer based rendering or view based rendering in the configuration.

SmartDeviceLink/SDLCarWindowViewController.h
joeljfischer: It does support visible screens, but there are some inherent limitations is how I would describe it.

SmartDeviceLink/SDLCarWindow.m
```diff
@@ -0,0 +1,226 @@
...
+    });
+}
+
+- (void)sdl_didReceiveVideoStreamStopped:(NSNotification *)notification {
+    dispatch_async(dispatch_get_main_queue(), ^{
+        // And also reset the streamingViewController's frame, because we are about to show it.
+        self.rootViewController.view.frame = [UIScreen mainScreen].bounds;
+        SDLLogD(@"Video stream ended, setting view controller frame back: %@", NSStringFromCGRect(self.rootViewController.view.frame));
+    });
+}
```
NicoleYarroch: Another issue that I am having on reconnects is that sometimes when I try to get the topmost view controller from the `UIApplication`, it returns the `SDLScreenshotViewController`. I assume that is is because the `SDLScreenshotViewController` is not being dismissed when video stream is stopped.

NicoleYarroch: This turned out to be a user error on my part and not an issue with the library.

NicoleYarroch:
I am getting one failed test case in the `SDLLifecycleManagerSpec` on line 212.

SmartDeviceLink/SDLCarWindow.m
NicoleYarroch: This turned out to be a user error on my part and not an issue with the library. In the example above I was setting the `rootViewController` using an on-screen view controller. If the `rootViewController` was set before the on-screen view controller rendered its layout, resizing the frame failed.

joeljfischer: This is also an issue with using the app's current `UIWindow rootViewController`. More investigation may need to be done about using it in this way.

SmartDeviceLink/SDLCarWindow.m
joeljfischer: What was the user error, is there anything we can document to help people if they encounter the same issue?

SmartDeviceLink/SDLCarWindow.m
joeljfischer: Can you describe your fix? Is there any documentation we can add?

SmartDeviceLink/SDLCarWindow.m
NicoleYarroch: It had to do with setting the `rootViewController` with an on-screen view controller at the wrong time. I added an in-depth description of the issue to a comment left on the `SDLCarWindow.m` class.

SmartDeviceLink/SDLCarWindow.m
joeljfischer: I added documentation around this to the configuration of `rootViewController` and to the guide content.

SmartDeviceLink/SDLCarWindow.m
joeljfischer: Done

SmartDeviceLink/SDLCarWindow.m
NicoleYarroch: The reason this is happening is because the `rootViewController` frame is resized in the `setRootViewController:rootViewController` method of the `SDLCarWindow` class even if video is not streaming. Since the frame also gets set in the `sdl_didReceiveVideoStreamStarted:notification` method, the `setRootViewController:rootViewController` method should check if video is currently streaming before setting the `rootViewController`'s frame

joeljfischer: Fixed

Issue #803, opened 2017-11-30
Audio stream is not stopped at hmi level none and background


t-yoshii:
### Bug Report
Audio stream is not stopped when hmi level are changed to none or background from limited or full.
According to this guide, audio streaming is only allowed at hmi level limited and full.
https://smartdevicelink.com/en/guides/iOS/mobile-navigation/audio-streaming/

##### Reproduction Steps
1. Start SDL app (appHMIType = Navigation) and connect to SDL core
2. Change HMI level to full
3. Change HMI level to background or none

##### Expected Behavior
Audio streaming and video streaming are stopped 

##### Observed Behavior
Audio streaming is not stopped. Only video streaming is stopped.

##### OS & Version Information
* iOS Version: 11.1.2
* SDL iOS Version: 5.0.0-rc3
* Testing Against: Our in-house testing SDL HMI devboard

##### Test Case, Sample Code, and / or Example App
PR will be created.

joeljfischer:
This does not seem to have been fixed in the AudioStreamManager

PR #807, opened 2017-12-05
AudioStreamManager

### LINKED ISSUE #795: [SDL 0113] SDLAudioStreamManager
Proposal: [SDLAudioStreamManager](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0113-audiostreammanager.md)

> This proposal provides an interface for automatically transcoding audio files – such as mp3 – to the SDL compatible PCM format, queueing these files, and playing them using the PCM audio stream available to `NAVIGATION` apps.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/347

Steering Committee Decision:

> When reviewing this proposal, there was a question about a use case for audio streaming. The author of this proposal advised that audio streaming would be used for navigation or projection apps only, and that navigation apps that are playing audio (voice of navigation directions, for example) often don't have directions available in PCM format, or the correct PCM format. The AudioStreamManager will take the format that the app has (MP3, for example), convert it into PCM format and play. This allows the app to use the manager instead of having to convert the file themselves.  After this discussion, the Steering Committee voted to accept this proposal.

joeljfischer:
Fixes #795, #803 

### Risk

### Testing Plan
Unit Tests will be added

### Summary
Adds an implementation of [SDL-0113](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0113-audiostreammanager.md). This proposal provides an interface for automatically transcoding audio files – such as mp3 – to the SDL compatible PCM format, queueing these files, and playing them using the PCM audio stream available to `NAVIGATION` apps.

##### Enhancements
* Adds `SDLAudioStreamManager` that does on-the-fly transcoding of audio files to an SDL compatible PCM format and can send that audio to be played. This is for `NAVIGATION` applications only.

##### Bug Fixes
* Audio stream shouldn't be viewed as available in HMI NONE.

### Tasks Remaining:
- [x] Finish Unit Tests
- [x] SDLStreamingMediaManager integration

### CLA


joeljfischer:
@NicoleYarroch fixed

SmartDeviceLink/SDLAudioStreamManager.h
```diff
@@ -0,0 +1,60 @@
...
+
+ @note This happens on a global background thread and will provide an error callback using the delegate if the conversion fails.
+
+ @param fileURL File URL to convert
+ */
+- (void)pushWithFileURL:(NSURL *)fileURL;
+
+/**
+ Play the next item in the queue. If an item is currently playing, it will continue playing and this item will begin playing after it is completed. If all items are playing via `playAll`, this will do nothing.
+
```
NicoleYarroch: I can't find a "playAll" setting.

SmartDeviceLink/SDLAudioStreamManager.h
```diff
@@ -0,0 +1,60 @@
...
+- (instancetype)initWithManager:(id<SDLStreamingAudioManagerType>)streamManager NS_DESIGNATED_INITIALIZER;
+
+/**
+ Push a new file URL onto the queue after converting it into the correct PCM format for streaming binary data.
+
+ @note This happens on a global background thread and will provide an error callback using the delegate if the conversion fails.
+
+ @param fileURL File URL to convert
+ */
+- (void)pushWithFileURL:(NSURL *)fileURL;
```
NicoleYarroch: You should probably note that the conversion is asynchronous. That is, if the user uploads a bunch of sound files at once, the files will not necessarily play in the order they were uploaded.

joeljfischer: What do you mean by "uploads?" They should absolutely play in the order in which they were "pushed"

SmartDeviceLink/SDLAudioStreamManager.h
joeljfischer: There is no `playAll` setting, and was none in the agreed upon interface. If you wanted to do something like this, right now you'd have to be the delegate and play the next one when the previous one finishes.

NicoleYarroch: It says "playAll" in the comment:
> + Play the next item in the queue. If an item is currently playing, it will continue playing and this item will begin playing after it is completed. If all items are playing via `playAll`, this will do nothing.

SmartDeviceLink/SDLAudioStreamManager.h
NicoleYarroch: I uploaded 3 sound files right after eachother other in this order: [A (large file), B (small file), C (small file)] and it played in this order: [B, A, C]

SmartDeviceLink/SDLAudioStreamManager.h
joeljfischer: Ooops 😅

SmartDeviceLink/SDLAudioStreamManager.h
joeljfischer: What does the code look like to push and play here? Did you wait for them all to encode? Did you check the queue to see what files were there in what order? Any chance you did any debugging to see why? They're encoded on a serial queue, so one absolutely should not finish before one started before it.

NicoleYarroch: I see what is happening. For each file I pushed, I also triggered a `playNextWhenReady` notification at the same time since I wanted them to play right after each other. The delegate is notified when a track finished based on the length of the song and not on when the track actually finished.

PR #810, opened 2017-12-07
SDLManager hmiLevel should be nullable

### LINKED ISSUE #805: The app will crash if the hmiLevel is set to nil
### Bug Report
The `SDLManager`'s `hmiLevel` property is assumed to be non null. It's `hmiLevel` is obtained from the `SDLLifecycleManager`,  but the `SDLLifecycleManager`'s `hmiLevel`is nullable. Since `hmiLevel` can be`nil`, the app will crash if the `hmiLevel` is set to `nil`. Checking the nullability of the `hmiLevel` generates a compiler warning. 

##### Reproduction Steps
1. Connect the app to a TDK using an iAP connection and USB cord.
1. Start the app
1. Pull the USB cord

##### Expected Behavior
The app does not crash

##### Observed Behavior
Since, the hmiLevel gets set to `nil` on disconnect , the app may crash if there is any logic in app that checks the `hmiLevel` and assumes that the `hmiLevel` is non null.

* iOS Version: 11
* SDL iOS Version: 5.0.0
* Testing Against: TDK (SYNC 3), App using Swift

This error was found while testing the Livio Music Player app.

joeljfischer:
Fixes #805 

### Risk

### Testing Plan
n/a

### Summary
Fix a possible crash when `hmiLevel`, marked as `nonnull`, but can be `nil` is checked and is `nil`.

##### Bug Fixes
* Fix a possible crash when accessing `SDLManager hmiLevel`

### CLA

### DIFF
--- a/SmartDeviceLink/SDLManager.h
+++ b/SmartDeviceLink/SDLManager.h
@@ -38,7 +38,7 @@ typedef void (^SDLManagerReadyBlock)(BOOL success, NSError *_Nullable error);
 /**
  *  The current HMI level of the running app.
  */
-@property (copy, nonatomic, readonly) SDLHMILevel hmiLevel;
+@property (copy, nonatomic, readonly, nullable) SDLHMILevel hmiLevel;
 
 /**
  *  The current audio streaming state of the running app.


SmartDeviceLink/SDLManager.h
```diff
@@ -38,7 +38,7 @@ typedef void (^SDLManagerReadyBlock)(BOOL success, NSError *_Nullable error);
 /**
  *  The current HMI level of the running app.
  */
-@property (copy, nonatomic, readonly) SDLHMILevel hmiLevel;
+@property (copy, nonatomic, readonly, nullable) SDLHMILevel hmiLevel;
```
NicoleYarroch: I am getting this warning in the *.m* file: `Nullability specifier 'nonnull' conflicts with existing specifier '_Nullable'`.

It should be `- (nullable SDLHMILevel)hmiLevel {...}`

joeljfischer: Fixed

PR #817, opened 2017-12-07
IAP Transport Optimizations

### LINKED ISSUE #816: SDLIAPTransport Does Not Reconnect Successfully
### Bug Report
The current IAP transport implementation often fails during the reconnection logic. Rapid disconnects/reconnects amplify the issue.

##### Reproduction Steps
1. Connect a SDL app.
2. Disconnect a SDL app, or initiate a transport mode switch if supported on the head unit
3. Observe if app successfully reconnects
4. Repeat steps 1-3

##### Expected Behavior
SDL app successfully connects

##### Observed Behavior
SDL apps often do not connect

This most often is due to two small logical issues inside the transport adapter

1. When the proxy reconnects after a delay, it holds a reference to the existing accessory. In many cases the accessory could have disconnected and reconnected inside of that delay, and the referenced accessory becomes invalid. Attempting to connect to the invalid accessory causes connection failure.

2. There is a race condition that occurs with the protocolIndexTimer which in many cases can cause the start selector to fire after the cancel selector.

* iOS Version: iOS 10 and 11, including subversions.
* SDL iOS Version: Issue observed in 5.0 release, and 4.7.3 release.
* Testing Against: Head unit

PR 817

Toyota-BSalahat:
Fixes #816

### Risk

### Testing Plan
Changes should be covered under any existing transport unit tests. Changes have been tested on our head units and our TDK. Changes have also been tested by a SDL application partner.

### Summary
Transport layer no longer uses reference to previously connected accessory during retry/reconnection. Fresh accessory is always queried from the accessory manager.

Race condition on protocol index timer no longer causes reconnection to fail.


SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -139,7 +139,7 @@ - (void)sdl_stopEventListening {
  *  @param notification Contains information about the connected accessory
  */
 - (void)sdl_accessoryConnected:(NSNotification *)notification {
-    EAAccessory *accessory = notification.userInfo[EAAccessoryKey];
+    //EAAccessory *accessory = notification.userInfo[EAAccessoryKey];
```
joeljfischer: Please remove the commented out code

SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -506,7 +509,7 @@ - (SDLStreamHasBytesHandler)sdl_controlStreamHasBytesHandlerForAccessory:(EAAcce
         SDLLogD(@"Control Stream will switch to protocol %@", indexedProtocolString);
         
         // Destroy the control session
-        [strongSelf.protocolIndexTimer cancel];
+        //[strongSelf.protocolIndexTimer cancel];
```
joeljfischer: Please remove the commented out code

Toyota-BSalahat:
@joeljfischer the branch for this PR should now be based on develop. There was one minor merge conflict in the - (void)sdl_accessoryConnected:(NSNotification *)notification function from where you removed the retryCounter reset that you may want to double check.

PR #818, opened 2017-12-07
Wrap TCP transport connect logic inside of an operation queue

### LINKED ISSUE #760: Attempting to start sdl tcp connection using dispatch async causes connection to never succeed/fail
### Bug Report
Wrapping a SDLManager (configured for TCP) start call using the standard iOS dispatch async method results in the connection never returning a success or failure in its callback block. Presumably this is caused by GCD causing some failure in the initialization of the NSRunLoop. 
This issue does not occur when establishing an iAP connection.

##### Reproduction Steps
1. Pull down the branch located [here](https://github.com/Toyota-BSalahat/sdl_ios/tree/bug/TCPConnectionBackgrounding)
2. Uncomment the NSOperationQueue wrapper I added in the connect function inside SDLTCPTransport.m
3. Run the example app and attempt to connect to a valid instance of Manticore

##### Expected Behavior
The example app's UI is not frozen while a TCP connection is attempted, the app connects or fails to connect to the Manticore instance.

##### Observed Behavior
App will remain in the "connecting" state forever. SDLManagers start callback is never called with failure or success.

* iOS Version: iOS 11.0-11.02
* SDL iOS Version: 4.7.3
* Testing Against: The modified example app linked above

[Modified example app](https://github.com/Toyota-BSalahat/sdl_ios/tree/bug/TCPConnectionBackgrounding)

The issue can be toggled on/off by removing/adding the NSOperation Queue wrapper I added to the SDLTCPTransport.m connect function.

Establishing the run loop on its on operation queue seems to adequately address this problem, but there may be other solutions.

Toyota-BSalahat:
Fixes #760

### Risk

### Testing Plan
Changes should be covered under any current TCP connection unit tests. Changes have been tested against Manticore and emulated SDL core instances.

### Summary
The TCP transport's connect function is now processed on its own operation queue.

### DIFF
--- a/SmartDeviceLink/SDLTCPTransport.m
+++ b/SmartDeviceLink/SDLTCPTransport.m
@@ -46,20 +46,22 @@ static void TCPCallback(CFSocketRef socket, CFSocketCallBackType type, CFDataRef
 }
 
 - (void)connect {
-    SDLLogD(@"Attemping to connect");
-
-    int sock_fd = call_socket([self.hostName UTF8String], [self.portNumber UTF8String]);
-    if (sock_fd < 0) {
-        SDLLogE(@"Server not ready, connection failed");
-        return;
-    }
-
-    CFSocketContext socketCtxt = {0, (__bridge void *)(self), NULL, NULL, NULL};
-    socket = CFSocketCreateWithNative(kCFAllocatorDefault, sock_fd, kCFSocketDataCallBack | kCFSocketConnectCallBack, (CFSocketCallBack)&TCPCallback, &socketCtxt);
-    CFRunLoopSourceRef source = CFSocketCreateRunLoopSource(kCFAllocatorDefault, socket, 0);
-    CFRunLoopRef loop = CFRunLoopGetCurrent();
-    CFRunLoopAddSource(loop, source, kCFRunLoopDefaultMode);
-    CFRelease(source);
+    [[NSOperationQueue mainQueue] addOperationWithBlock:^{
+        SDLLogD(@"Attemping to connect");
+        
+        int sock_fd = call_socket([self.hostName UTF8String], [self.portNumber UTF8String]);
+        if (sock_fd < 0) {
+            SDLLogE(@"Server not ready, connection failed");
+            return;
+        }
+        
+        CFSocketContext socketCtxt = {0, (__bridge void *)(self), NULL, NULL, NULL};
+        socket = CFSocketCreateWithNative(kCFAllocatorDefault, sock_fd, kCFSocketDataCallBack | kCFSocketConnectCallBack, (CFSocketCallBack)&TCPCallback, &socketCtxt);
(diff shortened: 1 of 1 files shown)


joeljfischer:
After a number of tests between master and this branch, I'm not seeing any differences in behavior. All of the TCPTransport appears to still occur in the main thread / queue, it's essentially the same as calling `dispatch_async(dispatch_get_main_queue(), ^{});`

Toyota-BSalahat:
I will double check, but when tested vs master I was seeing different behavior.

On master, calling connect with a TCP configuration within a dispatch async call as described in the issue would result in the connect logic firing, but the connection never being established, and the manager.start callback never firing with a success or error. From what I could tell, this was caused by an issue that dispatch async was causing with attaching to the run loop.

With the operation queue wrapper, this issue was not seen to occur. I assume this is caused by a minor difference between how the run loop is attached when called from the application level with dispatch async vs locally with the operation queue which results in the connection logic completing successfully.

This simply prevents the phone from hanging between calling manager.start with a TCP connection, and the connection being established. For end-to-end background support, I think we would have to migrate to using a NSThread like we do in the IAP transport.

Toyota-BSalahat:
@joeljfischer I can confirm I still see the issue on the 5.0 proxy. If you dispatch the manager.start call with a TCP config from the app side, the manager.start callback is not firing and a connection is never established.

joeljfischer:
@Toyota-BSalahat Sorry if there was a miscommunication, I meant that I'm not seeing any real differences in behavior between this PR and master as it currently exists. I'm not entirely sure #760 is an issue, after all, you can just not wrap it in a dispatch_async call, right? What is this attempting to solve?

Toyota-BSalahat:
@joeljfischer we definitely see the change in behavior with our testing. Using the standard dispatch to background the connection logic from the app-side without the changes in this PR prevents the TCP connection from ever forming.

As it stands, calling the connection logic on the main thread without a background dispatch locks up an application's UI. This makes it difficult/impossible to cancel the connection operation. This issue does not occur when using IAP transport, since IAP transport has its runloop in its own thread. TCP does not appear to have this advantage.

It's a minor annoyance, but we would like to see the TCP connection be able to run in a standard dispatch_async call. I think if possible in the future it would be good to give the TCP run loop its own thread (or another full solution to actually get the processing off the main thread). But this alleviates a minor pain point we see now.

joeljfischer:
👍 Sounds good. I would love to be able to use https://www.github.com/robbiehanson/CocoaAsyncSocket

It's dual licensed with public domain, but it would be adding a dependency and may be slight overkill.

joeljfischer:
@Toyota-BSalahat I did not notice this targeted master until after the merge, it's been reverted. Please re-submit – mentioning this PR – and target `develop` instead.

joeljfischer:
@Toyota-BSalahat I'm still waiting on this

Toyota-BSalahat:
@joeljfischer, should be updated to point to develop on PR #828

PR #829, opened 2017-12-19
Wrap TCP transport connect logic inside of an operation queue

### LINKED ISSUE #760: Attempting to start sdl tcp connection using dispatch async causes connection to never succeed/fail

Toyota-BSalahat:
Fixes #760, updated version of PR #818

This PR is [ready] for review.

Risk

This PR makes [no] API changes.

Testing Plan

Changes should be covered under any current TCP connection unit tests. Changes have been tested against Manticore and emulated SDL core instances.

Summary

The TCP transport's connect function is now processed on its own operation queue


Toyota-BSalahat:
@joeljfischer the last PR was pulling from the wrong forked branch. This one should be correct with no conflicts.

PR #857, opened 2018-01-31
Batch sending of RPC Requests

### LINKED ISSUE #723: [SDL 0087] Sequentially Send Multiple RPCs
Proposal: [Sequentially Send Multiple RPCs](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0087-send-multiple-rpcs-ios.md)

> This proposes to add an API allowing the developer to sequentially send multiple RPCs and to be notified when all have completed sending.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/263

Steering Committee Decision:

> The Steering Committee voted to accept this proposal with the following revisions: revise to add Android counterpart to this proposal. Additionally, when implementing, will need to investigate at what rate RPCs can be sent and received, as well as which RPCs do not need to be sent asynchronously.  Revisions to add the Android counterpart are now included within the .md file.

joeljfischer:
Fixes #723 

### Risk

### Testing Plan
Unit tests will be added

### Summary
Add support for batch sending RPCs sequentially. Either one right after another, or waiting for the last to return before sending the next.

##### Enhancements
* Add support for batch sending RPCs sequentially. Either one right after another, or waiting for the last to return before sending the next.

### Tasks Remaining:
- [x] Testing
- [x] Add sequential RPC operation
- [x] Hook up operations to `SDLLifecycleManager`

### CLA


joeljfischer:
@NicoleYarroch 

> Known issue: On both SYNC 3 and Manticore, sending multiple requests simultaneously usually failed after first 2-5.

What do you mean by "sending multiple requests simultaneously"? And does it recover?

NicoleYarroch:
@joeljfischer 

To test this PR, I sent 13 `CreateInteractionChoiceSet`s at once. If I send them one-by-one, so the next one is only sent after a response is received from Core, everything works fine. If I send them all at once, without waiting for a response, the first 2-5 will succeed but the rest will fail with `GENERIC_ERROR`. The latter error is true for both Manticore and SYNC 3. Both sends work when testing with SDL_Core 4.4.1 and the Generic HMI.

This error has been reported to the Manticore team. Due to more pressing issues, they will not be able to investigate immediately.

PR #878, opened 2018-02-26
Video stream is stopped as soon as app enters background

### LINKED ISSUE #858: Video stream is stopped and restarted immediately when app enters background
### Bug Report
When app enters the background, video streaming should pause until the app re-enters the foreground. Currently, video streaming stops and then restarts immediately. A `SDLVideoStreamDidStart` notification is sent even when the app is in the background. 

##### Reproduction Steps
1. Connect a video streaming app (`appType` = `NAVIGATION`) to Core.
1. Once video streaming has started, put the app on the iPhone into the background.

##### Expected Behavior
The `SDLVideoStreamDidStart` notification is not sent until the app on the iPhone is brought to the foreground.

##### Observed Behavior
The `SDLVideoStreamDidStart` notification is sent while the app is still in the background.

* iOS Version: 11.2
* SDL iOS Version: 5.1
* Testing Against: SDL Core 4.4

NicoleYarroch:
Fixes #858

### Risk

### Testing Plan
Completed the smoke tests for video streaming from the SDL_iOS test plan.

### Summary
* Video streaming was not stopped when the app on the phone entered the background. When the app re-entered the foreground, video streaming was stopped and then immediately restarted. Video streaming is now stopped as soon as the app enters the background.

##### Breaking Changes
* None

### CLA

### DIFF
--- a/SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
+++ b/SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
@@ -291,7 +291,9 @@ typedef void(^SDLVideoCapabilityResponseHandler)(SDLVideoStreamingCapability *_N
 
     [self sdl_sendBackgroundFrames];
     [self.touchManager cancelPendingTouches];
-    self.restartVideoStream = YES;
+
+    [self sdl_stopAudioSession];
+    [self sdl_stopVideoSession];
 }
 
 // Per Apple's guidelines: https://developer.apple.com/library/content/documentation/iPhone/Conceptual/iPhoneOSProgrammingGuide/StrategiesforHandlingAppStateTransitions/StrategiesforHandlingAppStateTransitions.html


t-yoshii:
@NicoleYarroch @joeljfischer 
Should AudioStreaming be stopped at application becomes inactive state?

VideoStreaming should be stopped because VideoStreaming cannot continue in inactive state due to iOS limitations. However, AudioStreaming can continue even in inactive state. 
For navigation usability, I think it is good to continue navigation guidance by AudioStreaming even in inactive state.

Can I remove this line? If you think OK, I will create issue and PR.
https://github.com/smartdevicelink/sdl_ios/pull/878/commits/509c92748c16b61746beea4a2fecc51065c2252d#diff-78de67f5fc597b24f722af79a3a11837R295

joeljfischer:
@t-yoshii Because there is only one audio stream available, if the app does not give it up while it is in the inactive state, I think there will be issues with having multiple navigation apps and switching between them. When the app is in the inactive state, it should use the `Speak` command and the TBT API. In any case, if you wanted to pursue this further, this could not be a bug fix, it would have to be a proposal.

PR #915, opened 2018-03-27
Remove old soft button handlers

### LINKED ISSUE #515: Soft Buttons handlers are never removed
### Bug Report
We store handlers for button events, including `SoftButton`s. When these handlers are no longer relevant, they should be removed from memory in the iOS library. Currently, however, it appears there is no mechanism in place to do so.

Old `SoftButton`s are removed from use when a new `Show` message is sent with `SoftButton`s. When a new `Show` message that contains `SoftButton`s is received, we should remove all old `SoftButton` handlers.

##### Reproduction Steps
1. Connect an app
1. Send a show with `SoftButton`s
1. Send a show with new `SoftButton`s

##### Expected Behavior
The old `SoftButton` handlers are removed from memory.

##### Observed Behavior
The old `SoftButton` handlers remain in memory.

* iOS Version: n/a
* SDL iOS Version: 4.5.2
* Testing Against: n/a

SatbirTanda:
Fixes #515 

This PR is ready for review.

### Risk
This PR makes no API changes.

### Testing Plan
Tests added

### Summary
Removes the button handlers stored for previous SDLShow Requests

### CLA

### DIFF
--- a/SmartDeviceLink/SDLResponseDispatcher.m
+++ b/SmartDeviceLink/SDLResponseDispatcher.m
@@ -112,8 +112,10 @@ NS_ASSUME_NONNULL_BEGIN
         SDLScrollableMessage *scrollableMessage = (SDLScrollableMessage *)request;
         [self sdl_addToCustomButtonHandlerMap:scrollableMessage.softButtons];
     } else if ([request isKindOfClass:[SDLShow class]]) {
-        [self sdl_removeOldButtonHandlers];
         SDLShow *show = (SDLShow *)request;
+        if (show.softButtons.count != 0) {
+            [self sdl_removeOldButtonHandlers];
+        }
         [self sdl_addToCustomButtonHandlerMap:show.softButtons];
     } else if ([request isKindOfClass:[SDLPerformAudioPassThru class]]) {
         SDLPerformAudioPassThru *performAudioPassThru = (SDLPerformAudioPassThru *)request;
--- a/SmartDeviceLink/SDLResponseDispatcher.m
+++ b/SmartDeviceLink/SDLResponseDispatcher.m
@@ -106,16 +106,16 @@ NS_ASSUME_NONNULL_BEGIN
             self.buttonHandlerMap[buttonName] = subscribeButton.handler;
         }
     } else if ([request isKindOfClass:[SDLAlert class]]) {
+        [self sdl_removeOldButtonHandlersForAlerts];
         SDLAlert *alert = (SDLAlert *)request;
-        [self sdl_removeOldButtonHandlersFor:alert];
         [self sdl_addToCustomButtonHandlerMap:alert.softButtons];
     } else if ([request isKindOfClass:[SDLScrollableMessage class]]) {
+        [self sdl_removeOldButtonHandlersForScrollableMessages];
         SDLScrollableMessage *scrollableMessage = (SDLScrollableMessage *)request;
-        [self sdl_removeOldButtonHandlersFor:scrollableMessage];
         [self sdl_addToCustomButtonHandlerMap:scrollableMessage.softButtons];
     } else if ([request isKindOfClass:[SDLShow class]]) {
         SDLShow *show = (SDLShow *)request;
-        [self sdl_removeOldButtonHandlersFor:show];
+        [self sdl_removeOldButtonHandlersForShow:show];
         [self sdl_addToCustomButtonHandlerMap:show.softButtons];
     } else if ([request isKindOfClass:[SDLPerformAudioPassThru class]]) {
         SDLPerformAudioPassThru *performAudioPassThru = (SDLPerformAudioPassThru *)request;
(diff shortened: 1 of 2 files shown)


SmartDeviceLink/SDLResponseDispatcher.m
```diff
@@ -140,6 +141,17 @@ - (void)clear {
     _audioPassThruHandler = nil;
 }
 
+- (void)sdl_removeOldButtonHandlers {
+    for(SDLRPCCorrelationId *key in self.rpcRequestDictionary.allKeys) {
```
joeljfischer: This appears to remove all old buttons on every new `Show` request, but this will cause very bad behavior. For example:

* Show 1
text1, text2, 3 soft buttons

* Show 2
update text 1

Now all 3 soft buttons no longer work even though they're still displayed on the screen.

SmartDeviceLink/SDLResponseDispatcher.m
```diff
@@ -112,6 +112,7 @@ - (void)storeRequest:(SDLRPCRequest *)request handler:(nullable SDLResponseHandl
         SDLScrollableMessage *scrollableMessage = (SDLScrollableMessage *)request;
         [self sdl_addToCustomButtonHandlerMap:scrollableMessage.softButtons];
     } else if ([request isKindOfClass:[SDLShow class]]) {
+        [self sdl_removeOldButtonHandlers];
```
joeljfischer: Remember that `Show`s are not the only place with soft buttons, Alerts and ScrollableMessages do as well.

SmartDeviceLink/SDLResponseDispatcher.m
```diff
@@ -113,6 +113,9 @@ - (void)storeRequest:(SDLRPCRequest *)request handler:(nullable SDLResponseHandl
         [self sdl_addToCustomButtonHandlerMap:scrollableMessage.softButtons];
     } else if ([request isKindOfClass:[SDLShow class]]) {
         SDLShow *show = (SDLShow *)request;
+        if (show.softButtons.count != 0) {
```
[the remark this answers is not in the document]
SatbirTanda: When you say button 1 is updated, how does a button update occur?  Usually when you send a show request it replaces all buttons?

SmartDeviceLink/SDLResponseDispatcher.m
```diff
@@ -140,6 +143,43 @@ - (void)clear {
     _audioPassThruHandler = nil;
 }
 
+- (void)sdl_removeOldButtonHandlersFor:(SDLRPCRequest *)request {
+    for(SDLRPCCorrelationId *key in self.rpcRequestDictionary.allKeys) {
+        if ([self.rpcRequestDictionary[key] isKindOfClass:[SDLAlert class]] && [request isKindOfClass:[SDLAlert class]]) {
+            SDLAlert *oldAlert = (SDLAlert *)self.rpcRequestDictionary[key];
+            SDLAlert *newAlert = (SDLAlert *)request;
+            if(newAlert.softButtons.count != 0) {
```
joeljfischer: If a new alert or scrollable message, come in, we don't need to check if the new alert has soft buttons. Because only one can be active at a time, we know we can simply clear the old ones.

Please format your braces consistently with an additional space as well.

SmartDeviceLink/SDLResponseDispatcher.m
```diff
@@ -140,6 +143,43 @@ - (void)clear {
     _audioPassThruHandler = nil;
 }
 
+- (void)sdl_removeOldButtonHandlersFor:(SDLRPCRequest *)request {
+    for(SDLRPCCorrelationId *key in self.rpcRequestDictionary.allKeys) {
+        if ([self.rpcRequestDictionary[key] isKindOfClass:[SDLAlert class]] && [request isKindOfClass:[SDLAlert class]]) {
```
joeljfischer: We already know the request class type based on where this came from, ideally, we don't want to re-check it possibly hundreds of times in this loop. Instead of one big method, there should be smaller `sdl_removeButtonHandlersFor:(SDLAlert *)alert` (and such for the others') as well.

SmartDeviceLinkTests/DevAPISpecs/SDLResponseDispatcherSpec.m
```diff
@@ -201,6 +205,31 @@
             });
         });
         
+        context(@"with an old show request", ^{
```
joeljfischer: Additional tests should be done for alerts / scrollable messages

SmartDeviceLink/SDLResponseDispatcher.m
```diff
@@ -140,6 +143,43 @@ - (void)clear {
...
+                    }
+                }
+            }
+        } else if([self.rpcRequestDictionary[key] isKindOfClass:[SDLShow class]] && [request isKindOfClass:[SDLShow class]]) {
+            SDLShow *oldShow = (SDLShow *)self.rpcRequestDictionary[key];
+            SDLShow *newShow = (SDLShow *)request;
+            if(newShow.softButtons.count != 0) {
+                for (SDLSoftButton *softButton in oldShow.softButtons) {
+                    // Check if new show's softButton array contains ID of old soft button. (This means button is being updated).
+                    if (![[newShow.softButtons valueForKey:@"softButtonID"] containsObject:softButton.softButtonID]) {
```
joeljfischer: I think you were correct and I was mistaken that a button could be updated. I tested, and it just replaces all the buttons. This should be unnecessary.

Issue #900, opened 2018-03-16
[SDL 0141] Supporting simultaneous multiple transports


theresalech:
*Note that separate discussions will need to take place to confirm proxy implementation details for this proposal.*

Proposal: [Supporting simultaneous multiple transports](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0141-multiple-transports.md)

> This proposal aims to support multiple transports between Core and Proxy.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/405

Steering Committee Decision:

>The Steering Committee has voted to accept this proposal with revisions. The revisions are as follows:

- Introduce a new Control Frame, `RegisterSecondaryTransport`, which will be used once a Secondary Transport is connected to notify Core that the Proxy has established a connection and tell Core which session is registering.
- Rename `Transport Config Update` to `TransportEventUpdate`. The original parameters will be refactored to a flat structure.
- StartServiceACK for the RPC service will contain a Secondary Transport param that will contain an array of potential secondary transports instead of a single value. The service to transport priority parameters will become flat instead of document style; each service param will contain an ordered array in terms of priority of potential transports.
- Proxy connection retry strategy will include the new `TransportEventUpdate` Control Frame. When a connection is interrupted/ended, a `TransportEventUpdate` control frame will be sent to the proxy on the primary transport. If the proxy is to retry (in case of WiFi), it will include the IP address and Port. If the proxy is not to retry, the frame will contain either empty strings for IP address and port or not include the parameters at all.
- Will recommend, but not require, using 5GHz
- Proxy implementations will be discussed outside of this proposal.

>The proposal .md was updated on March 16, 2018 to include the agreed upon revisions.

theresalech:
Closed via #947

PR #947, opened 2018-05-01
Multiple simultaneous transports

shoamano83:
Fixes https://github.com/smartdevicelink/sdl_ios/issues/900

### Risk
(Addition of internal methods in SDLStreamingMediaManager.h.)

### Testing Plan
Unit tests are included in this PR.

### Summary
This PR includes necessary changes for multiple-transports feature.

##### Breaking Changes
* SDLStreamingMediaLifecycleManager is split into `SDLStreamingMediaVideoLifecycleManager` and `SDLStreamingMediaAudioLifecycleManager` as the streamings have to be controlled separately.
* Protocol version is bumped to 5.1.0.

##### Enhancements
* Addition of `SecondaryTransportManager`.
* Support of new control frames as well as new parameters in BSON.
* Addition of protection against `protocolDelegateTable` in SDLProtocol.

##### Bug Fixes
* Minor test case fixes in SDLStreamingMediaAudioLifecycleManagerSpec and SDLProtocolSpec.

### Tasks Remaining:
- None

### CLA


shoamano83:
I have rebased the commits onto latest `develop` branch, resolving conflicts caused by recent major version update.

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -203,16 +206,26 @@ - (void)didEnterStateStarted {
...
+                          secondaryTransportManager:self.secondaryTransportManager];
     }
 #pragma clang diagnostic pop
 
-    if (self.streamManager != nil) {
-        [self.streamManager startWithProtocol:self.proxy.protocol];
+    // if secondary transport manager is used, streaming media manager will be started through
+    // onAudioServiceProtocolUpdated and onVideoServiceProtocolUpdated
+    if (self.secondaryTransportManager == nil) {
+        if (self.streamManager != nil) {
```
joeljfischer: Can this be streamlined into the above `if`?

SmartDeviceLink/SDLStreamingProtocolListener.h
```diff
@@ -0,0 +1,41 @@
...
+/**
+ *  Called when protocol instance for audio service has been updated.
+ *
+ *  If `newProtocol` is nil, it indicates that underlying transport
+ *  becomes unavailable.
+ *
+ *  @param oldProtocol protocol instance that has been used for audio streaming.
+ *  @param newProtocol protocol instance that will be used for audio streaming.
+ */
+- (void)onAudioServiceProtocolUpdated:(nullable SDLProtocol *)oldProtocol to:(nullable SDLProtocol *)newProtocol;
```
joeljfischer: Please rename to `audioServiceProtocolDidUpdateFromOldProtocol:toNewProtocol:`

SmartDeviceLink/SDLStreamingProtocolListener.h
```diff
@@ -0,0 +1,41 @@
...
+/**
+ *  Called when protocol instance for video service has been updated.
+ *
+ *  If `newProtocol` is nil, it indicates that underlying transport
+ *  becomes unavailable.
+ *
+ *  @param oldProtocol protocol instance that has been used for video streaming.
+ *  @param newProtocol protocol instance that will be used for video streaming.
+ */
+- (void)onVideoServiceProtocolUpdated:(nullable SDLProtocol *)oldProtocol to:(nullable SDLProtocol *)newProtocol;
```
joeljfischer: Please rename to `audioServiceProtocolDidUpdateFromOldProtocol:toNewProtocol:`

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+
+/// Timeout for receiving Register Secondary Transport ACK frame
+static const float RegisterTransportTime = 10.0;
+
+/// Delay to retry connection after secondary transport is disconnected or registration is failed
+static const float RetryConnectionDelay = 15.0;
+
+
+/// An internal class to receive event from primary transport.
+@interface PrimaryProtocolListener : NSObject <SDLProtocolListener>
```
joeljfischer: Every class should have its own file (outside of perhaps extraordinarily small, struct-like ones on a case-by-case basis).

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+- (void)sdl_handleTransportRegistered {
+    // secondary transport is now ready
+    [self.stateMachine transitionToState:SDLSecondaryTransportStateRegistered];
+}
+
+- (void)sdl_handleTransportRegisterFailed {
+    [self.stateMachine transitionToState:SDLSecondaryTransportStateReconnecting];
+}
+
+- (void)sdl_handleTransportRegisterTimeout {
```
joeljfischer: This one-liner is only called from one place, just move the code there?

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    // if the state is still Connecting, go back to Configured state and retry immediately
+    if ([self.stateMachine isCurrentState:SDLSecondaryTransportStateConnecting]) {
+        SDLLogD(@"Retry secondary transport connection after registration timeout");
+        [self.stateMachine transitionToState:SDLSecondaryTransportStateConfigured];
+    }
+}
+
+- (void)sdl_handleProtocolClosed {
```
joeljfischer: This one-liner is only called from one place, just move the code there?

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    if ([self sdl_isTransportOpened]) {
+        [self.stateMachine transitionToState:SDLSecondaryTransportStateReconnecting];
+    }
+}
+
+- (void)sdl_handleAppBecomeActive {
```
joeljfischer: This one-liner is only called from one place, just move the code there?

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    if (([self.stateMachine isCurrentState:SDLSecondaryTransportStateConfigured])
+        && self.transportSelection == SDLTransportSelectionTCP && [self sdl_isTCPReady]) {
+        SDLLogD(@"Resuming TCP transport since the app becomes foreground");
+        [self.stateMachine transitionToState:SDLSecondaryTransportStateConnecting];
+    }
+}
+
+- (void)sdl_handleAppResignedActive {
```
joeljfischer: This one-liner is only called from one place, just move the code there?

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    if ([self sdl_isTransportOpened] && self.transportSelection == SDLTransportSelectionTCP) {
+        SDLLogD(@"Disconnecting TCP transport since the app will go to background");
+        [self.stateMachine transitionToState:SDLSecondaryTransportStateConfigured];
+    }
+}
+
+- (BOOL)sdl_isTransportOpened {
+    return [self.stateMachine isCurrentState:SDLSecondaryTransportStateConnecting] ||
```
joeljfischer: Move `||` down to the next line

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+          [self.streamingProtocolListener onVideoServiceProtocolUpdated:oldProtocol to:newProtocol];
+      }];
+}
+
+- (void)sdl_updateService:(UInt8)service
+        allowedTransports:(nullable NSArray<SDLTransportClassBox *> *)transportList
+         primaryAvailable:(BOOL)primaryTransportAvailable
+       secondaryAvailable:(BOOL)secondaryTransportAvailable
+    transportUpdatedBlock:(void (^)(SDLProtocol * _Nullable oldProtocol, SDLProtocol * _Nullable newProtocol))transportUpdatedBlock {
+
```
joeljfischer: Extra line

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+        }
+
+        transportUpdatedBlock([self sdl_getProtocolFromTransportClass:oldTransport],
+                              [self sdl_getProtocolFromTransportClass:newTransport]);
+    }
+}
+
+- (nullable SDLProtocol *)sdl_getProtocolFromTransportClass:(SDLTransportClass)transportClass {
+    switch (transportClass) {
+        case SDLTransportClassPrimary:   return self.primaryProtocol;
```
joeljfischer: Don't align the `return` sections, just put them one space after the `:`

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+        case SDLTransportClassSecondary: return self.secondaryProtocol;
+        default:                         return nil;
+    }
+}
+
+- (nullable NSString *)sdl_getTransportClassName:(SDLTransportClass)transportClass {
+    switch (transportClass) {
+        case SDLTransportClassPrimary:   return @"primary";
```
joeljfischer: Don't align the `return` sections, just put them one space after the `:`

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    return [array copy];
+}
+
+- (SDLTransportSelection)sdl_getTransportSelectionFromProtocol:(SDLProtocol *)protocol {
+    if ([protocol.transport isMemberOfClass:[SDLIAPTransport class]]) {
+        return SDLTransportSelectionIAP;
+    } else if ([protocol.transport isMemberOfClass:[SDLTCPTransport class]]) {
+        return SDLTransportSelectionTCP;
+    } else {
+        SDLLogD(@"Unknown type of primary transport");
```
joeljfischer: Should be an error log

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    // Actually there is nothing to "convert" here. We just check the range and recreate an array.
+    NSMutableArray<SDLTransportClassBox *> *array = [NSMutableArray new];
+    for (NSNumber *num in transports) {
+        int transport = [num intValue];
+        if (transport == 1) {
+            [array addObject:@(SDLTransportClassPrimary)];
+        } else if (transport == 2) {
+            [array addObject:@(SDLTransportClassSecondary)];
+        } else {
+            SDLLogV(@"Unknown transport class received: %d", transport);
```
joeljfischer: Should be an error log

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    });
+}
+
+
+#pragma mark - Transport management
+
+// try establishing secondary transport. Returns NO if failed
+- (BOOL)sdl_connectSecondaryTransport {
+    if (self.secondaryTransport != nil) {
+        SDLLogW(@"Secondary transport is already started.");
```
joeljfischer: This isn't enough detail. Maybe "Attempting to connect secondary transport, but it's already started"

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+            return [self sdl_startIAPSecondaryTransport];
+        default:
+            SDLLogW(@"Unknown transport type for secondary transport: %ld", (long)self.transportSelection);
+            return NO;
+    }
+}
+
+- (BOOL)sdl_disconnectSecondaryTransport {
+    if (self.secondaryTransport == nil) {
+        SDLLogD(@"Secondary transport is already stopped.");
```
joeljfischer: Warning log, should have more detail "Attempted to disconnect secondary transport, but it's already stopped"

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    SDLLogD(@"Disconnect secondary transport");
+    [self.secondaryTransport disconnect];
+    self.secondaryTransport = nil;
+
+    return YES;
+}
+
+- (BOOL)sdl_startTCPSecondaryTransport {
+    if (![self sdl_isTCPReady]) {
+        SDLLogD(@"Cannot start TCP secondary transport");
```
joeljfischer: Should this be an error log? Or at least a warning log?

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    // this is for video and audio services
+    [self.secondaryProtocol storeHeader:self.primaryProtocolListener.primaryRPCHeader forServiceType:SDLServiceTypeRPC];
+
+    [self.secondaryTransport connect];
+    return YES;
+}
+
+- (BOOL)sdl_isTCPReady {
+    if ([self.ipAddress length] == 0) {
+        SDLLogD(@"IP address is empty");
```
joeljfischer: No logs are needed in this method

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+        return appState;
+    }
+}
+
+
+#pragma mark - SDLProtocolListener Implementation
+
+// called on transport's thread, notifying that the transport is established
+- (void)onProtocolOpened {
+    SDLLogD(@"secondary transport connected");
```
joeljfischer: Capitalization

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: I noticed your another comment https://github.com/smartdevicelink/sdl_ios/pull/947#discussion_r205576128 after I did this change. Let me think about it again.

SmartDeviceLink/SDLLifecycleManager.m
shoamano83: Sure, done in commit da0b0fcf5b240b479d4360de430ee5ac9681e9d6.

SmartDeviceLink/SDLSecondaryTransportManager.h
```diff
@@ -0,0 +1,62 @@
...
+ Secondary transport manager does most of the things required for the feature. It listens
+ on the primary transport and receives necessary information through Version Negotiation
+ and TransportEventUpdate control frame. It initiates secondary transport's connection
+ and sets up SDLProtocol that runs on the transport. Then it starts streaming media
+ manager with appropriate SDLProtocol instance. When the secondary transport is
+ disconnected, this manager retries connection with a regular interval.
+ */
+@interface SDLSecondaryTransportManager : NSObject <SDLProtocolListener>
+
+/// state of this manager
```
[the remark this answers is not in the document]
shoamano83: My bad, the three-slashes comment should be in two or more lines. I fixed all of them in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6.

SmartDeviceLink/SDLSecondaryTransportManager.h
```diff
@@ -0,0 +1,62 @@
...
+ @return A new secondary transport manager
+ */
+- (instancetype)initWithStreamingProtocolListener:(id<SDLStreamingProtocolListener>)streamingProtocolListener;
+
+/**
+ *  Start the manager.
+
+ @param primaryProtocol protocol that runs on the main (primary) transport
+ */
+- (void)startWithProtocol:(SDLProtocol *)primaryProtocol;
```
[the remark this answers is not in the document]
shoamano83: Fixed in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+}
+
+@end
+
+@interface SDLSecondaryTransportManager ()
+
+/// state of this manager
+@property (strong, nonatomic, readwrite) SDLStateMachine *stateMachine;
+/// dedicated queue that the state machine will run on
+@property (copy, nonatomic) dispatch_queue_t stateMachineQueue;
```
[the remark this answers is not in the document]
shoamano83: I agree. What is the appropriate queue we should use here? I'm pretty sure that we don't want to use the main queue (main thread). Or do you mean that we don't need a dedicated queue for this state machine at all?

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+typedef NSNumber SDLServiceTypeBox;
+
+typedef NS_ENUM(NSInteger, SDLTransportClass) {
+    SDLTransportClassInvalid = 0,
+    SDLTransportClassPrimary = 1,
+    SDLTransportClassSecondary = 2,
+};
+typedef NSNumber SDLTransportClassBox;
+
+typedef NS_ENUM(NSInteger, SDLTransportSelection) {
```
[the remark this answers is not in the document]
shoamano83: Fixed in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+@interface SDLSecondaryTransportManager ()
+
+/// state of this manager
+@property (strong, nonatomic, readwrite) SDLStateMachine *stateMachine;
+/// dedicated queue that the state machine will run on
+@property (copy, nonatomic) dispatch_queue_t stateMachineQueue;
+
+/// instance of the protocol that runs on primary transport
+@property (weak, nonatomic) SDLProtocol *primaryProtocol;
+/// a sub-class to catch Start Service ACK and Transport Config Update frames
```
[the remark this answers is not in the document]
shoamano83: Should be fixed in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6.

SmartDeviceLink/SDLStreamingProtocolListener.h
```diff
@@ -0,0 +1,41 @@
...
+//  Copyright © 2018 Xevo Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@class SDLProtocol;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol SDLStreamingProtocolListener <NSObject>
```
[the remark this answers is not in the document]
shoamano83: OK, the protocol name as well as file name and related variable names are updated in commit 7ddb37ac76a529ee06c21ca6f6ad3348a4d3919a.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+
+// a listener to notify protocol change
+@property (weak, nonatomic) id<SDLStreamingProtocolListener> streamingProtocolListener;
+
+/// Configuration sent by system; list of transports that are allowed to carry audio service
+@property (strong, nonatomic, nullable) NSArray<SDLTransportClassBox *> *transportsForAudioService;
+/// Configuration sent by system; list of transports that are allowed to carry video service
+@property (strong, nonatomic, nullable) NSArray<SDLTransportClassBox *> *transportsForVideoService;
+/// A map to remember which service is currently running on which transport
+@property (strong, nonatomic) NSMutableDictionary<SDLServiceTypeBox *, SDLTransportClassBox *> *assignedTransport;
```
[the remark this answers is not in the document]
shoamano83: Fixed in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+
+- (BOOL)sdl_isTransportOpened {
+    return [self.stateMachine isCurrentState:SDLSecondaryTransportStateConnecting] ||
+           [self.stateMachine isCurrentState:SDLSecondaryTransportStateRegistered];
+}
+
+
+#pragma mark - Starting / Stopping / Restarting services
+
+- (void)sdl_handleTransportUpdate:(BOOL)primaryAvailable secondaryAvailable:(BOOL)secondaryAvailable {
```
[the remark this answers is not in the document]
shoamano83: Fixed in commit af0f0c9f8cad79373944a732e2f0298c608d1995.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    [self sdl_stopManager];
+}
+
+- (void)didEnterStateStarted {
+    [self sdl_startManager];
+}
+
+- (void)didEnterStateConfigured {
+    if ((self.transportSelection == SDLTransportSelectionTCP && [self sdl_isTCPReady]) ||
+        self.transportSelection == SDLTransportSelectionIAP) {
```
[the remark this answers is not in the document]
shoamano83: Fixed in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+- (void)sdl_configureManager:(nullable NSArray<SDLTransportSelectionBox *> *)secondaryTransports
+          transportsForAudio:(nullable NSArray<SDLTransportClassBox *> *)transportsForAudio
+          transportsForVideo:(nullable NSArray<SDLTransportClassBox *> *)transportsForVideo {
+
+    if (![self.stateMachine isCurrentState:SDLSecondaryTransportStateStarted]) {
+        SDLLogW(@"SecondaryTransportManager ignores duplicate Start Service ACK frame");
+        return;
+    }
+
+    SDLTransportSelection secondaryTransportSelection;
```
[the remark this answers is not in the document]
shoamano83: I added `SDLSecondaryTransportTypeDisabled` instead in commit da0b0fcf5b240b479d4360de430ee5ac9681e9d6.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    self.transportsForVideoService = nil;
+
+    self.ipAddress = nil;
+    self.tcpPort = -1;
+}
+
+- (void)sdl_configureManager:(nullable NSArray<SDLTransportSelectionBox *> *)secondaryTransports
+          transportsForAudio:(nullable NSArray<SDLTransportClassBox *> *)transportsForAudio
+          transportsForVideo:(nullable NSArray<SDLTransportClassBox *> *)transportsForVideo {
+
```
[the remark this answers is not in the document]
shoamano83: Fixed in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+        return;
+    }
+
+    SDLTransportSelection secondaryTransportSelection;
+    if (secondaryTransports == nil || [secondaryTransports count] == 0) {
+        SDLLogW(@"Did not receive secondary transport type from system. Secondary transport is disabled.");
+        secondaryTransportSelection = SDLTransportSelectionDisabled;
+    } else {
+        // current proposal says the list should contain only one element
+        SDLTransportSelectionBox *transportSelection = [secondaryTransports objectAtIndex:0];
```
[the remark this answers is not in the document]
shoamano83: Fixed in commit da0b0fcf5b240b479d4360de430ee5ac9681e9d6.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+          transportsForAudio:(nullable NSArray<SDLTransportClassBox *> *)transportsForAudio
+          transportsForVideo:(nullable NSArray<SDLTransportClassBox *> *)transportsForVideo {
+
+    if (![self.stateMachine isCurrentState:SDLSecondaryTransportStateStarted]) {
+        SDLLogW(@"SecondaryTransportManager ignores duplicate Start Service ACK frame");
+        return;
+    }
+
+    SDLTransportSelection secondaryTransportSelection;
+    if (secondaryTransports == nil || [secondaryTransports count] == 0) {
```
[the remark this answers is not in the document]
shoamano83: Fixed in commit da0b0fcf5b240b479d4360de430ee5ac9681e9d6.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Fixed in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6. Let me know if you want to align `||` after `return`.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Fixed in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: OK, fixed in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Fixed in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Fixed in commit c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+/// dedicated queue that the state machine will run on
+@property (copy, nonatomic) dispatch_queue_t stateMachineQueue;
+
+/// instance of the protocol that runs on primary transport
+@property (weak, nonatomic) SDLProtocol *primaryProtocol;
+/// a sub-class to catch Start Service ACK and Transport Config Update frames
+@property (strong, nonatomic) PrimaryProtocolListener *primaryProtocolListener;
+
+/// Selected type of secondary transport. If 'SDLTransportSelectionDisabled' then secondary transport is disabled.
+@property (assign, nonatomic) SDLTransportSelection transportSelection;
```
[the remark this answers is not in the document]
shoamano83: Fixed with commits c6d6d98423e3372a89ad1f52331bd5f1b8c65fd6 and ba8d81cb81d8b50e3c222d75ea14e6e4c36dd15c.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+        // stop all services, including those running on primary transport
+        SDLLogD(@"Stopping audio / video services on both transports");
+        [self sdl_handleTransportUpdate:NO secondaryAvailable:NO];
+
+        [self.stateMachine transitionToState:SDLSecondaryTransportStateStopped];
+    });
+}
+
+- (void)dealloc {
+    SDLLogD(@"SDLSecondaryTransportManager dealloc");
```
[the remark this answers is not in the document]
shoamano83: Fixed in commit b734d113d4918e8fe6af7ed6c303c017d0582083.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+                                @(SDLServiceTypeVideo):@(SDLTransportClassInvalid)} mutableCopy];
+    self.transportSelection = SDLTransportSelectionDisabled;
+    self.transportsForAudioService = nil;
+    self.transportsForVideoService = nil;
+
+    self.ipAddress = nil;
+    self.tcpPort = -1;
+}
+
+- (void)sdl_configureManager:(nullable NSArray<SDLTransportSelectionBox *> *)secondaryTransports
```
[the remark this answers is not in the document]
shoamano83: Updated in commit 3c8e2775df4f5fa17914014207b89fce467921fd.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+          transportsForAudio:(nullable NSArray<SDLTransportClassBox *> *)transportsForAudio
```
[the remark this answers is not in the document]
shoamano83: Updated in commit 3c8e2775df4f5fa17914014207b89fce467921fd.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+        secondaryTransportSelection = SDLTransportSelectionDisabled;
+    } else {
+        // current proposal says the list should contain only one element
+        SDLTransportSelectionBox *transportSelection = [secondaryTransports objectAtIndex:0];
+        secondaryTransportSelection = [transportSelection integerValue];
+    }
+
+    SDLTransportSelection primaryTransportSelection = [self sdl_getTransportSelectionFromProtocol:self.primaryProtocol];
+    if (primaryTransportSelection == secondaryTransportSelection) {
+        SDLLogW(@"Same transport is specified for both primary and secondary transport. Secondary transport is disabled.");
```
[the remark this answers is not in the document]
shoamano83: This happens when Core is wrongly configured. I agree it's not a warning to app developers, but I also want it a non-debug message. If we had a INFO level log, that would be helpful.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+}
+
+- (void)sdl_handleTransportEventUpdate {
+    if ([self.stateMachine isCurrentState:SDLSecondaryTransportStateStarted]) {
+        // The system sent Transport Event Update frame prior to Start Service ACK! Just keep the information and do nothing here.
+        SDLLogD(@"Received TCP transport information prior to Start Service ACK");
+        return;
+    }
+    if (self.transportSelection != SDLTransportSelectionTCP) {
+        return;
```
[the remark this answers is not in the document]
shoamano83: Done in commit 5009bc7f0137147a9211bbf3725e2dee7b5cba98.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    // this will trigger audio / video streaming if they are allowed on primary transport
+    [self sdl_handleTransportUpdate:YES secondaryAvailable:NO];
+
+    [self.stateMachine transitionToState:SDLSecondaryTransportStateConfigured];
+}
+
+- (void)sdl_handleTransportEventUpdate {
+    if ([self.stateMachine isCurrentState:SDLSecondaryTransportStateStarted]) {
+        // The system sent Transport Event Update frame prior to Start Service ACK! Just keep the information and do nothing here.
+        SDLLogD(@"Received TCP transport information prior to Start Service ACK");
```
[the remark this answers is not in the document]
shoamano83: This case is acceptable and thus I think it's not a warning. I can tweak the level if you still prefer.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+        // Disconnect current transport. If the IP address is available then we will reconnect immediately.
+        SDLLogD(@"TCP transport information updated, disconnecting current transport");
+        [self.stateMachine transitionToState:SDLSecondaryTransportStateConfigured];
+    } else if ([self.stateMachine isCurrentState:SDLSecondaryTransportStateReconnecting]) {
+        SDLLogD(@"TCP transport information updated, aborting reconnection timer");
+        [self.stateMachine transitionToState:SDLSecondaryTransportStateConfigured];
+    }
+}
+
+- (void)sdl_handleTransportRegistered {
```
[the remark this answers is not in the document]
shoamano83: Updated in commit b734d113d4918e8fe6af7ed6c303c017d0582083.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+    // secondary transport is now ready
+    [self.stateMachine transitionToState:SDLSecondaryTransportStateRegistered];
+}
+
+- (void)sdl_handleTransportRegisterFailed {
```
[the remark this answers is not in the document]
shoamano83: Updated in commit b734d113d4918e8fe6af7ed6c303c017d0582083.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Updated in commit b734d113d4918e8fe6af7ed6c303c017d0582083.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Updated in commit b734d113d4918e8fe6af7ed6c303c017d0582083.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Updated in commit b734d113d4918e8fe6af7ed6c303c017d0582083.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Updated in commit b734d113d4918e8fe6af7ed6c303c017d0582083.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Fixed in commit 5009bc7f0137147a9211bbf3725e2dee7b5cba98.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Fixed in commit 5009bc7f0137147a9211bbf3725e2dee7b5cba98.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Thanks, updated in commit 5009bc7f0137147a9211bbf3725e2dee7b5cba98.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Updated in commit 5009bc7f0137147a9211bbf3725e2dee7b5cba98.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: IMHO this is not an error case, so I hesitate to output a warning log. (For example, Wi-Fi of the head unit isn't available, or the SDL app is not in foreground.) Again, if we had a INFO level log that would be more suitable.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Hmm, `sdl_isTCPReady` can return `NO` for various reasons. So I thought debug logs would be helpful for developers. I can remove them, but I prefer to keep.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: The class is once again split into dedicated file in commit f21418021582633997b7684b0ce128c88dfe3cac.

SmartDeviceLink/SDLProtocol.m
```diff
@@ -87,15 +97,23 @@ - (UInt8)sdl_retrieveSessionIDforServiceType:(SDLServiceType)serviceType {
 #pragma mark - SDLTransportDelegate
 
 - (void)onTransportConnected {
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners;
+    @synchronized(self.protocolDelegateTable) {
```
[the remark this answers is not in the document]
shoamano83: Correct me if I'm wrong, but I think `NSHashTable` isn't thread-safe. That is, if we create a new array using `allObjects` and another thread updates the table at the same time, then the array will be corrupted. The lock is added to prevent such multi-threading issue. I know such issue is very unlikely to happen though.

In Java, we have a convenient class called `ConcurrentHashMap`.

SmartDeviceLink/SDLProtocol.m
```diff
@@ -507,66 +566,98 @@ - (void)handleHeartbeatForSession:(Byte)session {
...
+            SDLLogE(@"Register Secondary Transport NAK'd, reason: %@", payload.reason);
+        } break;
+        default:
+            break;
+    }
+}
+
+- (NSArray<id<SDLProtocolListener>> *)sdl_getProtocolListeners {
+    @synchronized(self.protocolDelegateTable) {
+        return self.protocolDelegateTable.allObjects;
```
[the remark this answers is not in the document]
shoamano83: Please refer to the comment above.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+SDLSecondaryTransportState *const SDLSecondaryTransportStateConfigured = @"Configured";
+SDLSecondaryTransportState *const SDLSecondaryTransportStateConnecting = @"Connecting";
+SDLSecondaryTransportState *const SDLSecondaryTransportStateRegistered = @"Registered";
+SDLSecondaryTransportState *const SDLSecondaryTransportStateReconnecting = @"Reconnecting";
+
+/// Timeout for receiving Register Secondary Transport ACK frame
+static const float RegisterTransportTime = 10.0;
+
+/// Delay to retry connection after secondary transport is disconnected or registration is failed
+static const float RetryConnectionDelay = 15.0;
```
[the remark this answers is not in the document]
shoamano83: I updated to 5 seconds in commit 114f9c69f68b0255331236585a338e49069dd62e. I am not still sure for the best value though.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+
+/// An internal class to receive event from primary transport.
+@interface PrimaryProtocolListener : NSObject <SDLProtocolListener>
+
+@property (weak, nonatomic) SDLProtocol *primaryProtocol;
+@property (copy, nonatomic) SDLProtocolHeader *primaryRPCHeader;
+
+@end
+
+@implementation PrimaryProtocolListener
```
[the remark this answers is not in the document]
shoamano83: The reason that I created this class was that secondary transport manager has to receive events from two `SDLProtocol`s (primary and secondary) through `SDLProtocolListener`. If we receive all events in `SDLSecondaryTransportManager`, the events will be mixed up and I cannot distinguish whether an event comes from primary transport or secondary transport.

Another option was to extend `SDLProtocolListener` protocol so that every callback includes an instance of SDLProtocol. This seemed to require much more changes so I didn't choose it. (Also, `SDLProtocol` was public at that time.)

Based on your suggestion, I created dedicated files for the class, renaming it to `SDLPrimaryProtocolDelegate`. The notifications are replaced with method calls. Please let me know if you have better idea.

SmartDeviceLink/SDLSecondaryTransportManager.m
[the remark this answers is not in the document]
shoamano83: Just for follow-up. IMHO a state machine only makes sense if it runs on a single thread. Accessing a state machine from multiple threads doesn't sound like a good idea, as two threads may verify and update the state at the same time, causing a corrupted result.

The state machine in secondary transport manager is indeed accessed by multiple threads - the queue of primary protocol, the queue of secondary protocol and the main thread. So I thought I have three options:

1) create a dedicated queue (thread) and run the state machine on it
2) reuse a queue (thread) from somewhere and run the state machine on it
3) Add locks everywhere in the state machine

I choose option 1 for my current implementation, but If we have a queue that can be reused, option 2 seems better. I don't like option 3.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,806 @@
...
+/// timer to check Register Secondary Transport ACK response on secondary transport
+@property (strong, nonatomic, nullable) SDLTimer *registerTransportTimer;
+
+// a listener to notify protocol change
+@property (weak, nonatomic) id<SDLStreamingProtocolListener> streamingProtocolListener;
+
+/// Configuration sent by system; list of transports that are allowed to carry audio service
+@property (strong, nonatomic, nullable) NSArray<SDLTransportClassBox *> *transportsForAudioService;
+/// Configuration sent by system; list of transports that are allowed to carry video service
+@property (strong, nonatomic, nullable) NSArray<SDLTransportClassBox *> *transportsForVideoService;
```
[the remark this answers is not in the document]
shoamano83: Hmm I prefer current approach. When Core doesn't send the list, the array is null. When it sends an empty list, the array is non-null and is empty.
I could make the array to be always non-null but I think that would cause more confusion.

shoamano83:
Hi @joeljfischer , I think I went through all of your comments now. Could you kindly go through them, as I have a few questions in some of the comments? Let me know if this PR still needs some more works. Thanks!

SmartDeviceLink/SDLProtocol.h
```diff
@@ -80,6 +93,13 @@ extern NSString *const SDLProtocolSecurityErrorDomain;
  */
 - (void)endServiceWithType:(SDLServiceType)serviceType;
 
+/**
+ *  Sends a Register Secondary Transport control frame to Core
+ *
+ *  @param payload The data to send in the message
+ */
+- (void)registerSecondaryTransport:(nullable NSData *)payload;
```
joeljfischer: This should be `registerSecondaryTransportWithPayload:`, but this is only ever called with `nil`, so the parameter is unnecessary.

SmartDeviceLink/SDLProtocol.m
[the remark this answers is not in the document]
joeljfischer: `allObjects` will create a copy of the objects. The array will not be corrupted. Unless you have a concrete issue that can be traced back to this, this is simply slowing down the library for no clear benefit.

SmartDeviceLink/SDLProtocol.m
```diff
@@ -87,15 +97,23 @@ - (UInt8)sdl_retrieveSessionIDforServiceType:(SDLServiceType)serviceType {
...
         if ([listener respondsToSelector:@selector(onProtocolOpened)]) {
             [listener onProtocolOpened];
         }
     }
 }
 
 - (void)onTransportDisconnected {
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners;
+    @synchronized(self.protocolDelegateTable) {
```
joeljfischer: Also unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -459,7 +494,8 @@ - (void)handleProtocolStartServiceACKMessage:(SDLProtocolMessage *)startServiceA
     self.serviceHeaders[@(startServiceACK.header.serviceType)] = [startServiceACK.header copy];
 
     // Pass along to all the listeners
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
+    for (id<SDLProtocolListener> listener in listeners) {
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -469,7 +505,8 @@ - (void)handleProtocolStartServiceACKMessage:(SDLProtocolMessage *)startServiceA
 - (void)handleProtocolStartServiceNAKMessage:(SDLProtocolMessage *)startServiceNAK {
     [self sdl_logControlNAKPayload:startServiceNAK];
 
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -480,7 +517,8 @@ - (void)handleProtocolEndServiceACKMessage:(SDLProtocolMessage *)endServiceACK {
     // Remove the session id
     [self.serviceHeaders removeObjectForKey:@(endServiceACK.header.serviceType)];
 
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -490,13 +528,34 @@ - (void)handleProtocolEndServiceACKMessage:(SDLProtocolMessage *)endServiceACK {
 - (void)handleProtocolEndServiceNAKMessage:(SDLProtocolMessage *)endServiceNAK {
     [self sdl_logControlNAKPayload:endServiceNAK];
 
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -507,66 +566,98 @@ - (void)handleHeartbeatForSession:(Byte)session {
     SDLProtocolMessage *message = [SDLProtocolMessage messageWithHeader:header andPayload:nil];
     [self sdl_sendDataToTransport:message.data onService:header.serviceType];
 
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -490,13 +528,34 @@ - (void)handleProtocolEndServiceACKMessage:(SDLProtocolMessage *)endServiceACK {
...
+    for (id<SDLProtocolListener> listener in listeners) {
         if ([listener respondsToSelector:@selector(handleProtocolEndServiceNAKMessage:)]) {
             [listener handleProtocolEndServiceNAKMessage:endServiceNAK];
         }
     }
 }
 
+- (void)handleProtocolRegisterSecondaryTransportACKMessage:(SDLProtocolMessage *)registerSecondaryTransportACK {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
+    for (id<SDLProtocolListener> listener in listeners) {
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -490,13 +528,34 @@ - (void)handleProtocolEndServiceACKMessage:(SDLProtocolMessage *)endServiceACK {
...
+            [listener handleProtocolRegisterSecondaryTransportACKMessage:registerSecondaryTransportACK];
+        }
+    }
+}
+
+- (void)handleProtocolRegisterSecondaryTransportNAKMessage:(SDLProtocolMessage *)registerSecondaryTransportNAK {
+    [self sdl_logControlNAKPayload:registerSecondaryTransportNAK];
+
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
+    for (id<SDLProtocolListener> listener in listeners) {
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -507,66 +566,98 @@ - (void)handleHeartbeatForSession:(Byte)session {
...
         if ([listener respondsToSelector:@selector(handleHeartbeatForSession:)]) {
             [listener handleHeartbeatForSession:session];
         }
     }
 }
 
 - (void)handleHeartbeatACK {
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -507,66 +566,98 @@ - (void)handleHeartbeatForSession:(Byte)session {
...
+    for (id<SDLProtocolListener> listener in listeners) {
         if ([listener respondsToSelector:@selector(handleHeartbeatACK)]) {
             [listener handleHeartbeatACK];
         }
     }
 }
 
+- (void)handleTransportEventUpdateMessage:(SDLProtocolMessage *)transportEventUpdate {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
+    for (id<SDLProtocolListener> listener in listeners) {
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -507,66 +566,98 @@ - (void)handleHeartbeatForSession:(Byte)session {
...
+
 - (void)onProtocolMessageReceived:(SDLProtocolMessage *)msg {
     // Control service (but not control frame type) messages are TLS handshake messages
     if (msg.header.serviceType == SDLServiceTypeControl) {
         [self sdl_processSecurityMessage:msg];
         return;
     }
 
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -507,66 +566,98 @@ - (void)handleHeartbeatForSession:(Byte)session {
...
+    for (id<SDLProtocolListener> listener in listeners) {
         if ([listener respondsToSelector:@selector(onProtocolMessageReceived:)]) {
             [listener onProtocolMessageReceived:msg];
         }
     }
 }
 
 - (void)onProtocolOpened {
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -507,66 +566,98 @@ - (void)handleHeartbeatForSession:(Byte)session {
...
+    for (id<SDLProtocolListener> listener in listeners) {
         if ([listener respondsToSelector:@selector(onProtocolOpened)]) {
             [listener onProtocolOpened];
         }
     }
 }
 
 - (void)onProtocolClosed {
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
```diff
@@ -507,66 +566,98 @@ - (void)handleHeartbeatForSession:(Byte)session {
...
+    for (id<SDLProtocolListener> listener in listeners) {
         if ([listener respondsToSelector:@selector(onProtocolClosed)]) {
             [listener onProtocolClosed];
         }
     }
 }
 
 - (void)onError:(NSString *)info exception:(NSException *)e {
-    for (id<SDLProtocolListener> listener in self.protocolDelegateTable.allObjects) {
+    NSArray<id<SDLProtocolListener>> *listeners = [self sdl_getProtocolListeners];
```
joeljfischer: Unnecessary

SmartDeviceLink/SDLProtocol.m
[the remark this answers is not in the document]
joeljfischer: Please remove

SmartDeviceLink/SDLProtocol.m
```diff
@@ -507,66 +566,98 @@ - (void)handleHeartbeatForSession:(Byte)session {
...
+        case SDLFrameInfoStartServiceNACK: // fallthrough
+        case SDLFrameInfoEndServiceNACK:
+            if (nakMessage.header.version >= 5) {
+                SDLControlFramePayloadNak *endServiceNakPayload = [[SDLControlFramePayloadNak alloc] initWithData:nakMessage.payload];
+                NSArray<NSString *> *rejectedParams = endServiceNakPayload.rejectedParams;
+                if (rejectedParams.count > 0) {
+                    SDLLogE(@"Start Service NAK'd, service type: %@, rejectedParams: %@", @(nakMessage.header.serviceType), rejectedParams);
+                }
+            }
+            break;
```
joeljfischer: Move up a line

SmartDeviceLink/SDLProtocol.h
shoamano83: OK, removed with commit b4455699e2f589b9e46db24e96e86015575f6442.
The `payload` parameter was added in case we want to extend the control frame in future. Now that `SDLProtocol` class is not public, we can freely update the code whenever we want to extend.

SmartDeviceLink/SDLProtocol.m
shoamano83: I remember your preference is to add `{ ... }` in each `case` block, so I updated the code into such style. Please refer to commit b4455699e2f589b9e46db24e96e86015575f6442.

SmartDeviceLink/SDLProtocol.m
shoamano83: I don't know why, I cannot add any comment to your review comment above. So I'm adding here.

Well, Apple's document says that `NSHashTable` isn't a thread-safe class.
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html

I created a simple code that accesses to a `NSHashTable` instance from two different queues. I didn't see any issue if one of the queue is calling `addObject` while another is calling `allObjects`. However, if one of the queue is calling `addObject` and `removeObject`, and the other is calling `allObjects`, I saw the program crashes on the second queue.
My sample code is attached: [NSHashTableTest.zip](https://github.com/smartdevicelink/sdl_ios/files/2289745/NSHashTableTest.zip)

I do understand your concern that locking slows down the message processing. Another solution is to access `protocolDelegateTable` from a single queue, which is `SDLProtocol`'s `_receiveQueue`. When we call `addObject` or `removeObject` on the table, we do it inside `dispatch_sync(_receiveQueue, ^{ ... });`. Also, `onTransportConnected` and `onTransportDisconnected` will be updated to acquire the list of listeners inside `dispatch_sync(_receiveQueue, ^{ ... });`. In this way, I think we can achieve thread safety while removing the locks from the callbacks. I think this idea requires adding dedicated methods like `addProtocolListener` and `removeProtocolListener` in `SDLProtocol` class. So  I hesitated to implement it at the first place (since `SDLProtocol` was a public class at that time.)

If this sounds good for you, then I will investigate the code again and will update the code in such way.

joeljfischer: Thanks for doing that investigation, it's greatly appreciated. I agree that `@synchronized` is necessary unless we only access on one queue. Since we currently are not, (and I don't think we should heavily use `dispatch_sync` as that will probably be slower). I would also like to note, however, that we don't need to use `allObjects` to create an array. We can simply loop over the `table` as is (though it would have to be synchronized.

SmartDeviceLink/SDLSecondaryTransportManager.m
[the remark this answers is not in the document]
joeljfischer: I'm still not very happy with it, but I understand. You are correct that ideally, the `SDLProtocolListener` protocol would (1) be renamed to `SDLProtocolDelegate`, and (2) the methods would be of the form: `-(void)protocol:didReceiveStartServiceACK:(SDLProtocolMessage *)startServiceACK`.

Ideally those changes would be made, but if you don't want to take that on, I'm okay with a rename of `SDLPrimaryProtocolDelegate` to `SDLSecondaryTransportPrimaryProtocolHandler`, which is obnoxiously long, but more descriptive.

SmartDeviceLink/SDLSecondaryTransportManager.m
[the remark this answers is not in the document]
joeljfischer: Perhaps the `lifecycleQueue` from the `SDLLifecycleManager`? We have far too many queues already in the SDL library. We already have a plan to vastly reduce that number, as currently, when SDL is added to a complex app, the number of queues and threads is enormous, bogging down the performance of the app. We should try to reuse a queue if possible, if not, I'll live with (1) for now, with the caveat that it needs to be removed in the next version when the restructuring of queues happens.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,699 @@
...
+    // sdl_handleTransportUpdateWithPrimaryAvailable is called in stop method
+    [self sdl_disconnectSecondaryTransport];
+}
+
+- (void)didEnterStateReconnecting {
+    __weak typeof(self) weakSelf = self;
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(RetryConnectionDelay * NSEC_PER_SEC)), _stateMachineQueue, ^{
+        if ([weakSelf.stateMachine isCurrentState:SDLSecondaryTransportStateReconnecting]) {
+            SDLLogD(@"Retry secondary transport after disconnection or registration failure");
+            [self.stateMachine transitionToState:SDLSecondaryTransportStateConfigured];
```
joeljfischer: Should use `weakself`

SmartDeviceLink/SDLSecondaryTransportManager.m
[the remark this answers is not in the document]
joeljfischer: It may closer match to what core sends, but I think it complicates the logic. When we ask the question "what are the transports for audio" we don't expect `NULL`, we expect the answer to be "none". Or, perhaps more clearly, if we ask, "how many transports for audio do we have", we don't expect the answer to be `NULL`, it would be `0` if there are none. Not `NULL` in some cases and `0` in others.

SmartDeviceLink/SDLSecondaryTransportManager.m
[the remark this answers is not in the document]
joeljfischer: Perhaps verbose, in that case. It's not something the app developer should really care about.

SmartDeviceLink/SDLSecondaryTransportManager.m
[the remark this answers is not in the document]
shoamano83: Thanks for understanding. I'd love to make the overhaul but that will introduce quite a few changes to non-related classes. I hope I can just rename `SDLPrimaryProtocolDelegate` for now, and make the other changes in another PR.

SmartDeviceLink/SDLSecondaryTransportManager.m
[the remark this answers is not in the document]
shoamano83: Thank you for your suggestion. Reusing `lifecycleQueue` seems good. Please give me a little more time as I need to update the test code as well.

SmartDeviceLink/SDLSecondaryTransportManager.m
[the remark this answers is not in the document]
shoamano83: OK, understood. I updated the code in commit b32fac3ea910101e1856f927bd573bdf0fdfe4f2.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,699 @@
...
+    }
+}
+
+- (nullable NSArray<SDLTransportClassBox *> *)sdl_convertServiceTransports:(nullable NSArray<NSNumber *> *)transports {
+    if (transports == nil) {
+        return nil;
+    }
+
+    // Actually there is nothing to "convert" here. We just check the range and recreate an array.
+    NSMutableArray<SDLTransportClassBox *> *array = [NSMutableArray new];
```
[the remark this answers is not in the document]
shoamano83: Fixed in commit bd3c8ebbe1d263b358ad58ac3840056a47057d3d.

SmartDeviceLink/SDLProtocol.m
joeljfischer: That is true, however, the rest of the receive calls would need to be in a `dispatch_async`, which is likely to complicate our logic.

SmartDeviceLink/SDLSecondaryTransportManager.m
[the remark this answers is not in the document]
shoamano83: The renaming has been performed in commit 2306f1c70c2eaed0833182f9d62ec10ca9cb9881.

SmartDeviceLink/SDLSecondaryTransportManager.m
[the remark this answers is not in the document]
shoamano83: I implemented the change in commit 097c8abeb2f86db3ca2c6e12db461a9c660c27ef.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,711 @@
...
+    self.primaryProtocolHandler = [[SDLSecondaryTransportPrimaryProtocolHandler alloc] initWithSecondaryTransportManager:self primaryProtocol:primaryProtocol];
+
+    [self.stateMachine transitionToState:SDLSecondaryTransportStateStarted];
+}
+
+- (void)stop {
+    SDLLogD(@"SDLSecondaryTransportManager stop");
+
+    // this method must be called in SDLLifecycleManager's state machine queue
+    dispatch_assert_queue(self.stateMachineQueue);
```
shoamano83: This assertion creates a failure in a unit test `a_lifecycle_manager__when_started__in_the_ready_state__stopping_the_manager__when_receiving_an_unregister_response__should_stop` of SDLLifecycleManagerSpec.

Looking at the code, I think this is rather an issue of `SDLLifecycleManager`. There is a case when `didEnterStateUnregistering` method is called from `SDLResponseDispatcher`, and in that case `didEnterStateUnregistering` method may not be called on the state machine queue. (I think the state machine methods should be always invoked on the dedicated queue.)

Right now I am not sure if this is a real bug or it's just an issue specific to that unit test case.

shoamano83:
I think I completed updating the code based on the suggestions.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: I found another case where an app terminates because of this `dispatch_assert_queue()`. When I disables a transport (like turning off BT), SDLLifecycleManager's `sdl_stopManager:` calls `transitionToState` method on the *main queue*, which eventually triggers SDLSecondaryTransportManager's `startWithPrimaryProtocol` and reaches `dispatch_assert_queue`.
https://github.com/smartdevicelink/sdl_ios/blob/develop/SmartDeviceLink/SDLLifecycleManager.m#L251

I think it is better to make it an error log rather than an assertion, so that developers will not face app terminations during development. I put my change with commit 2eaa2fd. Hopefully I can create another PR to fix the queue issue.

That commit also resolves the unit test failure.

shoamano83: The issue of the state machine running on unexpected queues should be addressed by https://github.com/smartdevicelink/sdl_ios/pull/1069.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,715 @@
...
+- (void)sdl_configureManager:(nullable NSArray<SDLSecondaryTransportTypeBox *> *)availableSecondaryTransports
+          availableTransportsForAudio:(nullable NSArray<SDLTransportClassBox *> *)availableTransportsForAudio
+          availableTransportsForVideo:(nullable NSArray<SDLTransportClassBox *> *)availableTransportsForVideo {
+    if (![self.stateMachine isCurrentState:SDLSecondaryTransportStateStarted]) {
+        SDLLogW(@"SecondaryTransportManager ignores duplicate Start Service ACK frame");
+        return;
+    }
+
+    SDLSecondaryTransportType secondaryTransportType = SDLSecondaryTransportTypeDisabled;
+    if (availableSecondaryTransports == nil || availableSecondaryTransports.count == 0) {
```
joeljfischer: `availableSecondaryTransports.count` will return `0` if it's `nil`. In this case, it's okay to only have `if(availableSecondaryTransports.count == 0)`

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,715 @@
...
+    } else if (secondaryTransportType == SDLSecondaryTransportTypeIAP) {
+        SDLLogW(@"Starting IAP as secondary transport, which does not usually happen");
+    }
+
+    // If SDL Core did not send transport list for the service, start it on Primary Transport.
+    if (availableTransportsForAudio == nil) {
+        SDLLogW(@"Transport for audio service is not specified by the system, using primary transport.");
+        availableTransportsForAudio = @[@(SDLTransportClassPrimary)];
+    }
+    if (availableTransportsForVideo == nil) {
```
joeljfischer: This `if` and the above are kind of redundant and can be combined, see above comment

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,715 @@
...
+        SDLLogW(@"Did not receive secondary transport type from system. Secondary transport is disabled.");
+        secondaryTransportType = SDLSecondaryTransportTypeDisabled;
+    } else {
+        // current proposal says the list should contain only one element
+        SDLSecondaryTransportTypeBox *transportType = availableSecondaryTransports[0];
+        secondaryTransportType = [transportType integerValue];
+    }
+
+    SDLSecondaryTransportType primaryTransportType = [self sdl_getTransportTypeFromProtocol:self.primaryProtocol];
+    if (primaryTransportType == secondaryTransportType) {
```
joeljfischer: I think this `if` can be `if (primaryTransportType == secondaryTransportType || primaryTransportType == SDLSecondaryTransportTypeDisabled)` and the below `if`s removed.

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -0,0 +1,715 @@
...
+typedef NSNumber SDLServiceTypeBox;
+
+typedef NS_ENUM(NSInteger, SDLTransportClass) {
+    SDLTransportClassInvalid = 0,
+    SDLTransportClassPrimary = 1,
+    SDLTransportClassSecondary = 2,
+};
+typedef NSNumber SDLTransportClassBox;
+
+typedef NS_ENUM(NSInteger, SDLSecondaryTransportType) {
```
joeljfischer: These types are also used to describe the primary transport, so perhaps `SDLTransportType` is better?

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Fixed in commit 068f8c142989fb8035fffed02f9c5d818b25d8cb.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Hmm, unfortunately we cannot combine the logics. In the case of `primaryTransportType == secondaryTransportType` we set `secondaryTransportType` to `SDLSecondaryTransportTypeDisabled`, while in the case of `availableTransportsForAudio == nil` we won't update `secondaryTransportType`. Besides, `primaryTransportType` will not be `SDLSecondaryTransportTypeDisabled`.

I agree that intializing `availableTransportsForAudio` multiple times looks redundant, so I tweaked the logic in commit 068f8c142989fb8035fffed02f9c5d818b25d8cb. The default values are set in the first place, then they will be overwritten. Hope it will make the code cleaner.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: I updated the code in commit commit 068f8c142989fb8035fffed02f9c5d818b25d8cb. Hope the code is cleaner now.

SmartDeviceLink/SDLSecondaryTransportManager.m
shoamano83: Actually I preferred `SDLTransportType`. The thing was, we already have a `@protocol` named `SDLTransportType`. If I rename the enum to `SDLTransportType` it won't break the build, but it may give confusions to developers. Can I go on to rename the enum?

Some alternatives: `SDLTransportTypeEnum`, `SDLTransportKind`, `SDLTransportUsage`, `SDLTransportEmployment`

shoamano83:
Thanks for the info. I updated `SDLLogFileModuleMap.m` file in commit 518015176789e425626989205cf0659616f63be8.

PR #958, opened 2018-05-14
Fixed image soft buttons getting sent early

### LINKED ISSUE #955: The screen manager's soft button manager sends image buttons before images are uploaded
### Bug Report
The screen manager's soft button manager sends image buttons before said images are uploaded.

The code below:
```
...
} else if (([self sdl_currentStateHasImages] && ![self sdl_allCurrentStateImagesAreUploaded])
               && (self.softButtonCapabilities ? !self.softButtonCapabilities.imageSupported : YES)) {
...
```

should be:
```
...
} else if ((([self sdl_currentStateHasImages] && ![self sdl_allCurrentStateImagesAreUploaded])
               && (self.softButtonCapabilities ? self.softButtonCapabilities.imageSupported.boolValue : NO))
               || (self.softButtonCapabilities ? !self.softButtonCapabilities.imageSupported.boolValue : YES)) {
...
```

##### Reproduction Steps
1. Connect app to Core and open SDL app. (App should have at least 1 soft button)

##### Expected Behavior
If the button has an image not yet uploaded to Core, the manager should wait to send the image button till the image has been successfully uploaded to Core.

##### Observed Behavior
If the button has an image not yet uploaded to Core, the manager still sends the image button. Core returns an `INVALID_DATA` error.

* iOS Version: n/a
* SDL iOS Version: 5.3
* Testing Against: TDK (SYNC 3)

The SDL_iOS library example app.

NicoleYarroch:
Fixes #955

### Risk

### Testing Plan
Added additional test cases to the *SDLSoftButtonManagerSpec*.

### Summary
* Previously, the image soft buttons were getting sent even if the artworks where not yet uploaded to the remote because `softButtonCapabilities.imageSupported` always returned `false`. This is due to  the `imageSupported` datatype being a `NSNumber`.

##### Bug Fix
* Image soft buttons are no longer sent by the *SDLSoftButtonManager* before the image is uploaded to the remote.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLSoftButtonManager.m
+++ b/SmartDeviceLink/SDLSoftButtonManager.m
@@ -203,14 +203,12 @@ NS_ASSUME_NONNULL_BEGIN
     self.inProgressUpdate = [[SDLShow alloc] init];
     self.inProgressUpdate.mainField1 = self.currentMainField1 ?: @"";
 
-    BOOL buttonHasImages = [self sdl_currentStateHasImages];
-    BOOL allButtonImagesAreUploaded = [self sdl_allCurrentStateImagesAreUploaded];
     BOOL headUnitSupportsImages = self.softButtonCapabilities ? self.softButtonCapabilities.imageSupported.boolValue : NO;
 
     if (self.softButtonObjects == nil) {
         SDLLogV(@"Soft button objects are nil, sending an empty array");
         self.inProgressUpdate.softButtons = @[];
-    } else if ((buttonHasImages && !allButtonImagesAreUploaded)
+    } else if (([self sdl_currentStateHasImages] && ![self sdl_allCurrentStateImagesAreUploaded])
                || !headUnitSupportsImages) {
         // The images don't yet exist on the head unit, or we cannot use images, send a text update, if possible. Otherwise, don't send anything yet.
         NSArray<SDLSoftButton *> *textOnlyButtons = [self sdl_textButtonsForCurrentState];
--- a/SmartDeviceLink/SDLSoftButtonManager.m
+++ b/SmartDeviceLink/SDLSoftButtonManager.m
@@ -202,13 +202,17 @@ NS_ASSUME_NONNULL_BEGIN
     self.inProgressHandler = [handler copy];
     self.inProgressUpdate = [[SDLShow alloc] init];
     self.inProgressUpdate.mainField1 = self.currentMainField1 ?: @"";
+
+    BOOL buttonHasImages = [self sdl_currentStateHasImages];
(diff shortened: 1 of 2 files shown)


joeljfischer:
> Image soft buttons are no longer sent by the SDLSoftButtonManager before the image is uploaded to the remote.

Isn't this a bugfix?

SmartDeviceLink/SDLSoftButtonManager.m
```diff
@@ -205,8 +205,9 @@ - (void)sdl_updateWithCompletionHandler:(nullable SDLSoftButtonUpdateCompletionH
     if (self.softButtonObjects == nil) {
         SDLLogV(@"Soft button objects are nil, sending an empty array");
         self.inProgressUpdate.softButtons = @[];
-    } else if (([self sdl_currentStateHasImages] && ![self sdl_allCurrentStateImagesAreUploaded])
-               && (self.softButtonCapabilities ? !self.softButtonCapabilities.imageSupported : YES)) {
+    } else if ((([self sdl_currentStateHasImages] && ![self sdl_allCurrentStateImagesAreUploaded])
+               && (self.softButtonCapabilities ? self.softButtonCapabilities.imageSupported.boolValue : NO))
+               || (self.softButtonCapabilities ? !self.softButtonCapabilities.imageSupported.boolValue : YES)){
```
joeljfischer: This if statement is super confusing, the parenthesis and newlines don't line up like you'd expect, there are too many parenthesis than are necessary `((A && B) && C)` == `(A && B && C)`, you use a BOOL value in one section and reverse check it in another. Please simplify it and pull out the `imagesSupported` BOOL into a local variable.

Also, I'm not sure this is correct, it seems to be saying (if i'm following the parenthesis right) that if:

1. The current state has images AND the images are not uploaded AND the system doesn't support images, send text buttons
2. If the head unit doesn't support images, send text buttons

NicoleYarroch: Fixed

SmartDeviceLink/SDLSoftButtonManager.m
```diff
@@ -202,13 +202,17 @@ - (void)sdl_updateWithCompletionHandler:(nullable SDLSoftButtonUpdateCompletionH
     self.inProgressHandler = [handler copy];
     self.inProgressUpdate = [[SDLShow alloc] init];
     self.inProgressUpdate.mainField1 = self.currentMainField1 ?: @"";
+
+    BOOL buttonHasImages = [self sdl_currentStateHasImages];
+    BOOL allButtonImagesAreUploaded = [self sdl_allCurrentStateImagesAreUploaded];
```
joeljfischer: This and the above line are unnecessary for clarity, only the below line helps.

PR #980, opened 2018-05-24
Suspend and resume instead of restart video stream

### LINKED ISSUE #979: Suspend and resume instead of restart video stream
### Issue Report
Currently the streaming media manager closes and reopens the video service and recreates a new video encoder in case the app moves from Background to Active (visible) application state on the phone. This issue improves the video stream by keeping the video stream open recreating the video encoder only.

##### Reproduction Steps
1. Start the MobileNav app on the iPhone
2. Connect the phone to an SDL head unit and select the app after it registers
3. Wait for the app to start streaming video to the head unit
4. Press Home button on the phone
5. Reopen the app on the phone

##### Expected Behavior
At 4. The SDL media manager should only suspend the video stream
At 5. The app should resume the video stream immediately

##### Observed Behavior
Because the SDL media manager restarts the video stream which takes a few seconds

* SDL iOS Version: 5.2

ghost:
Fixes #979 

### Risk
This PR makes **minor** API changes:
- New video stream state Suspended added. It'll occur if the stream was ready and the app entered inactive state
- Property `shouldRestartVideoStream` is deprecated as not used anymore and can be deleted with the next major version.

### Testing Plan
There was a unit test if the app enters inactive state. The test was changed to match the new state.

##### Enhancements
* Resuming a suspended video stream is faster (<1s) than restarting it (>1s)

### CLA

### DIFF
--- a/SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
+++ b/SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
@@ -443,7 +443,9 @@ typedef void(^SDLVideoCapabilityResponseHandler)(SDLVideoStreamingCapability *_N
 }
 
 - (void)didEnterStateVideoStreamSuspended {
+    SDLLogD(@"Video stream suspended");
     
+    [[NSNotificationCenter defaultCenter] postNotificationName:SDLVideoStreamSuspendedNotification object:nil];
 }
 
 - (void)didEnterStateVideoStreamShuttingDown {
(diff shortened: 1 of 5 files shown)


SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
```diff
@@ -434,6 +442,10 @@ - (void)didEnterStateVideoStreamReady {
     }
 }
 
+- (void)didEnterStateVideoStreamSuspended {
```
NicoleYarroch: This method is unused

NicoleYarroch:
Every time the video comes to the foreground, a `SDLVideoStreamDidStartNotification` is broadcast. However, no corresponding `SDLVideoStreamDidStopNotification` is sent. Perhaps there should be another notification called `SDLVideoStreamSuspendedNotification` that is broadcast when the video stream is suspended.

NicoleYarroch (changes requested):
I left one comment on the PR about fixing notifications.

SmartDeviceLink/SDLStreamingMediaLifecycleManager.m
kshala-ford: True. I wasn‘t sure if the state methods *must* be implemented. I can remove it if it‘s not necessary.

kshala-ford: @NicoleYarroch I added the requested suspended notification to the method.

NicoleYarroch (approved):

PR #1004, opened 2018-07-02
Fix softbuttons disappearing when exiting HMI NONE

### LINKED ISSUE #1000: Soft Buttons sent via RPC disappear
### Bug Report
On v5.2.0+, when sending a soft button via the Show RPC, they flicker, and appear then disappear.

> Nicole was chatting with me on Monday and she thought that the `screenManager` might be doing something things that were causing my screen display issues. Just to be clear, I updated from 5.11 to 5.20 and did not make a single change (still using RPCs), but my buttons were appearing and then disappearing instantly.  Using screen manager to handle button display fixed the disappearing issue, but looks like it requires me to transition several other things so RPCs and screenMgr dont collide.

##### Reproduction Steps
```
on hmiLevel didChangeToLevel:
if newLevel == .Full && manager is ready;
Upload a set of base images for icons if needed

Iterate on the 4 items (our 4 command types, like TodaysForecast or WeatherAlerts).
Make SDLAddCommands for the VR strings.
Make buttons for each command.
Send addcommand via manager
End repeat.
Add buttons via SDLShow via manager
Add help phrases to unit via SDLSetGlobalProperties

then
dispatch after 1 second: SDLSetDisplayLayout with SDLPredefinedLayoutTextAndSoftButtonsWithGraphic
```

##### Expected Behavior
Soft buttons remain on the screen

##### Observed Behavior
Soft buttons disappear

* iOS Version: n/a
* SDL iOS Version: v.5.2.0+
* Testing Against: unknown

joeljfischer:
Fixes #1000

### Risk

### Testing Plan
Unit tests were run and smoke tests performed

### Summary
This PR makes fixes for developers who are not using the screen manager in their app.

##### Bug Fixes
* When running the app without using the soft button manager, buttons should not reset when exiting HMI NONE.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLSoftButtonManager.m
+++ b/SmartDeviceLink/SDLSoftButtonManager.m
@@ -116,11 +116,11 @@ NS_ASSUME_NONNULL_BEGIN
         }
     }
 
-    _softButtonObjects = softButtonObjects;
-
-    for (SDLSoftButtonObject *button in _softButtonObjects) {
+    for (SDLSoftButtonObject *button in softButtonObjects) {
         button.manager = self;
     }
+
+    _softButtonObjects = softButtonObjects;
 
     [self updateWithCompletionHandler:nil];
 }
--- a/SmartDeviceLink/SDLSoftButtonManager.m
+++ b/SmartDeviceLink/SDLSoftButtonManager.m
@@ -50,7 +50,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (strong, nonatomic, nullable) SDLDisplayCapabilities *displayCapabilities;
 @property (strong, nonatomic, nullable) SDLSoftButtonCapabilities *softButtonCapabilities;
 
-@property (assign, nonatomic) BOOL waitingOnHMILevelUpdateToSetButtons;
+@property (assign, nonatomic) BOOL waitingOnHMILevelUpdateToUpdate;
 
 @end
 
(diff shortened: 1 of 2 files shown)


NicoleYarroch:
An empty array is sent as part of the `inProgressUpdate.softButtons` when the developer does not set the `softButtonObjects` for the `screenManager`. This happens when `sdl_softButtonsForCurrentState` returns an empty array. This can overwrite soft buttons sent in a separate `SDLShow` RPC.
https://github.com/smartdevicelink/sdl_ios/blob/906c99ab2e621e9e710ad0cdf57eaab8d2d7fe24/SmartDeviceLink/SDLSoftButtonManager.m#L197-L200

joeljfischer:
@NicoleYarroch I think you're looking in the wrong spot. Being set there isn't an issue, because that method should never get called if no soft button objects have ever been set. I just pushed a fix that should prevent display layout changes from causing an update

NicoleYarroch:
@joeljfischer The commit you pushed did not fix the issue I mentioned. What is happening is that the screen manager tells the soft button manager to update in the `endUpdatesWithCompletionHandler` method even if no soft buttons were set. 

https://github.com/smartdevicelink/sdl_ios/blob/093b7f5feeb0c9014d054b5c8e8f8562198bc230/SmartDeviceLink/SDLScreenManager.m#L209-L215

It is not possible to just check if the `softButtonObjects.count > 0` in the `SDLSoftButtonManager`'s `updateWithCompletionHandler:` method because setting an empty array is supposed to clear the soft buttons when using the screen manager.

joeljfischer:
@NicoleYarroch Should be fixed, I tested fairly extensively

PR #1085, opened 2018-09-27
Check state before disconnecting proxy

### LINKED ISSUE #1079: IAP-BT to IAP-USB causes [Lifecycle] Failed to register the app
[HelloSDLBT2USBConsole.log](https://github.com/smartdevicelink/sdl_ios/files/2658521/HelloSDLBT2USBConsole.log)
### Bug Report
Connecting device from IAP-BT to IAP-USB causes manager to stop registering to HU

##### Reproduction Steps
Ensure you have SDL enabled App on Phone.
Pair and connect iPhone over BT to Sync 3.2v2
App should register on SYNC
Now, connect the same device via USB to SYNC
(Internally, when the device connects over USB the device and SYNC setups another iAP session over USB. Once SYNC can determine that it is the same device which is connected over BT and USB, SYNC will proceed to disconnect the iAP session over BT)
Note the messages received on App (SDL disconnects, doesn't appear on head unit)
 
##### Expected Behavior
App should un-register since they were connected over BT and then immediately re-register over USB

##### Observed Behavior
App unregisters and then disconnects the manager completely over USB

* iOS Version: iOS 11, iOS 12
* SDL iOS Version: 6.0.1
* Testing Against: SYNC 3.2v2

SatbirTanda:
Fixes #1079 

This PR is ready for review.

### Risk
This PR makes no API changes.

### Testing Plan
Test added 

### Summary
Check lifecycle state before allowing SDLManager to invoke stop method

##### Bug Fixes
* App will display when switching from iAP BT to iAP USB if developer calls stop in managerDidDisconnect delegate method

### CLA

### DIFF
--- a/SmartDeviceLink/SDLLifecycleManager.m
+++ b/SmartDeviceLink/SDLLifecycleManager.m
@@ -161,14 +161,16 @@ SDLLifecycleState *const SDLLifecycleStateReady = @"Ready";
 }
 
 - (void)stop {
-    dispatch_sync(_lifecycleQueue, ^{
-        SDLLogD(@"Lifecycle manager stopped");
-        if ([self.lifecycleStateMachine isCurrentState:SDLLifecycleStateReady]) {
-            [self.lifecycleStateMachine transitionToState:SDLLifecycleStateUnregistering];
-        } else {
-            [self.lifecycleStateMachine transitionToState:SDLLifecycleStateStopped];
-        }
-    });
+    if (![self.lifecycleStateMachine isCurrentState:SDLLifecycleStateReconnecting]) {
+        dispatch_sync(_lifecycleQueue, ^{
+            SDLLogD(@"Lifecycle manager stopped");
+            if ([self.lifecycleStateMachine isCurrentState:SDLLifecycleStateReady]) {
+                [self.lifecycleStateMachine transitionToState:SDLLifecycleStateUnregistering];
+            } else {
+                [self.lifecycleStateMachine transitionToState:SDLLifecycleStateStopped];
+            }
+        });
+    }
 }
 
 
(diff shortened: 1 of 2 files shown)


SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -161,14 +161,16 @@ - (void)sdl_startWithReadyHandler:(SDLManagerReadyBlock)readyHandler {
 }
 
 - (void)stop {
```
joeljfischer: Why is this method being called at all in this situation? Your issue doesn't make it clear why `stop` would be called.

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -161,14 +161,16 @@ - (void)sdl_startWithReadyHandler:(SDLManagerReadyBlock)readyHandler {
...
-    dispatch_sync(_lifecycleQueue, ^{
-        SDLLogD(@"Lifecycle manager stopped");
-        if ([self.lifecycleStateMachine isCurrentState:SDLLifecycleStateReady]) {
-            [self.lifecycleStateMachine transitionToState:SDLLifecycleStateUnregistering];
-        } else {
-            [self.lifecycleStateMachine transitionToState:SDLLifecycleStateStopped];
-        }
-    });
+    if (![self.lifecycleStateMachine isCurrentState:SDLLifecycleStateReconnecting]) {
```
joeljfischer: Without a very good reason, we don't want to make the developer's call to `stop` simply not work in some circumstances. That can be a source of bugs and confusion.

SmartDeviceLink/SDLLifecycleManager.m
SatbirTanda: Some apps are checking the firmware version of the External Accessory that is connected to make sure they want the app to show on that specific hardware (ex.  Navigation apps may not want to be shown on SYNC 1.1 but do want to be shown on SYNC 3).  So they are observing the `EAAccessoryDidConnectNotification` and `EAAccessoryDidDisconnectNotification` themselves and calling `start` after checking the EAAccessory `firmwareRevision` property of `EAAccessory`.  They call `stop` on `EAAccessoryDidDisconnectNotification`.  Even if an app calls `stop` in `managerDidDisconnect` delegate method we see this issue.

PR #1114, opened 2018-10-24
fix to call sdl_retryEstablishSession after data sesion is closed

### LINKED ISSUE #1113: sdl_retryEstablishSession is not called after EA data session is closed
### Bug Report
We would like to enable connection retry logic in SDLIAPTransport when EA session is closed and the accessory is still connected.

One of our customer's head unit requires Bluetooth HFP connection with iPhone as a prerequisite to use connected services, including SDL. The head unit closes EA sessions between the phone when Bluetooth link is removed.

Now, when the head unit is connected to the phone with both Bluetooth and USB, it establishes EA sessions over USB. When Bluetooth link is disconnected, it closes EA sessions to shut down connected services. iOS library receives NSStreamEventEndEncountered, however since USB cable is still connected, it doesn't receive accessory disconnected event.

We would like to run sdl_retryEstablishSession in such case, so that once Bluetooth link is restored and EA session is established again, SDL gets available. Without the retries, the user is forced to remove USB cable once (to generate accessory disconnected event) then plug it again to enable SDL.

##### Reproduction Steps
Let HU close EA session w/o disconnection of accessory

##### Expected Behavior
Proxy starts reconnection

##### Observed Behavior
Proxy does not start reconnection

* iOS Version: All
* SDL iOS Version: 6.1.0
* Testing Against: Our customer's HU

I will create PR

t-yoshii:
Fixes #1113

### Risk

### Testing Plan
I tested with our customer's HU

### Summary
Call sdl_retryEstablishSession  after HU closed EA sessions.

##### Breaking Changes
* [Breaking change info]

##### Enhancements
* Proxy starts reconnection sequence after HU closes EA session

### CLA

### DIFF
--- a/SmartDeviceLink/SDLIAPTransport.m
+++ b/SmartDeviceLink/SDLIAPTransport.m
@@ -564,7 +564,7 @@ int const ProtocolIndexTimeoutSeconds = 10;
             strongSelf.session.streamDelegate = nil;
             strongSelf.session = nil;
             
-            // We don't call sdl_retryEstablishSession here because the stream end event usually fires when the accessory is disconnected
+            [strongSelf sdl_retryEstablishSession];
         });
         
         // To prevent deadlocks the handler must return to the runloop and not block the thread


joeljfischer:
Hi @t-yoshii, because this is a change to connection code, we need additional information including possible side-effects and what head units you have regression tested against.

@jordynmackool In addition to our own testing, this may require all OEMs and steering committee members to be aware of it and to have the opportunity to do their own testing against their head units. Can we please add this PR to the agenda as a discussion item?

t-yoshii:
Hello @joeljfischer ,

Thank you for comments.

> additional information including possible side-effects

Regarding to possible side-effects, when user disconnects accessory, "NSStreamEventEndEncountered" and "Accessory Disconnected" events can be notified within a short period of time. In this situation, after this PR's change, `sdl_retryEstablishSession` is called by "NSStreamEventEndEncountered" but reconnection attempt is terminated immediately by "Accessory Disconnected" event. This behavior may cause some timing issues, however, as far as I checked, SDLIAPSession performs state transition correctly even in this case. So, I do not think this causes problems.
I do not have any other concerns now.


> what head units you have regression tested against.

Actually, there is no real HU which send "NSStreamEventEndEncountered" now. This behavior change is planned at next model. We implemented it to our dev board in advance and tested against it.

PR #1201, opened 2019-03-21
API cleanup

joeljfischer:

### Risk
This PR makes **major** API changes (all API changes are internal to PR #1200).

### Testing Plan
Unit tests have been updated

### Summary
This PR updates some APIs before release for ease-of-use.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLManager.h
+++ b/SmartDeviceLink/SDLManager.h
@@ -85,11 +85,6 @@ typedef void (^SDLManagerReadyBlock)(BOOL success, NSError *_Nullable error);
 @property (strong, nonatomic, readonly, nullable) SDLRegisterAppInterfaceResponse *registerResponse;
 
 /**
- *  The auth token, if received. This should be used to log into a user account. Primarily used for cloud apps with companion app stores.
- */
-@property (strong, nonatomic, readonly, nullable) NSString *authToken;
-
-/**
  *  The manager's delegate.
  */
 @property (weak, nonatomic, nullable) id<SDLManagerDelegate> delegate;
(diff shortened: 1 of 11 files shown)


SmartDeviceLink/SDLManager.h
```diff
@@ -84,11 +84,6 @@ typedef void (^SDLManagerReadyBlock)(BOOL success, NSError *_Nullable error);
  */
 @property (strong, nonatomic, readonly, nullable) SDLRegisterAppInterfaceResponse *registerResponse;
 
-/**
```
NicoleYarroch: This should remain (even though it will always be `nil`) because the library should support the protocol. Android has the same accessor method in their library: https://github.com/smartdevicelink/sdl_android/blob/6f05f2d8cf18aa43369806d1e86b8ce90b144976/android/sdl_android/src/main/java/com/smartdevicelink/managers/SdlManager.java#L384

SmartDeviceLink/SDLManager.m
```diff
@@ -97,10 +97,6 @@ - (nullable SDLRegisterAppInterfaceResponse *)registerResponse {
     return self.lifecycleManager.registerResponse;
 }
 
--(nullable NSString *)authToken {
```
NicoleYarroch: Should remain.

SmartDeviceLink/SDLManager.h
joeljfischer: Core said that it will never work though. It would be better to add it when it's actually functional, otherwise it's just confusing. @joeygrover

PR #1209, opened 2019-03-28
Hit test for a single tap now checked on main thread

### LINKED ISSUE #1207: Using the haptic input feature freezes the app when a single tap is detected
### Bug Report
When using the haptic input feature to detect the tapped view in a navigation app, the app freezes whenever a single tap is detected. This is because the library does not check on main thread if a view was tapped.

##### Reproduction Steps
1. Create a navigation app with some `UIViews` in the `rootViewController` (Make sure to send a SDLDidUpdateProjectionView notification to the library)
1. Connect to Core
1. Perform a single tap on a view on the SDL app

##### Expected Behavior
The touch manager should notify the subscriber right away that a view was tapped.

##### Observed Behavior
The app freezes for a few seconds before the notification is sent by the touch manager.

In the `SDLFocusableItemLocator` class, the `viewForPoint:` method should check if a tap is inside a view on the main thread.

* iOS Version: n/a
* SDL iOS Version: 6.2
* Testing Against: SYNC 3 and SDL_Core 5.1.0 + sdl_hmi

NicoleYarroch:
Fixes #1207 

### Risk

### Testing Plan
Test cases for the `SDLTouchManager` were updated.

### Summary
Checking if a single tap is inside a `UIView` no longer causes the app to freeze for a few seconds. All checking is now done on the main thread.

##### Breaking Changes
* n/a

##### Bug Fixes
* Checking if a single tap is inside a `UIView` no longer causes the app to freeze for a few seconds. All checking is now done on the main thread.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLTouchManager.m
+++ b/SmartDeviceLink/SDLTouchManager.m
@@ -420,9 +420,24 @@ static NSUInteger const MaximumNumberOfTouches = 2;
         strongSelf.singleTapTouch = nil;
         [strongSelf sdl_cancelSingleTapTimer];
         if ([strongSelf.touchEventDelegate respondsToSelector:@selector(touchManager:didReceiveSingleTapForView:atPoint:)]) {
-            UIView *hitView = (self.hitTester != nil) ? [self.hitTester viewForPoint:point] : nil;
-            [strongSelf.touchEventDelegate touchManager:strongSelf didReceiveSingleTapForView:hitView atPoint:point];
+            [self sdl_getSingleTapHitView:point hitViewHandler:^(UIView * _Nullable selectedView) {
+                [strongSelf.touchEventDelegate touchManager:strongSelf didReceiveSingleTapForView:selectedView atPoint:point];
+            }];
         }
+    });
+}
+
+- (void)sdl_getSingleTapHitView:(CGPoint)point hitViewHandler:(nullable void (^)(UIView * __nullable hitView))hitViewSelector {
+    if (!self.hitTester) {
+        return hitViewSelector(nil);
+    }
+
+    NSOperationQueue *currentQueue = NSOperationQueue.currentQueue;
+    dispatch_async(dispatch_get_main_queue(), ^{
+        UIView *hitView = [self.hitTester viewForPoint:point];
+        dispatch_async(currentQueue.underlyingQueue, ^{
+            return hitViewSelector(hitView);
+        });
     });
 }
 
@@ -440,4 +455,3 @@ static NSUInteger const MaximumNumberOfTouches = 2;
 @end
 
 NS_ASSUME_NONNULL_END
-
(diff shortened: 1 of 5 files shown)


NicoleYarroch:
After doing more testing, I realized that only the single tap touch was not being hit tested on the main thread. I added a HAX to perform the hit test on the main thread and then notify the delegate on the background thread.

SmartDeviceLink/SDLTouchManager.m
```diff
@@ -420,12 +420,34 @@ - (void)sdl_initializeSingleTapTimerAtPoint:(CGPoint)point {
...
+                [strongSelf.touchEventDelegate touchManager:strongSelf didReceiveSingleTapForView:selectedView atPoint:point];
+            }];
         }
     });
 }
 
+/**
+ *  HAX to preserve the thread on which the delegate is notified for single taps; returning on the main thread would be a breaking change. All other touch gestures currently notify the delegate on the main thread. The single tap timer runs on a background thread so when a single tap is detected the hit test needs to be done on the main thread and then the result is returned on a background thread.
+ *
+ * Checks if a single tap is inside a view. As the single tap timer is run on a background thread, the check is done on a main thread and then the result is returned on a background thread. This is done to ma
```
joeljfischer: Unfinished sentence

SmartDeviceLink/SDLTouchManager.m
```diff
@@ -420,12 +420,34 @@ - (void)sdl_initializeSingleTapTimerAtPoint:(CGPoint)point {
...
+ *
+ *  @param point            Screen coordinates of the tap gesture
+ *  @param hitViewHandler   A handler that returns the view the point is inside of; nil if the point does not lie inside a view
+ */
+- (void)sdl_getSingleTapHitView:(CGPoint)point hitViewHandler:(nullable void (^)(UIView * __nullable hitView))hitViewHandler {
+    if (!self.hitTester) {
+        return hitViewHandler(nil);
```
joeljfischer: Need to check if `hitViewHandler` exists since it's nullable.

SmartDeviceLink/SDLTouchManager.m
```diff
@@ -420,12 +420,34 @@ - (void)sdl_initializeSingleTapTimerAtPoint:(CGPoint)point {
...
+    }
+
+    dispatch_async(dispatch_get_main_queue(), ^{
+        UIView *hitView = [self.hitTester viewForPoint:point];
+        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+            return hitViewHandler(hitView);
```
joeljfischer: See above

SmartDeviceLink/SDLTouchManager.m
NicoleYarroch: Fixed. Added check.

SmartDeviceLink/SDLTouchManager.m
NicoleYarroch: Fixed. Removed.

SmartDeviceLink/SDLTouchManager.m
NicoleYarroch: Fixed. Added check.

PR #1215, opened 2019-04-04
Update System Capability Manager to auto-cache`OnSystemCapability`

### LINKED ISSUE #1212: Update the System Capability Manager to handle `OnSystemCapability` notifications and auto-caching
### Bug Report
Use `OnSystemCapability` when available to auto-update stored system capability properties. the `update` method would then simply return the cached property in the callback immediately instead of going to get the newest value (on v5.1+ systems only, older systems would have to still get the value by polling).

The app services capability, when updated (only exists on v5.1+ systems) will need to be merged since it only contains the updates, not a full new object.

* iOS Version: n/a
* SDL iOS Version: 6.2
* Testing Against: n/a

NicoleYarroch:
Fixes #1212 

### Risk

### Testing Plan
Test cases added to the system capability manager

### Summary
* Support `OnSystemCapability` and `GetSystemCapability` with the `subscribe` parameter while also backward compatibility supporting head units that don't support those updates.
* Support updating `AppServicesCapabilities` parameter with the change set update.

##### Enhancements
* Updated the `SDLSystemCapabilityManager` to auto-update the cache based on all `GetSystemCapabilityResponse` and `OnSystemCapability` notifications (v5.1+ only) with full backward-compatibility support.

### Tasks Remaining:
- [x] Fix broken unit tests for the system capability manger

### CLA

### DIFF
--- a/SmartDeviceLink/SDLSystemCapabilityManager.m
+++ b/SmartDeviceLink/SDLSystemCapabilityManager.m
@@ -75,6 +75,9 @@ NS_ASSUME_NONNULL_BEGIN
     return self;
 }
 
+/**
+ *  Resets the capabilities when a transport session is closed.
+ */
 - (void)stop {
     SDLLogD(@"System Capability manager stopped");
     _displayCapabilities = nil;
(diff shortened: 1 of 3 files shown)


SmartDeviceLink/SDLSystemCapabilityManager.m
```diff
@@ -119,43 +149,128 @@ - (void)sdl_displayLayoutResponse:(SDLRPCResponseNotification *)notification {
...
+            self.appServicesCapabilities = systemCapability.appServicesCapabilities;
+        } else {
+            NSMutableDictionary *cachedAppServicesCapabilities = [NSMutableDictionary dictionary];
+            for (NSUInteger i = 0; i < self.appServicesCapabilities.appServices.count; i += 1) {
+                SDLAppServiceRecord *record = self.appServicesCapabilities.appServices[i].updatedAppServiceRecord;
+                cachedAppServicesCapabilities[record.serviceID] = record;
+            }
+
+            for (NSUInteger i = 0; i < systemCapability.appServicesCapabilities.appServices.count; i += 1) {
+                SDLAppServiceRecord *updatedRecord = systemCapability.appServicesCapabilities.appServices[i].updatedAppServiceRecord;
```
joeljfischer: You'll need to at least check the `updateReason` to remove the ones that no longer are published.

SmartDeviceLink/SDLSystemCapabilityManager.m
```diff
@@ -119,43 +149,128 @@ - (void)sdl_displayLayoutResponse:(SDLRPCResponseNotification *)notification {
...
-        self.videoStreamingCapability = systemCapabilityResponse.videoStreamingCapability;
+        self.videoStreamingCapability = systemCapability.videoStreamingCapability;
     } else if ([systemCapabilityType isEqualToEnum:SDLSystemCapabilityTypeAppServices]) {
-        self.appServicesCapabilities = systemCapabilityResponse.appServicesCapabilities;
+        if (!self.appServicesCapabilities) {
+            self.appServicesCapabilities = systemCapability.appServicesCapabilities;
+        } else {
+            NSMutableDictionary *cachedAppServicesCapabilities = [NSMutableDictionary dictionary];
+            for (NSUInteger i = 0; i < self.appServicesCapabilities.appServices.count; i += 1) {
+                SDLAppServiceRecord *record = self.appServicesCapabilities.appServices[i].updatedAppServiceRecord;
```
joeljfischer: I don't think this is probably the best approach. It's going to be super slow. It's probably better to just store a mutable dictionary of the values privately and then the public `appServicesCapability` property has a getter that assembles it based the `[dictionary allValues]` call. Hopefully that makes sense.

Issue #1237, opened 2019-04-15
Example apps `resetConnection` crashes app


NicoleYarroch:
### Bug Report
If starting the `sdlManager` fails with an error, the example apps call the `resetConnection` method which attempts to stop the manager. This crashes the app.

This should be fixed in the Swift and the Obj-C example apps.

##### Reproduction Steps
1. Connect the example app to a TDK
2. Pull the cord before the app registers. 
>  There was an error while starting up: Optional(Error Domain=com.sdl.lifecyclemanager.error Code=-2 "Could not find a connection" UserInfo={NSLocalizedRecoverySuggestion=Have you tried turning it off and on again?, NSLocalizedDescription=Could not find a connection, NSLocalizedFailureReason=The SDL library could not find a current connection to an SDL hardware device})

##### Expected Behavior
The app does not crash

##### Observed Behavior
The app crashes.

##### OS & Version Information
* iOS Version: 12.1.2
* SDL iOS Version: 6.2
* Testing Against: SYNC 3.0

PR #1244, opened 2019-04-18
Fix auto-reconnect / manual disconnect of example apps

### LINKED ISSUE #1227: Add a toggle to the example apps to use default auto-connect / reconnect behavior
### Bug Report
The current example apps don't use default connection behavior (which allows the user to start and stop at will), but sometimes this can be undesirable. We should add a settings pane and add the ability to toggle using default connection behavior.

* iOS Version: n/a
* SDL iOS Version: 6.2.0
* Testing Against: n/a

### LINKED ISSUE #1243: SDLManagerDelegate methods not getting called
### Bug Report
The `SDLManagerDelegate` methods stop getting notified when the manager is started from the `managerDidDisconnect` method. It appears that starting the manager when the manager is already started breaks the `SDLManagerDelegate`. More investigation is needed as I am not sure this is the root of the issue. 

Ideally when the manager is already started it should ignore the new start attempt.

##### Reproduction Steps
1. Connect the SDL_iOS Swift example app to a TDK with a USB cord. (The Obj-C example app does not have this bug)
2. Disconnect the USB cord
3. Connect the app using a USB cord. 
4. The `managerDidDisconnect` method will be called. It will attempt to restart the SDLManager even though the manager is already started. 

##### Expected Behavior
All the `SDLManagerDelegate` methods should be called at the appropriate times.

##### Observed Behavior
The `SDLManagerDelegate` methods stop getting called.

* iOS Version: 12.1
* SDL iOS Version: 6.2
* Testing Against: SYNC 3.0

The SDL_iOS Swift example app in the library.

joeljfischer:
Fixes #1227, #1237, #1243 

### Risk

### Testing Plan
Smoke testing

### Summary
This PR removes a bit of code about autoreconnecting and makes a UI fix

##### Bug Fixes
* Fixes the swift example app autoreconnect

### CLA

### DIFF
--- a/Example Apps/Example Swift/ProxyManager.swift
+++ b/Example Apps/Example Swift/ProxyManager.swift
@@ -141,13 +141,8 @@ private extension ProxyManager {
 extension ProxyManager: SDLManagerDelegate {
     /// Called when the connection beween this app and SDL Core has closed.
     func managerDidDisconnect() {
-        delegate?.didChangeProxyState(ProxyState.stopped)
+        delegate?.didChangeProxyState(ProxyState.searching)
         firstHMILevelState = .none
-
-        // If desired, automatically start searching for a new connection to Core
-        if ExampleAppShouldRestartSDLManagerOnDisconnect.boolValue {
-            startManager()
-        }
     }
 
     /// Called when the state of the SDL app has changed. The state limits the type of RPC that can be sent. Refer to the class documentation for each RPC to determine what state(s) the RPC can be sent.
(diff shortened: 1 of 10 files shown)


Example Apps/Example Swift/ProxyManager.swift
```diff
@@ -49,13 +49,14 @@ extension ProxyManager {
...
-    func resetConnection() {
+    func stopConnection() {
         guard sdlManager != nil else {
-            delegate?.didChangeProxyState(ProxyState.stopped)
+            delegate?.didChangeProxyState(.stopped)
             return
         }
 
         sdlManager.stop()
+        delegate?.didChangeProxyState(.stopped)
```
NicoleYarroch: I would move `delegate?.didChangeProxyState(.stopped)` to the `managerDidDisconnect` method.

PR #1252, opened 2019-05-02
Fix Lockscreen Window's Existence Messing with App Autorotation

### LINKED ISSUE #1250: SDLLockScreenPresenter overrides native app's supportedInterfaceOrientations and shouldAutorotate properties
### Bug Report
If a developer wants to create an app that supports both portrait and landscape orientation for some (not all) of their view controllers (with SDL integrated), then they should expect their portrait only view controllers to stay the same even if the device is rotated.  When `SDLLockScreenPresenter` is initialized, it creates a new `UIWindow` object that, for some reason, alters the properties of view controllers in other`UIWindow` objects in the app to match that of it's `rootViewController` - including those view controllers that are in the native app's `UIWindow` object.  This results in the status bar auto-rotating to conform to `SDLScreenshotViewController`'s `supportedInterfaceOrientations` and `shouldAutorotate` properties.

##### Reproduction Steps
1. Set the UIViewControllerBasedStatusBarAppearance property to YES in Info.plist of the app
```xml	
        <key>UIViewControllerBasedStatusBarAppearance</key>
	<true/>
```
2. Make sure your app supports Landscape left, landscape right, portrait (can be done in Target or Info.plist)
3. Now, you should be able to choose (through your code) which view controllers will support which orientations.
4.  Thus if one view controller only wanted to support portrait, then override supportedInterfaceOrientations to return .portrait 
5.  Create an `SDLManager` object and call its `start` method (in AppDelegate).
6.  Use app and rotate the device to landscape - you will see the status bar rotate even though the view controller supports portrait only

##### Expected Behavior
View controllers that only support the portrait orientation should not see the status bar rotate when the device rotates

##### Observed Behavior
View controllers that only support the portrait orientation do see the status bar rotate when the device rotates

* iOS Version: 12.2
* SDL iOS Version: 6.2.0
* Testing Against: There is no need to connect the iOS device to a Head Unit

I've attached two example apps that display the bug, one that initializes its AppDelegate's window property in code, and the other that initializes via Storyboards.
[StatusBarBug-Nibs.zip](https://github.com/smartdevicelink/sdl_ios/files/3119370/StatusBarBug-Nibs.zip)
[StatusBarBug-StoryBoard.zip](https://github.com/smartdevicelink/sdl_ios/files/3119371/StatusBarBug-StoryBoard.zip)

joeljfischer:
Fixes #1250 

### Risk

### Testing Plan
Smoke testing

### Summary
This PR fixes an [apple bug](http://openradar.appspot.com/radar?id=4985838173683712) where having a UIWindow (it does not even need to be active, simply created) like we do with the lockscreen can break the app's view controller based rotation.

##### Bug Fixes
* Fixes a case where the lock screen's existence (not even necessarily its visibility) can affect app view-controller based rotation preferences.

### CLA


SatbirTanda:
@joeljfischer Looks good @Ford would like a hotfix for this issue in 6.2.0

PR #1256, opened 2019-05-08
Throw an exception if SDL is setup before the app's root view controller

### LINKED ISSUE #1255: v6.2.1 Causes Infinite Recursion if the App's Window is Setup after SDL
### Bug Report
After the v6.2.1 lock screen fix, if an app sets up its app window _after_ it sets up SDL, it will cause infinite recursion due to the lock screen window being the only window that exists.

##### Reproduction Steps
1. Setup SDL in app delegate
2. Setup window in app delegate _after_ SDL
3. Run the app

##### Expected Behavior
App does not crash or gives explicit instruction to the developer on when to setup SDL.

##### Observed Behavior
App crashes without any way for the developer to know what's happening.

* iOS Version: n/a
* SDL iOS Version: v6.2.1
* Testing Against: n/a

joeljfischer:
Fixes #1255 

### Risk

### Testing Plan
Smoke tests performed

### Summary
Throw an exception with detailed information if the lock screen is setup before the app’s window.

##### Bug Fixes
* If SDL is setup before the app's window and root view controller, fix an infinite recursion crash by throwing our own exception with detailed information on how to fix it.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLScreenshotViewController.m
+++ b/SmartDeviceLink/SDLScreenshotViewController.m
@@ -12,9 +12,9 @@
 
 @interface SDLScreenshotViewController ()
 
-    @property (nonatomic, strong) UIImageView *imageView;
+@property (nonatomic, strong) UIImageView *imageView;
 
-    @end
+@end
 
 @implementation SDLScreenshotViewController
 
@@ -80,5 +80,4 @@
     self.imageView.image = image;
 }
 
-
-    @end
+@end
--- a/SmartDeviceLink/SDLScreenshotViewController.m
+++ b/SmartDeviceLink/SDLScreenshotViewController.m
@@ -8,11 +8,13 @@
 
 #import "SDLScreenshotViewController.h"
 
+#import "SDLError.h"
+
 @interface SDLScreenshotViewController ()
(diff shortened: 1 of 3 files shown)


SmartDeviceLink/SDLScreenshotViewController.m
```diff
@@ -30,18 +32,20 @@ - (instancetype)init {
...
 }
 
-// HAX: https://github.com/smartdevicelink/sdl_ios/issues/1250
+    // HAX: https://github.com/smartdevicelink/sdl_ios/issues/1250
 - (UIInterfaceOrientationMask)supportedInterfaceOrientations {
     UIViewController *viewController = [self sdl_topMostControllerForWindow:[UIApplication sharedApplication].windows[0]];
 
-    if (viewController != nil) {
+    if (viewController == self) {
+        @throw [NSException sdl_invalidLockscreenSetupException];
```
NicoleYarroch: Why not just assert and return a default value? Isn't the orientation only an issue for navigation apps using `CarWindow`. Seems kind of pointless to crash a developer's non-navigation app for api that doesn't affect their app.

PR #1262, opened 2019-05-14
Deprecate RPC superclass initializers

### LINKED ISSUE #1204: Make SDLRPCMessage + Superclass initializers unavailable to developers
### Bug Report
Initializers are currently available on all RPC types that may be confusing to users such as `initWithName:`

This would be a major version change and would require a proposal. However, we could deprecate them in the meantime.

##### Reproduction Steps
n/a

* iOS Version: n/a
* SDL iOS Version: 6.2-develop
* Testing Against: n/a

joeljfischer:
Fixes #1204 

### Risk

### Testing Plan
Unit testing and smoke testing

### Summary
This PR deprecates RPC superclass initializers so that they are not clearly available for developers to use, which can be confusing. Ideally, these methods would be "protected" methods and not be available at all, but because they were publicly available, they have to be deprecated until the next major version.

##### Bug Fixes
* Deprecate RPC superclass methods to reduce confusion for developers.

### Tasks Remaining:
- [x] Fix unit tests
- [x] Smoke test

### CLA

### DIFF
--- a/SmartDeviceLink/SDLRPCMessage.h
+++ b/SmartDeviceLink/SDLRPCMessage.h
@@ -7,11 +7,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface SDLRPCMessage : SDLRPCStruct <NSCopying> {
-    NSMutableDictionary<NSString *, id> *function;
-    NSMutableDictionary<NSString *, id> *parameters;
-    NSString *messageType;
-}
+@interface SDLRPCMessage : SDLRPCStruct <NSCopying>
 
 /**
  *  Convenience init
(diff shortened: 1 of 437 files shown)


SmartDeviceLinkTests/RPCSpecs/SuperclassSpecs/SDLRPCMessageSpec.m
```diff
@@ -33,62 +39,86 @@
...
     it(@"Should get correctly when initialized", ^ {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
         SDLRPCMessage* testMessage = [[SDLRPCMessage alloc] initWithDictionary:[@{SDLRPCParameterNameResponse:
                                                                                       @{SDLRPCParameterNameParameters:
                                                                                             @{@"age":@25},
                                                                                         SDLRPCParameterNameOperationName:@"Nameless"}} mutableCopy]];
+#pragma clang diagnostic pop
         
-        expect([testMessage getParameters:@"age"]).to(equal(@25));
```
NicoleYarroch: Leave the test methods for `getParameters` since it is still accessible

SmartDeviceLink/SDLRPCMessage.h
```diff
@@ -69,6 +57,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (strong, nonatomic, readonly) NSString *name;
 
+/**
+ The JSONRPC parameters
```
NicoleYarroch: JSONRPC should be JSON RPC

PR #1277, opened 2019-05-23
Fixed file manager race condition

### LINKED ISSUE #1264: 'SDLManagerDelegate`s are not called if the data session is established successfully but the lifecycle manager does not transition to state ready
### Bug Report
The app's `SDLManagerDelegate`s are not called if the data session is established successfully but the lifecycle manager never ends up in state `SDLLifecycleStateReady`. 

This appears to be some sort of race condition in the lifecycle manager. The lifecycle manager was stuck in state `SDLLifecycleStateSettingUpManagers` so even though Core was sending notifications the app did not get the notifications because the lifecycle manager never finished setting up. 

##### Reproduction Steps
1. Race condition. I only had it happen once.  

##### Expected Behavior
If the lifecycle manager errors out when setting up, the data session should be destroyed and re-created.

##### Observed Behavior
The data session stays open, but the app does not get notifications from Core. This means the SDL app UI is empty. 

* iOS Version: 12.1.4
* SDL iOS Version: 6.2
* Testing Against: SYNC 3.4 

Swift example app in the SDL_iOS library

NicoleYarroch:
Fixes #1264 

### Risk

### Testing Plan
* Unit tests added

### Summary
* Fixed the file manager adding file uploads to the transaction queue after the file manager transitioned to state shutdown. The operation would never finish so any uploads added after the file manager transitioned back to the ready state would not run. 

##### Breaking Changes
* n/a

##### Bug Fixes
* Fixed the file manager adding file uploads to the transaction queue after the file manager has shutdown. 

### CLA

### DIFF
--- a/SmartDeviceLink/SDLFileManager.m
+++ b/SmartDeviceLink/SDLFileManager.m
@@ -404,8 +404,6 @@ SDLFileManagerState *const SDLFileManagerStateStartupError = @"StartupError";
             if ([self sdl_canFileBeUploadedAgain:file maxUploadCount:maxUploadCount failedFileUploadsCount:self.failedFileUploadsCount]) {
                 SDLLogD(@"Attempting to resend file with name %@ after a failed upload attempt", file.name);
                 return [self sdl_uploadFile:file completionHandler:handler];
-            } else {
-                SDLLogE(@"File named %@ failed to upload. Max number of upload attempts reached", file.name);
             }
         }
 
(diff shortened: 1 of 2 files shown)


SmartDeviceLink/SDLFileManager.m
```diff
@@ -526,11 +524,28 @@ - (float)sdl_uploadPercentage:(float)totalBytes uploadedBytes:(float)uploadedByt
  *  @return                 True if the file still needs to be (re)sent to Core; false if not.
  */
 - (BOOL)sdl_canFileBeUploadedAgain:(nullable SDLFile *)file maxUploadCount:(UInt8)maxUploadCount failedFileUploadsCount:(NSMutableDictionary<SDLFileName *, NSNumber<SDLUInt> *> *)failedFileUploadsCount {
-    if (!file || [self hasUploadedFile:file]) {
+    if (![self.currentState isEqualToString:SDLFileManagerStateReady]) {
+        SDLLogE(@"File named %@ failed to upload. The file manager has shutdown so the file will not be sent again.", file.name);
```
joeljfischer: Make this a warning. Change string to `File named %@ failed to upload. The file manager has shutdown so the file upload will not retry.`

SmartDeviceLink/SDLFileManager.m
```diff
@@ -526,11 +524,28 @@ - (float)sdl_uploadPercentage:(float)totalBytes uploadedBytes:(float)uploadedByt
...
+        return NO;
+    }
+
+    if (!file) {
+        SDLLogE(@"File named %@ can not be uploaded because it is not a valid file.", file.name);
```
joeljfischer: The `%@, file.name` won't work here because the file doesn't exist!

SmartDeviceLink/SDLFileManager.m
```diff
@@ -526,11 +524,28 @@ - (float)sdl_uploadPercentage:(float)totalBytes uploadedBytes:(float)uploadedByt
...
+        return NO;
+    }
+
+    if ([self hasUploadedFile:file]) {
+        SDLLogE(@"File named %@ has already been uploaded.", file.name);
```
joeljfischer: This should be a debug message, maybe a warning at the most.

SmartDeviceLink/SDLFileManager.m
NicoleYarroch: Fixed

SmartDeviceLink/SDLFileManager.m
NicoleYarroch: Fixed

SmartDeviceLink/SDLFileManager.m
NicoleYarroch: Fixed

PR #1295, opened 2019-06-05
Fixed the soft button operation not finishing if no image or text buttons sent

### LINKED ISSUE #1294: SDLSoftButtonReplaceOperation finish operation not called if no image buttons or text buttons uploaded
### Bug Report
The `SDLSoftButtonReplaceOperation` is never finished if the head unit does not support soft button images and some of the soft buttons do not have text (i.e. they are image only). So if other operations are added to the `transactionQueue` in the `SDLSoftButtonManager`, they never execute.

##### Reproduction Steps
1. Connect the Swift sdl_ios example app to Manticore via TCP transport
1. Make the `hmiLevel` `FULL` by selecting the app icon

##### Expected Behavior
All the soft buttons appear. 

##### Observed Behavior
No soft buttons ever appear.
What happens is that:
1. The RAI `displayCapabilities` is `nil` so no soft buttons are sent in the first `SDLSoftButtonReplaceOperation` added to the queue. This is because the example app has image-only soft buttons. 
1. When the app sends a `SetDisplayLayout` request, the soft button manager adds a new `SDLSoftButtonReplaceOperation` to the transaction queue with the new display capabilities but the operation never executes.

* iOS Version: n/a
* SDL iOS Version: 6.2
* Testing Against: Manticore

Swift sdl_ios example app.

NicoleYarroch:
Fixes #1294 

### Risk

### Testing Plan
Test cases added

### Summary
The `SDLSoftButtonReplaceOperation` is never finished if the head unit does not support soft button images and some of the soft buttons do not have text (i.e. they are image only). So if other operations are added to the `transactionQueue` in the `SDLSoftButtonManager`, they never execute.

##### Bug Fixes
* Fixed the soft button operation not finishing if no image or text buttons sent by the operation.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLSoftButtonReplaceOperation.m
+++ b/SmartDeviceLink/SDLSoftButtonReplaceOperation.m
@@ -56,11 +56,17 @@ NS_ASSUME_NONNULL_BEGIN
         SDLLogW(@"Soft button images are not supported. Attempting to send text-only soft buttons. If any button does not contain text, no buttons will be sent.");
 
         // Send text buttons if all the soft buttons have text
-        [self sdl_sendCurrentStateTextOnlySoftButtonsWithCompletionHandler:^{}];
+        __weak typeof(self) weakself = self;
+        [self sdl_sendCurrentStateTextOnlySoftButtonsWithCompletionHandler:^(BOOL success) {
+            if (!success) {
+                SDLLogE(@"Head unit does not support images and some of the soft buttons do not have text, so none of the buttons will be sent.");
+                [weakself finishOperation];
+            }
+        }];
     } else if ([self sdl_currentStateHasImages] && ![self sdl_allCurrentStateImagesAreUploaded]) {
         // If there are images that aren't uploaded
         // Send text buttons if all the soft buttons have text
-        [self sdl_sendCurrentStateTextOnlySoftButtonsWithCompletionHandler:^{}];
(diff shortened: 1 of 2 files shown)


SmartDeviceLink/SDLSoftButtonReplaceOperation.m
```diff
@@ -56,11 +56,17 @@ - (void)start {
         SDLLogW(@"Soft button images are not supported. Attempting to send text-only soft buttons. If any button does not contain text, no buttons will be sent.");
 
         // Send text buttons if all the soft buttons have text
-        [self sdl_sendCurrentStateTextOnlySoftButtonsWithCompletionHandler:^{}];
+        __weak typeof(self) weakself = self;
+        [self sdl_sendCurrentStateTextOnlySoftButtonsWithCompletionHandler:^(BOOL success) {
+            if (!success) {
+                SDLLogE(@"Head unit does not support images and some of the soft buttons do not have text, so none of the buttons will be sent.");
```
joeljfischer: Isn't this redundant with the warning at L224?

NicoleYarroch: I don't think so. The warning on L224 can be benign if the head unit supports images but the library can't send the text only buttons because not all buttons have text (i.e. eventually the image + text buttons will be sent). This error is only printed if the head unit does not support images and the library can't send text-only buttons.

PR #1317, opened 2019-06-24
Fixed session creation failing due to no background task

### LINKED ISSUE #1316: A new transport session is not created if a background task is not ongoing
### Bug Report
If app on device is closed when trying to create a connection with the SDL enabled accessory: 
1. The connection attempt can fail if the background task is not started.
2. The reconnection attempt can fail due to the background task being destroyed in the dealloc method of the `SDLIAPTransport` object. 

##### Reproduction Steps
1. Configure 13 apps with v.6.3 of the library 
2. Connect the phone to SYNC 3.4 TDK with BT
3. Go to apps tab and wait for all 13 apps to show up
  
##### Expected Behavior
All 13 apps show up on the SDL HMI.

##### Observed Behavior
Between 1 and 3 apps will sometimes fail to show up. From the logs it looks like the background task expires, so even though a reconnection attempt should be made by the lifecycle manager, nothing happens. 

* iOS Version: 12.1.4
* SDL iOS Version: 6.3
* Testing Against: SYNC 3.4 TDK (Build 19126_DEVTEST) + iPhone XS Max (iOS 12.1.4)

NicoleYarroch:
Fixes #1316

### Risk

### Testing Plan
* Smoke tests done with 13 apps

### Summary
1. If the app is not in the foreground when the `SDLIAPTransport` is created, a background task is now started.
2. The background task is no longer destroyed when the transport is deallocated. If the app on the device is not open, this stopped reconnection attempts since code execution stops when the background task is destroyed. 
3. Removed unnecessary listening for `EAAccessory` connections in the `SDLProxy` class.

##### Bug Fixes
* Fixed new transport sessions not being created if a background task is not ongoing.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLIAPTransport.m
+++ b/SmartDeviceLink/SDLIAPTransport.m
@@ -55,6 +55,11 @@ int const CreateSessionRetries = 3;
     // Get notifications if an accessory connects in future
     [self sdl_startEventListening];
 
+    if (UIApplication.sharedApplication.applicationState != UIApplicationStateActive) {
+        SDLLogV(@"Application not in foreground on transport init. Starting a background task");
+        [self sdl_backgroundTaskStart];
+    }
+
     // Wait for setup to complete before scanning for accessories
 
     return self;
@@ -63,7 +68,7 @@ int const CreateSessionRetries = 3;
 #pragma mark - Background Task
 
 /**
- *  Starts a background task that allows the app to search for accessories and while the app is in the background.
+ *  Starts a background task that allows the app to search for accessories while the app is in the background.
  */
 - (void)sdl_backgroundTaskStart {
     if (self.backgroundTaskId != UIBackgroundTaskInvalid) {
@@ -77,7 +82,7 @@ int const CreateSessionRetries = 3;
         [weakself sdl_backgroundTaskEnd];
     }];
 
-    SDLLogD(@"Started a background task with id: %lu", (unsigned long)self.backgroundTaskId);
+    SDLLogD(@"Background task started with id: %lu", (unsigned long)self.backgroundTaskId);
 }
 
 /**
--- a/SmartDeviceLink/SDLIAPTransport.m
+++ b/SmartDeviceLink/SDLIAPTransport.m
@@ -15,7 +15,6 @@
 #import "SDLIAPDataSession.h"
 #import "SDLIAPDataSessionDelegate.h"
 #import "SDLLogMacros.h"
-#import "SDLTimer.h"
 #import <CommonCrypto/CommonDigest.h>
 
 NS_ASSUME_NONNULL_BEGIN
@@ -81,7 +80,7 @@ int const CreateSessionRetries = 3;
  */
 - (void)sdl_stopEventListening {
     SDLLogV(@"SDLIAPTransport stopped listening for events");
-    [[NSNotificationCenter defaultCenter] removeObserver:self];
+    [[EAAccessoryManager sharedAccessoryManager] unregisterForLocalNotifications];
 }
 
 #pragma mark EAAccessory Notifications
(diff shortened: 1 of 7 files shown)


SmartDeviceLink/SDLIAPTransport.m
```diff
@@ -63,20 +68,21 @@ - (instancetype)init {
...
     if (self.backgroundTaskId != UIBackgroundTaskInvalid) {
         SDLLogV(@"A background task is already running. No need to start a background task.");
         return;
     }
 
+    __weak typeof(self) weakself = self;
     self.backgroundTaskId = [[UIApplication sharedApplication] beginBackgroundTaskWithName:BackgroundTaskName expirationHandler:^{
         SDLLogD(@"Background task expired");
-        [self sdl_backgroundTaskEnd];
+        [weakself sdl_backgroundTaskEnd];
```
joeljfischer: This, combined with removing it from `dealloc` would cause the background task to never end if `SDLIAPTransport` is deallocated. If we don't call `sdl_backgroundTaskEnd`, then iOS will discover that we don't end our background tasks and will start preventing us from using background time. We need to call `sdl_backgroundTaskEnd` when we're close to expiring.

We may need to move this to another class that won't be dealloced.

NicoleYarroch: Fixed. Moved the background task logic to the lifecycle manager.

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -219,6 +222,9 @@ - (void)didEnterStateStarted {
                                             tcpPort:@(self.configuration.lifecycleConfig.tcpDebugPort).stringValue
                           secondaryTransportManager:self.secondaryTransportManager];
     } else {
+        // start a background task so a session can be established even when the app is backgrounded.
+        [self sdl_backgroundTaskStart];
```
joeljfischer: I think this should always run when we are running SDL, not just in iAP.

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -829,6 +841,50 @@ - (void)videoServiceProtocolDidUpdateFromOldProtocol:(nullable SDLProtocol *)old
     }
 }
 
+#pragma mark - Background Task
```
joeljfischer: I'm going back and forth on whether this is code bloat. I think this code (and the property) should be in a new class that only manages the background task. It can be then held as a property by the lifecycle manager and started and stopped when the lifecycle manager is started and stopped.

SmartDeviceLink/SDLManager.m
```diff
@@ -164,6 +164,12 @@ - (void)unsubscribeFromRPC:(SDLNotificationName)rpcName withObserver:(id)observe
     [[NSNotificationCenter defaultCenter] removeObserver:observer name:rpcName object:nil];
 }
 
+#pragma mark - Lifecycle Destruction
+
+- (void)dealloc {
+    _lifecycleManager = nil;
```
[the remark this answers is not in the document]
NicoleYarroch: Fixed. Removed

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -72,6 +72,8 @@
 SDLLifecycleState *const SDLLifecycleStateUnregistering = @"Unregistering";
 SDLLifecycleState *const SDLLifecycleStateReady = @"Ready";
 
+NSString *const BackgroundTaskIAPTransportName = @"com.sdl.transport.iap.backgroundTask";
```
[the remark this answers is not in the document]
NicoleYarroch: Fixed. Renamed

SmartDeviceLink/SDLLifecycleManager.m
NicoleYarroch: Made new class called `SDLBackgroundTaskManager`

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -829,6 +843,12 @@ - (void)videoServiceProtocolDidUpdateFromOldProtocol:(nullable SDLProtocol *)old
     }
 }
 
+#pragma mark - Lifecycle Destruction
+
+- (void)dealloc {
+    [self.backgroundTaskManager endBackgroundTask];
```
joeljfischer: I think this can just be moved to the `backgroundTaskManager` dealloc.

SmartDeviceLink/SDLLifecycleManager.m
NicoleYarroch: Fixed

SmartDeviceLink/SDLLifecycleManager.m
NicoleYarroch: Fixed

Issue #1163, opened 2019-02-19
SDL 0207 - RPC message protection


jordynmackool:
Proposal: [SDL 0207 - RPC message protection](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0207-rpc-message-protection.md)

> This proposal proposes a design to protect (enable encryption for) RPC messages transmitted between a mobile application and the SDL.

https://github.com/smartdevicelink/sdl_evolution/issues/634

Steering Committee Decision:

>The Steering Committee voted to accept this proposal with revisions. The revisions will include adding a flag in the policy table per app that describes if they are required to comply with the encryption or not as noted in [this comment.](https://github.com/smartdevicelink/sdl_evolution/issues/634#issuecomment-460710662)
There was additional discussion in regards to whether or not this encryption will be used in core from the policy table, essentially authentication and validation or if this flag needs to be passed from core into the proxies. It was determined that it is almost inferred by the proxies if they get an on-permission change with the new RPCs that are delegating they need to be encrypted and the app itself would need encryption, so core should send the on-permission changes to an app in regards to the RPC encryptions unless that app requires an encryption.   

Revisions were made on 2.19.19

SatbirTanda:
Will begin implementing this proposal

PR #1320, opened 2019-06-25
Implement SDL-0207 RPC Message Protection

SatbirTanda:
Implements #1163 

This PR is ready for review.

### Risk
This PR makes minor API changes.

### Testing Plan
Add unit tests

### Summary
Add a public payloadProtected property to RPC Messages.  Also automatically check the requireEncryption flag of RPC being sent in the sendRPC method on SDLManager for automatically handling encryption if needed.

### CLA


SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -703,17 +769,26 @@ - (nullable NSString *)authToken {
...
+        return self.permissionManager.permissions[request.name].requireEncryption;
+    }
+    return NO;
+}
+
 #pragma mark SDL notification observers
 
 - (void)transportDidConnect {
-    SDLLogD(@"Transport connected");
+    if (![self.lifecycleStateMachine isCurrentState:SDLLifecycleStateReady]) {
```
joeljfischer: This doesn't appear to be related to this PR. Please remove.

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -843,6 +918,19 @@ - (void)videoServiceProtocolDidUpdateFromOldProtocol:(nullable SDLProtocol *)old
     }
 }
 
+- (void)encryptionServiceProtocolDidUpdateFromOldProtocol:(nullable SDLProtocol *)oldProtocol toNewProtocol:(nullable SDLProtocol *)newProtocol {
```
joeljfischer: This seems odd to me. Why wouldn't it just be stopped when the lifecycle manager stops, and started when it's started again?

SmartDeviceLink/SDLManager.h
```diff
@@ -54,6 +55,11 @@ typedef void (^SDLManagerReadyBlock)(BOOL success, NSError *_Nullable error);
  */
 @property (copy, nonatomic, readonly) SDLSystemContext systemContext;
 
+/**
+ *  The encryption manager to be used by the running app.
+ */
+@property (copy, nonatomic, readonly) SDLEncryptionManager *encryptionManager;
```
joeljfischer: This doesn't seem to be necessary.

SmartDeviceLink/SDLManager.m
```diff
@@ -76,6 +76,10 @@ - (nullable SDLHMILevel)hmiLevel {
     return self.lifecycleManager.hmiLevel;
 }
 
+- (SDLEncryptionManager *)encryptionManager {
```
joeljfischer: See above

SmartDeviceLink/SDLOnPermissionsChange.h
```diff
@@ -22,6 +22,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (strong, nonatomic) NSArray<SDLPermissionItem *> *permissionItem;
 
+/**
+ Describes whether or not the app needs the encryption permission
+ 
+ Optional boolean available since core 5.1
```
joeljfischer: This should reference the RPC spec, not Core. Also I don't think 5.1 is correct.

SmartDeviceLink/SDLPermissionManager.h
```diff
@@ -18,6 +18,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface SDLPermissionManager : NSObject
 
+@property (strong, nonatomic, readonly) NSMutableDictionary<SDLPermissionRPCName, SDLPermissionItem *> *permissions;
+
+@property (assign, nonatomic, readonly) BOOL requiresEncryption;
+
+@property (copy, nonatomic, nullable, readonly) SDLHMILevel currentHMILevel;
```
joeljfischer: Why is this necessary to be public?

SmartDeviceLink/SDLPermissionManager.h
```diff
@@ -18,6 +18,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface SDLPermissionManager : NSObject
 
+@property (strong, nonatomic, readonly) NSMutableDictionary<SDLPermissionRPCName, SDLPermissionItem *> *permissions;
```
joeljfischer: Why was this made public? It doesn't seem to be necessary.

SmartDeviceLink/SDLPermissionManager.h
```diff
@@ -18,6 +18,12 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+@property (assign, nonatomic, readonly) BOOL requiresEncryption;
```
joeljfischer: Documentation is needed for all public methods and properties. What does this do and why is it here?

SmartDeviceLink/SDLProtocol.m
```diff
@@ -165,7 +165,7 @@ - (SDLProtocolMessage *)sdl_createStartServiceMessageWithType:(SDLServiceType)se
     switch (serviceType) {
         case SDLServiceTypeRPC: {
             // Need a different header for starting the RPC service, we get the session Id from the HU, or its the same as the RPC service's
-            header = [SDLProtocolHeader headerForVersion:1];
+            header = [SDLProtocolHeader headerForVersion:3];
```
joeljfischer: This is incorrect according to the protocol spec it must be a version 1 header.

SmartDeviceLink/SDLProtocol.m
```diff
@@ -313,10 +317,21 @@ - (BOOL)sendRPC:(SDLRPCMessage *)message encrypted:(BOOL)encryption error:(NSErr
 
             // If we're trying to encrypt, try to have the security manager encrypt it. Return if it fails.
             // TODO: (Joel F.)[2016-02-09] We should assert if the service isn't setup for encryption. See [#350](https://github.com/smartdevicelink/sdl_ios/issues/350)
-            messagePayload = encryption ? [self.securityManager encryptData:rpcPayload.data withError:error] : rpcPayload.data;
+            if (encryption) {
```
joeljfischer: Why not just use the error above? Why was a new error added?

SmartDeviceLink/SDLProxy.m
```diff
@@ -294,6 +294,22 @@ - (void)sendRPC:(SDLRPCMessage *)message {
     }
 }
 
+- (void)sendEncryptedRPC:(SDLRPCMessage *)message {
```
joeljfischer: This duplicates the code in `sendRPC`, it should be refactored instead.

SmartDeviceLink/SDLProxy.m
```diff
@@ -294,6 +294,22 @@ - (void)sendRPC:(SDLRPCMessage *)message {
...
+    if ([message.getFunctionName isEqualToString:@"SubscribeButton"]) {
```
joeljfischer: This throws warnings

SmartDeviceLink/SDLProxy.m
```diff
@@ -371,6 +387,83 @@ - (BOOL)sdl_adaptButtonUnsubscribeMessage:(SDLUnsubscribeButton *)message {
 }
 #pragma clang diagnostic pop
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+- (BOOL)sdl_adaptButtonSubscribeMessageEncrypted:(SDLSubscribeButton *)message {
```
joeljfischer: Code duplication again.

SmartDeviceLink/SDLProxy.m
```diff
@@ -371,6 +387,83 @@ - (BOOL)sdl_adaptButtonUnsubscribeMessage:(SDLUnsubscribeButton *)message {
...
+            }
+            
+            return YES;
+        }
+    }
+    
+    return NO;
+}
+
+- (BOOL)sdl_adaptButtonUnsubscribeMessageEncrypted:(SDLUnsubscribeButton *)message {
```
joeljfischer: Code duplication

SmartDeviceLinkTests/DevAPISpecs/SDLConfigurationSpec.m
```diff
@@ -52,7 +52,7 @@
         });
 
         it(@"initWithLifecycle:lockScreen:logging:fileManager:", ^{
```
joeljfischer: The encryption manager stuff and the lifecycle manager all needs tests

SmartDeviceLink/SDLLifecycleManager.m
SatbirTanda: this is important because when a secure service is started for RPC type, it will recall all the methods when a normal RPC service is started (including `onProxyOpened` and `transportDidConnect`).  actually that fact is the main reason for creating a new `SDLEncryptionManager` and `SDLEncryptionLifecycleManager` class.  we must wait for `handleProtocolStartServiceACKMessage` after `startSecureServiceWithType` is called and this can only be handled properly if it is called after the initial unsecured service is started.  I was thinking we can have all the encryptions handled in EncryptionManager class RPC, Audio, Video to handle it nicely with the `SDLStreamingMediaManager`, but another option is to add `encrypted` flags to a lot of the methods.  I tried both, in the latter I noticed that it seemed to change more files than necessary

joeljfischer: The methods that should be changed for sending encrypted / non-encrypted RPCs only needs to be `SDLProtocol` right? It just needs to know if the service that it's being sent on is encrypted and then it can encrypt it right before it sends the data off to the transport.

Having the encryption lifecycle manager handle the setting up of the encryption seems like it would be okay, but I don't see any reason for it to be public, and it would just be a factoring out of the `SDLLifecycleManager`'s setup and teardown process.

SmartDeviceLink/SDLProtocol.m
SatbirTanda: When sent with headerVersion 1 SYNC sends a NACK.  According the Luxoft, it should be 3 or above, this result in ACK in testing

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -296,6 +306,8 @@ - (void)didEnterStateConnected {
     if (self.configuration.streamingMediaConfig.securityManagers != nil) {
         SDLLogD(@"Adding security managers");
         [self.proxy addSecurityManagers:self.configuration.streamingMediaConfig.securityManagers forAppId:self.configuration.lifecycleConfig.appId];
+    } else if (self.configuration.encryptionConfig.securityManagers != nil) {
+        [self.proxy addSecurityManagers:self.configuration.encryptionConfig.securityManagers forAppId:self.configuration.lifecycleConfig.appId];
```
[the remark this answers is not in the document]
SatbirTanda: Actually this checks to make sure that if a security manger is passed in from `streamingMediaConfig`, then we do not need to add the one from `encryptionCong`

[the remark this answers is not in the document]
joeljfischer: I see what you're saying, but that's actually _more_ incorrect. Imagine for a moment if they implement a streaming security manager for Toyota vehicles, and an RPC security manager for Ford vehicles. The Ford one will not be applied.

theresalech:
@SatbirTanda can you please advise if this is ready for re-review?

SatbirTanda:
@theresalech yes ready for review, we are still testing/adding test for verification with Luxoft

theresalech:
@SatbirTanda please advise when Ford has completed their testing.  We'll then perform our review once Ford has confirmed their testing is complete and the feature has been approved on your side.  Let me know if you have any questions about this!

SatbirTanda:
@theresalech Ford can confirm the RPC encryption is working.  We tested using SDL_Core in a docker container with proper certificates.

SmartDeviceLink/SDLProtocol.m
joeljfischer: I'm pretty sure that this will drop support for protocol v1 and v2 head units.

SmartDeviceLink/SDLProtocol.m
```diff
@@ -313,10 +313,21 @@ - (BOOL)sendRPC:(SDLRPCMessage *)message encrypted:(BOOL)encryption error:(NSErr
...
+                NSError *encryptError = nil;
```
joeljfischer: I think we can just reuse the above error which will be passed out if it fails.

SmartDeviceLink/SDLProxy.m
```diff
@@ -174,7 +174,7 @@ - (void)sdl_sendMobileHMIState {
     }
 
     SDLLogD(@"Mobile UIApplication state changed, sending to remote system: %@", HMIStatusRPC.hmiLevel);
-    [self sendRPC:HMIStatusRPC];
+    [self sendRPC:HMIStatusRPC withEncryption:NO];
```
joeljfischer: Anything with `NO` here and below will fail if encryption is required. This is probably more difficult to fix in this class than in the Lifecycle Manager because this class doesn't check which RPCs require encryption and which don't. You can ping me if you want to talk about how to deal with this.

SmartDeviceLink/SDLSequentialRPCRequestOperation.m
```diff
@@ -81,7 +81,7 @@ - (void)sdl_sendNextRequest {
 
     // Send the next request
     SDLRPCRequest *request = self.requests[self.currentRequestIndex];
-    [self.connectionManager sendConnectionRequest:(__kindof SDLRPCRequest *)request withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
+    [self.connectionManager sendConnectionRequest:(__kindof SDLRPCRequest *)request withEncryption:NO withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
```
joeljfischer: This will cause them to always fail if encryption is required.

SmartDeviceLink/SDLSoftButtonReplaceOperation.m
```diff
@@ -198,7 +198,7 @@ - (void)sdl_sendCurrentStateSoftButtonsWithCompletionHandler:(void (^)(void))han
     show.mainField1 = self.mainField1;
     show.softButtons = [softButtons copy];
 
-    [self.connectionManager sendConnectionRequest:show withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
+    [self.connectionManager sendConnectionRequest:show withEncryption:NO withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
```
joeljfischer: This will always fail if encryption is required.

SmartDeviceLink/SDLSoftButtonTransitionOperation.m
```diff
@@ -53,7 +53,7 @@ - (void)sdl_sendNewSoftButtons {
     newShow.mainField1 = self.mainField1;
     newShow.softButtons = [self sdl_currentStateSoftButtonsForObjects:self.softButtons];
 
-    [self.connectionManager sendConnectionRequest:newShow withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
+    [self.connectionManager sendConnectionRequest:newShow withEncryption:NO withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
```
joeljfischer: This will always fail if encryption is required

SmartDeviceLink/SDLSystemCapabilityManager.m
```diff
@@ -267,7 +267,7 @@ - (void)sdl_subscribeToSystemCapabilityUpdates {
  */
 - (void)sdl_sendGetSystemCapability:(SDLGetSystemCapability *)getSystemCapability completionHandler:(nullable SDLUpdateCapabilityHandler)handler {
     __weak typeof(self) weakSelf = self;
-    [self.connectionManager sendConnectionRequest:getSystemCapability withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
+    [self.connectionManager sendConnectionRequest:getSystemCapability withEncryption:NO withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
```
joeljfischer: This will always fail if encryption is required

SmartDeviceLink/SDLTextAndGraphicManager.m
```diff
@@ -205,7 +205,7 @@ - (void)sdl_updateWithCompletionHandler:(nullable SDLTextAndGraphicUpdateComplet
 
     if (self.inProgressUpdate == nil) { return; }
 
-    [self.connectionManager sendConnectionRequest:self.inProgressUpdate withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
+    [self.connectionManager sendConnectionRequest:self.inProgressUpdate withEncryption:NO withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
```
joeljfischer: This will always fail if encryption is required

SmartDeviceLinkTests/RPCSpecs/NotificationSpecs/SDLOnPermissionsChangeSpec.m
```diff
@@ -15,34 +15,39 @@
...
+SDLPermissionItem *item = [[SDLPermissionItem alloc] init];
 
 describe(@"Getter/Setter Tests", ^ {
     it(@"Should set and get correctly", ^ {
-        SDLOnPermissionsChange* testNotification = [[SDLOnPermissionsChange alloc] init];
+        SDLOnPermissionsChange *testNotification = [[SDLOnPermissionsChange alloc] init];
         
         testNotification.permissionItem = [@[item] mutableCopy];
-        
+        testNotification.requireEncryption = @1;
```
joeljfischer:
```suggestion
        testNotification.requireEncryption = @YES;
```

SmartDeviceLinkTests/RPCSpecs/NotificationSpecs/SDLOnPermissionsChangeSpec.m
```diff
@@ -15,34 +15,39 @@
...
+
         expect(testNotification.permissionItem).to(equal([@[item] mutableCopy]));
+        expect(testNotification.requireEncryption).to(beTrue());
```
joeljfischer:
```suggestion
        expect(testNotification.requireEncryption.boolValue).to(beTrue());
```

SmartDeviceLinkTests/RPCSpecs/NotificationSpecs/SDLOnPermissionsChangeSpec.m
```diff
@@ -15,34 +15,39 @@
...
     });
     
     it(@"Should get correctly when initialized", ^ {
         NSMutableDictionary<NSString *, id> *dict = [@{SDLRPCParameterNameNotification:
                                                            @{SDLRPCParameterNameParameters:
-                                                                 @{SDLRPCParameterNamePermissionItem:[@[item] mutableCopy]},
+                                                                 @{SDLRPCParameterNamePermissionItem:[@[item] mutableCopy],
```
joeljfischer:
```suggestion
                                                                 @{SDLRPCParameterNamePermissionItem:@[item],
```

Mutability should not longer be required.

SmartDeviceLinkTests/RPCSpecs/NotificationSpecs/SDLOnPermissionsChangeSpec.m
```diff
@@ -15,34 +15,39 @@
...
+                                                                   SDLRPCParameterNameRequireEncryption:@1},
```
joeljfischer:
```suggestion
                                                                   SDLRPCParameterNameRequireEncryption: @YES},
```

SmartDeviceLinkTests/RPCSpecs/StructSpecs/SDLPermissionItemSpec.m
```diff
@@ -15,42 +15,47 @@
...
 
 describe(@"Getter/Setter Tests", ^ {
     it(@"Should set and get correctly", ^ {
-        SDLPermissionItem* testStruct = [[SDLPermissionItem alloc] init];
+        SDLPermissionItem *testStruct = [[SDLPermissionItem alloc] init];
         
         testStruct.rpcName = @"RPCNameThing";
         testStruct.hmiPermissions = hmiPermissions;
         testStruct.parameterPermissions = parameterPermissions;
+        testStruct.requireEncryption = @1;
```
joeljfischer:
```suggestion
        testStruct.requireEncryption = @YES;
```

SmartDeviceLinkTests/RPCSpecs/StructSpecs/SDLPermissionItemSpec.m
```diff
@@ -15,42 +15,47 @@
...
         
         expect(testStruct.rpcName).to(equal(@"RPCNameThing"));
         expect(testStruct.hmiPermissions).to(equal(hmiPermissions));
         expect(testStruct.parameterPermissions).to(equal(parameterPermissions));
+        expect(testStruct.requireEncryption).to(beTrue());
```
joeljfischer:
```suggestion
        expect(testStruct.requireEncryption.boolValue).to(beTrue());
```

SmartDeviceLinkTests/RPCSpecs/StructSpecs/SDLPermissionItemSpec.m
```diff
@@ -15,42 +15,47 @@
...
     });
     
     it(@"Should get correctly when initialized", ^ {
         NSMutableDictionary<NSString *, id> *dict = [@{SDLRPCParameterNameRPCName:@"RPCNameThing",
                                                        SDLRPCParameterNameHMIPermissions:hmiPermissions,
-                                                       SDLRPCParameterNameParameterPermissions:parameterPermissions} mutableCopy];
+                                                       SDLRPCParameterNameParameterPermissions:parameterPermissions,
+                                                       SDLRPCParameterNameRequireEncryption:@1} mutableCopy];
```
joeljfischer:
```suggestion
                                                       SDLRPCParameterNameRequireEncryption: @YES};
```

SmartDeviceLinkTests/RPCSpecs/StructSpecs/SDLPermissionItemSpec.m
```diff
@@ -15,42 +15,47 @@
...
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
-        SDLPermissionItem* testStruct = [[SDLPermissionItem alloc] initWithDictionary:dict];
+        SDLPermissionItem *testStruct = [[SDLPermissionItem alloc] initWithDictionary:dict];
 #pragma clang diagnostic pop
         
         expect(testStruct.rpcName).to(equal(@"RPCNameThing"));
         expect(testStruct.hmiPermissions).to(equal(hmiPermissions));
         expect(testStruct.parameterPermissions).to(equal(parameterPermissions));
+        expect(testStruct.requireEncryption).to(beTrue());
```
joeljfischer:
```suggestion
        expect(testStruct.requireEncryption.boolValue).to(beTrue());
```

SmartDeviceLinkTests/DevAPISpecs/SDLPermissionsManagerSpec.m
```diff
@@ -19,6 +19,7 @@ @interface SDLPermissionManager ()
 @property (strong, nonatomic) NSMutableDictionary<SDLPermissionRPCName, SDLPermissionItem *> *permissions;
 @property (strong, nonatomic) NSMutableArray<SDLPermissionFilter *> *filters;
 @property (copy, nonatomic, nullable) SDLHMILevel currentHMILevel;
+@property (assign, nonatomic) BOOL requiresEncryption;
```
joeljfischer: This should be used somewhere in the spec.

theresalech:
@SatbirTanda can you please let us know when this PR is ready for re-review?  Thanks!

SatbirTanda:
@theresalech ready for review

joeljfischer:
I think we need to take a step back here; we need a consistent solution. I took a look at the Android PR for this feature and was troubled by the differences. I'll be meeting with our Android team soon to provide consistent direction to both yourself and the Android PR author.

SmartDeviceLinkTests/RPCSpecs/StructSpecs/SDLPermissionItemSpec.m
joeljfischer: This is not entirely fixed (you don't need a mutable copy)

joeljfischer:
@SatbirTanda This was not ready for review. Many comments above have not been addressed. Please either respond to them with reasons why the requested change is not required or make the requested changes.

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -299,9 +308,13 @@ - (void)didEnterStateConnected {
 
     // If we have security managers, add them to the proxy
     if (self.configuration.streamingMediaConfig.securityManagers != nil) {
-        SDLLogD(@"Adding security managers");
+        SDLLogD(@"Adding security managers from streamingMedia configuration");
         [self.proxy addSecurityManagers:self.configuration.streamingMediaConfig.securityManagers forAppId:self.configuration.lifecycleConfig.appId];
```
joeljfischer: Even though this is necessary, I think that deprecating taking security managers in the `SDLStreamingMediaConfiguration` is still necessary.

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -619,15 +647,23 @@ - (void)sendConnectionRequest:(__kindof SDLRPCRequest *)request withResponseHand
         return;
     }
 
-    [self sdl_sendRequest:request withResponseHandler:handler];
+    dispatch_async(_lifecycleQueue, ^{
```
joeljfischer: This should only change the encryption BOOL (though it will be on the request itself) to YES from NO, not visa-versa. That case should force a failure through the response handler (you will return `nil` response, the request, and an appropriate new error value).

SmartDeviceLink/SDLEncryptionLifecycleManager.m
```diff
@@ -0,0 +1,241 @@
...
+    SDLLogD(@"Sending secure rpc start service");
+    [self.protocol startSecureServiceWithType:SDLServiceTypeRPC payload:nil tlsInitializationHandler:^(BOOL success, NSError *error) {
+        if (error) {
+            SDLLogE(@"TLS setup error: %@", error);
+            [self.encryptionStateMachine transitionToState:SDLEncryptionLifecycleManagerStateStopped];
+        }
+    }];
+}
+
+- (void)sdl_stopEncryptionService {
```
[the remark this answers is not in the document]
SatbirTanda: this is called in `SDLEncryptionLifecycleManager` `stop` method

SmartDeviceLink/SDLEncryptionLifecycleManager.m
```diff
@@ -0,0 +1,241 @@
...
+    
+    // if startWithProtocol has not been called yet, abort here
+    if (!self.protocol) { return; }
+    
+    if (!self.isEncryptionReady) {
+        [self sdl_startEncryptionService];
+    }
+}
+
+- (void)sdl_permissionsDidChange:(SDLRPCNotificationNotification *)notification {
```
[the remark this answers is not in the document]
SatbirTanda: adding permission manager as oppose to observing this would require us to add a public method to permission manager that lets us see if any RPC contains the `requireEncryption` flag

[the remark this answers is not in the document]
joeljfischer: Why? You're not doing that right now in this manager, unless that's a bug?

SmartDeviceLink/SDLEncryptionLifecycleManager.m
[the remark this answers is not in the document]
joeljfischer: I see that now, it wasn't coming up in Xcode. Still, I don't think that this is correct. If the lifecycle manager is stopped, it should properly unregister. This will just end the service, which isn't correct. The service will be ended elsewhere.

SmartDeviceLink/SDLEncryptionLifecycleManager.m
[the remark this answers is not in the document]
SatbirTanda: yea its a bug - according to the proposal we should wait for HMI to be full, background, limited and we should wait for onpermissionchange to check if at least one RPC needs protection

SmartDeviceLink/SDLEncryptionLifecycleManager.m
```diff
@@ -0,0 +1,241 @@
...
+    
+    [self.rpcOperationQueue addOperation:op];
+}
+
+- (BOOL)isEncryptionReady {
+    return [self.encryptionStateMachine isCurrentState:SDLEncryptionLifecycleManagerStateReady];
+}
+
+- (void)sdl_startEncryptionService {
+    SDLLogV(@"Attempting to start Encryption Service");
```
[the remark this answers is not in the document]
SatbirTanda: since this queue is a reference to the rpc queue in `SDLLifecycleManager` this will pause all RPC operations in SDL. is that ok?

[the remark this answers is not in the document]
joeljfischer: Yeah that should be okay. The best thing to do would be to pause, then check if any of the RPC operations now require encryption once its set up, but that's pretty tricky.

SmartDeviceLink/SDLLifecycleManager.m
SatbirTanda: If the RPC requires encryption and the developer setPayloadProtected to NO then the handler would return error as ENCRYPTION_REQUIRED.   this would cause the developer to want to check whether an RPC needs encryption for each request.  that's why I was thinking this method should always check to if an RPC needs encryption from the permission manager and then change to BOOL from NO to YES

joeljfischer: Ah crap, I miswrote my comment. I meant that it should only change the encryption BOOL from NO to YES.

If the RPC requires encryption from the head unit side, but the developer says NO, then we can assume that the developer doesn't care if it's encrypted or not. But if they say YES, then we have to assume they want to require encryption. If the RPC requires encryption from the head unit side, then we can change it to YES for them to enable the RPC to work. If they set it to YES and we can't encrypt, then we should fail early.

I apologize for the mis-type in my first comment.

SmartDeviceLink/SDLEncryptionLifecycleManager.m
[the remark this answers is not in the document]
SatbirTanda: another reason why this class should observe `OnPermissionChange` is in this scenario:
on the first `OnPermissionChange` no RPCs have a requiredEncryption flag so encryption service is not started.  At a later moment in time, Core sends another `OnPermissionChange`, this time with changed permissions where an RPC needs encryption so the service needs to be started.  If we rely on permission manager, then we would have to add new methods to the public header and also it will be hard to observe a change.

SmartDeviceLink/SDLProtocol.m
SatbirTanda: In testing this always sends an ACK with encryption = 0 (failed to start due to encryption bit set to 0 in ACK)

SmartDeviceLink/SDLEncryptionLifecycleManager.m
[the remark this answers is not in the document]
SatbirTanda: According to apple docs `When the value of this property is NO, the queue actively starts operations that are in the queue and ready to execute. Setting this property to YES prevents the queue from starting any queued operations, but already executing operations continue to execute. You may continue to add operations to a queue that is suspended but those operations are not scheduled for execution until you change this property to NO.`. I see some RPCs fail to be sent/added to the queue when its suspended; think we should handle this differently

SatbirTanda:
@theresalech ready for re-review.  @joeljfischer I moved the checking of the encryption from the LifecycleManager to SDLProtocol.  I think the refactor of `SDLProxy` would end up making that class almost empty, which might be as a new proposal in a later release it seems out of scope for this.

SmartDeviceLink/SDLEncryptionLifecycleManager.m
```diff
@@ -0,0 +1,248 @@
...
+
+- (BOOL)rpcRequiresEncryption:(__kindof SDLRPCMessage *)rpc {
+    if (self.permissions[rpc.name].requireEncryption != nil) {
+        return self.permissions[rpc.name].requireEncryption.boolValue;
+    }
+    return NO;
+}
+
+- (BOOL)containsAtLeastOneRPCThatRequiresEncryption {
+    for (SDLPermissionItem *item in self.permissions) {
```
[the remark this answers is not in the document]
SatbirTanda: That occurs in the loop
```
    for (SDLPermissionItem *item in self.permissions) {
        SDLPermissionItem *currentItem = self.permissions[item.rpcName];
        if(currentItem.requireEncryption.boolValue) {
            return YES;
        }
    }
    return NO;
```

SmartDeviceLink/SDLLifecycleManager.m
SatbirTanda: `securityManagers` property and `init` methods cannot be deprecated easily in `SDLStreamingMediaConfiguration` due to the dependencies of `SDLStreamingAudioLifecycleManager` and `SDLStreamingVideoLifecycleManager` and related classes.  This should a new proposal written to refactor to code, like the refactor of `SDLProxy`

SatbirTanda:
@theresalech ready for rereview

SmartDeviceLink/SDLEncryptionLifecycleManager.m
```diff
@@ -0,0 +1,247 @@
...
+@property (weak, nonatomic) id<SDLConnectionManagerType> connectionManager;
+@property (weak, nonatomic) SDLProtocol *protocol;
+
+@property (strong, nonatomic, readwrite) SDLStateMachine *encryptionStateMachine;
+@property (copy, nonatomic, nullable) SDLHMILevel currentHMILevel;
+@property (strong, nonatomic) NSMutableDictionary<SDLPermissionRPCName, SDLPermissionItem *> *permissions;
+
+@end
+
+@implementation SDLEncryptionLifecycleManager
```
NicoleYarroch: Wrap this class in `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END`.

SmartDeviceLink/SDLEncryptionLifecycleManager.m
```diff
@@ -0,0 +1,247 @@
...
+    }
+}
+
+#pragma mark - SDL RPC Notification callbacks
+- (void)sdl_permissionsDidChange:(SDLRPCNotificationNotification *)notification {
+    if (![notification isNotificationMemberOfClass:[SDLOnPermissionsChange class]]) {
+        return;
+    }
+    
+    SDLOnPermissionsChange *onPermissionChange = notification.notification;
```
NicoleYarroch: Check `onPermissionChange.requireEncryption` here to see if it `false` and return early after emptying the `permissions` dictionary. The individual permissions only need to be checked if `requireEncryption` is `nil` or `true`.

SatbirTanda: Sounds good, made th changes

SatbirTanda:
@theresalech ready for review

SmartDeviceLink/SDLProtocol.m
```diff
@@ -310,9 +335,11 @@ - (BOOL)sendRPC:(SDLRPCMessage *)message encrypted:(BOOL)encryption error:(NSErr
             // If we're trying to encrypt, try to have the security manager encrypt it. Return if it fails.
             // TODO: (Joel F.)[2016-02-09] We should assert if the service isn't setup for encryption. See [#350](https://github.com/smartdevicelink/sdl_ios/issues/350)
             messagePayload = encryption ? [self.securityManager encryptData:rpcPayload.data withError:error] : rpcPayload.data;
```
NicoleYarroch: When I attempt to send an encrypted `AddCommand` it crashes here with a `EXC_BAD_ACCESS` error. Any idea what might be happening? Not sure if it is an issue with the library or the security manager.

NicoleYarroch:
`SDLProtocolMessage` outputs out an ominous message when it tries to print a description of data that is encrypted.

```
SDLRPCPayload error: *** -[NSConcreteMutableData subdataWithRange:]: range {268, 18446744073709551521} causes integer overflow
```

I would recommend changing the `description:` method to return early if the header is encrypted:
```
- (NSString *)description {
    // Print the header data.
    NSMutableString *description = [[NSMutableString alloc] init];
    [description appendString:self.header.description];

    if (self.header.encrypted) {
        // Printing encrypted data always fails
        return description;
    }
    ...
}

File: SmartDeviceLink/SDLLifecycleManager.m [hunk 42]
```diff
@@ -614,10 +631,23 @@ - (void)sendConnectionRequest:(__kindof SDLRPCRequest *)request withResponseHand
...
-
+    
+    if (!request.isPayloadProtected && [self.encryptionLifecycleManager rpcRequiresEncryption:request]) {
+        request.payloadProtected = YES;
+    }
+    
+    if (request.isPayloadProtected && !self.encryptionLifecycleManager.isEncryptionReady) {
+        SDLLogW(@"Encryption Manager not ready, request not sent (%@)", request);
+        if (handler) {
+            handler(request, nil, [NSError sdl_encryption_lifecycle_notReadyError]);
```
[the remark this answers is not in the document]
NicoleYarroch: I'd like to do option 1 but unfortunately it would be a breaking change to do the equivalent of completion handlers in Android. I think we should try to investigate option 2 where the operation queue is suspended until encryption is setup and see if it even works.

SmartDeviceLink/SDLEncryptionLifecycleManager.m
```diff
@@ -0,0 +1,255 @@
...
+    @synchronized(self.protocol.protocolDelegateTable) {
+        if (![self.protocol.protocolDelegateTable containsObject:self]) {
+            [self.protocol.protocolDelegateTable addObject:self];
+        }
+    }
+    
+}
+
+- (void)stop {
+    _permissions = nil;
```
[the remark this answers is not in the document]
SatbirTanda: think you might be on an older branch, please pull changes (this property was set to `nullable`).

SmartDeviceLink/SDLEncryptionLifecycleManager.m
```diff
@@ -0,0 +1,255 @@
...
+    }
+}
+
+- (void)sdl_permissionsDidChange:(SDLRPCNotificationNotification *)notification {
+    if (![notification isNotificationMemberOfClass:[SDLOnPermissionsChange class]]) {
+        return;
+    }
+    
+    SDLOnPermissionsChange *onPermissionChange = notification.notification;
+    self.requiresEncryption = onPermissionChange.requireEncryption.boolValue;
```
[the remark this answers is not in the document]
SatbirTanda: in testing I do not see this behavior.  the check is 
```objc
    if (![self.currentHMILevel isEqualToEnum:SDLHMILevelNone]
        && (self.requiresEncryption || [self containsAtLeastOneRPCThatRequiresEncryption])) {
        [self.encryptionStateMachine transitionToState:SDLEncryptionLifecycleManagerStateStarting];
    } else {
        SDLLogE(@"Encryption Manager is not ready to encrypt.");
    }
```

think there is a bug in your `sdl_preloaded_pt.json` file in core setup

SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -614,10 +631,23 @@ - (void)sendConnectionRequest:(__kindof SDLRPCRequest *)request withResponseHand
...
         return;
     }
-
+    
+    if (!request.isPayloadProtected && [self.encryptionLifecycleManager rpcRequiresEncryption:request]) {
+        request.payloadProtected = YES;
+    }
+    
+    if (request.isPayloadProtected && !self.encryptionLifecycleManager.isEncryptionReady) {
+        SDLLogW(@"Encryption Manager not ready, request not sent (%@)", request);
```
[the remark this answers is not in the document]
SatbirTanda: Yes ive tested this.  please see here: https://github.com/smartdevicelink/sdl_ios/pull/1320#discussion_r314551019

SmartDeviceLink/SDLLifecycleManager.m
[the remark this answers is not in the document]
SatbirTanda: Yes ive tested this. please see here: https://github.com/smartdevicelink/sdl_ios/pull/1320#discussion_r314551019

SmartDeviceLink/SDLEncryptionLifecycleManager.m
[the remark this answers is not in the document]
NicoleYarroch: According to the evolution proposal the new parameter, `requireEncryption` in the `OnPermissionsChange` notification is **not** mandatory.

Which means we have to account for these 4 scenarios in the library: 
1. app encryption: true, functional group: true
2. app encryption: false, functional group: true
3. app encryption: omitted , functional group: true
4. app encryption: omitted, functional group: false

NicoleYarroch:
Hi @SatbirTanda, I had a meeting with the Android team and we came up with an outline of how we think the encryption manager should handle some of the issues that have come up during review:

1. If the encryption lifecycle manager fails to setup the encrypted service, the manager should retry the encrypted service one more time before failing. 
2. It was decided that suspending the operation queue while setting up the encryption service was bad idea because it could a few seconds to download a certificate and setup the encrypted service. In that time the RPCs could become stale and there would be no point in sending them. 
3. To let developers know that the encryption manager has started we will add a delegate, `SDLServiceEncryptionDelegate`, to the `SDLEncryptionConfiguration`. 
	* `serviceEncryptionUpdatedOnService:(SDLServiceType)type encrypted:(BOOL)encrypted error:(NSError *__nullable error)error`
		* `type` will return whichever type had an encryption update (for now probably only `SDLServiceTypeRPC `, but it could also apply to video / audio in the future.
		* `encrypted` will return `true` if the the encryption service was setup successfully.
		* `encrypted` will return `false` if the service is presently not encrypted.
		* `error` will return any error that happens or `nil` if there is no error.
	* Any RPCs that need to be encrypted will not be added to the operation queue if sent before the encryption service has completed the setup (i.e. they will "fail" and the error should be passed back to the developer)
4. If the developer wants to sends their own encrypted RPCs they will have to start the encryption lifecycle manager themselves using the `startRPCEncryption` method in the `SDLManger` class. 
	* When the service starts/fails, the `SDLServiceEncryptionDelegate` method will be called will the result.

SmartDeviceLink/SDLEncryptionLifecycleManager.m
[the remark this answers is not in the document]
SatbirTanda: yes the bool value for nil object is false.  I factored out the check to be more clear.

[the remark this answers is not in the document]
SatbirTanda: BOOL property can't be nullable in objective-C.  Will have to switch to NSNumber to handle this properly

SatbirTanda:
@NicoleYarroch I disagree with `If the encryption lifecycle manager fails to setup the encrypted service, the manager should retry the encrypted service one more time before failing.`  The Audio and Video manager do not attempt to retry starting a secure service, there is no need to.  It would be making the same request twice.  In the case for the proxy transport connect it seems necessary because  of the nature of iAP / TCP.

SatbirTanda:
@theresalech ready for review

SmartDeviceLink/SDLEncryptionLifecycleManager.m
```diff
@@ -0,0 +1,274 @@
...
+
+@property (strong, nonatomic, readwrite) SDLStateMachine *encryptionStateMachine;
+@property (copy, nonatomic, nullable) SDLHMILevel currentHMILevel;
+@property (strong, nonatomic, nullable) NSMutableDictionary<SDLPermissionRPCName, SDLPermissionItem *> *permissions;
+@property (assign, nonatomic, nullable) NSNumber *requiresEncryption;
+@property (weak, nonatomic, nullable) id<SDLServiceEncryptionDelegate> delegate;
+
+@end
+
+@implementation SDLEncryptionLifecycleManager
```
NicoleYarroch: It has come to my attention from the Android team that all of our screen managers (i.e. the `SDLTextAndGraphicManager`, `SDLSoftButtonManager`, `SDLMenuManager`, etc.) will fail to work if the policy table requires that any of the RPCs used to start those managers be encrypted. 

I think the most efficient solution to this problem is to pull out the RPCs that require encryption into an array on this manager until the encryption manager has started and then add them back to the sent queue. Of course it does mean that stale RPCs will be sent but there does not seem to be an easier solution.

NicoleYarroch: Please disregard the above comment. Dealing with the sub-managers will be dealt with in a new proposal for the next release since there will be issues with head units that do not return an `OnPermissionChange` notification.

NicoleYarroch:
> @NicoleYarroch I disagree with `If the encryption lifecycle manager fails to setup the encrypted service, the manager should retry the encrypted service one more time before failing.` The Audio and Video manager do not attempt to retry starting a secure service, there is no need to. It would be making the same request twice, there is not really a timing issue here. In the case for the proxy transport connect it seems necessary because of the nature of iAP / TCP.

A new proposal will be submitted to handle this case for the next release

jordynmackool:
@SatbirTanda Please see the powerpoint attached from our alignment meeting on 2019-09-10. 

[Livio Feedback 2019-09-10; RPC Message Protection.pdf](https://github.com/smartdevicelink/sdl_ios/files/3597180/Livio.Feedback.2019-09-10.RPC.Message.Protection.pdf)

SatbirTanda:
@theresalech ready for review

SmartDeviceLink/SDLError.m
```diff
@@ -25,6 +26,44 @@
...
+
++ (NSError *)sdl_encryption_lifecycle_nack {
+    NSDictionary<NSString *, NSString *> *userInfo = @{
+                                                       NSLocalizedDescriptionKey: NSLocalizedString(@"Encryption Lifecycle received a negative acknowledgement", nil),
+                                                       NSLocalizedFailureReasonErrorKey: NSLocalizedString(@"The remote head unit sent a nACK.  Encryption service failed to start due to nACK.", nil),
+                                                       NSLocalizedRecoverySuggestionErrorKey: NSLocalizedString(@"Make sure your certificates are valid.", nil)
+                                                       };
+    
+    return [NSError errorWithDomain:SDLErrorDomainEncryptionLifecycleManager
+                               code:SDLEncryptionLifecycleManagerErrorNACK
```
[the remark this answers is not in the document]
joeygrover: Both are acceptable and mean the same thing; NAK is not a typo and commonly used. A quick google search will show you this. Please try to focus on fixing issues the project maintainer finds so this feature can be included in the release and we don't have to pull it out.

SmartDeviceLink/SDLPermissionManager.h
```diff
@@ -84,6 +89,12 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)removeObserverForIdentifier:(SDLPermissionObserverIdentifier)identifier;
 
+
+/**
+ *  Check whether or not an RPC needs encryption.
+ */
+- (BOOL)rpcRequiresEncryption:(__kindof SDLRPCMessage *)rpc;
```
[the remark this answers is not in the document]
joeljfischer: Nicole's comment isn't quite right. It should be `SDLPermissionRPCName` like other APIs in the `SDLPermissionManager`. It should not be `SDLRPCMessage` because that requires an overly verbose and complicated API. `SDLPermissionRPCName` isn't a perfect solution, but it's consistent with the other permission manager APIs and can be documented well and aligns relatively well with Android's usage of `FunctionID` which contains both the `FunctionID` and the function name.

SmartDeviceLink/SDLError.m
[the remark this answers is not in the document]
SatbirTanda: Makes sense sorry about that, mainly wanted to mention there is inconsistency in the code between `NACK` and `NAK`.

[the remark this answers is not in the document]
joeygrover: Inconsistencies aren't good either. Our [protocol spec defines them as NAK](https://github.com/smartdevicelink/protocol_spec#23-frame-header-fields) so that should be the standard we use in the project. Not sure when `SDLFrameInfoStartServiceNACK` was introduced, but I would say it is the outlier, but the pay off has to be worth the effort (deprecation, breaking changes, etc).

SatbirTanda:
@theresalech ready for review

SmartDeviceLink/SDLError.m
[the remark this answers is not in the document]
SatbirTanda: Sounds good, good thing we have a knack for this

SmartDeviceLink/SDLEncryptionLifecycleManager.m
```diff
@@ -0,0 +1,276 @@
...
+
+- (void)startEncryptionService {
+    [self.encryptionStateMachine transitionToState:SDLEncryptionLifecycleManagerStateStarting];
+}
+
+- (void)sdl_startEncryptionService {
+    SDLLogV(@"Attempting to start Encryption Service");
+    if (!self.protocol || !self.currentHMILevel) {
+        SDLLogV(@"Encryption manager is not yet started");
+        [self.delegate serviceEncryptionUpdatedOnService:SDLServiceTypeRPC encrypted:NO error:[NSError sdl_encryption_lifecycle_notReadyError]];
```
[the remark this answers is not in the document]
SatbirTanda: a few examples of this can be seen with `SDLManagerDelegate` in `SDLLifecycleManager`.  ex line 297, 518, etc

SmartDeviceLink/SDLPermissionManager.m
```diff
@@ -26,6 +26,7 @@ @interface SDLPermissionManager ()
 @property (strong, nonatomic) NSMutableDictionary<SDLPermissionRPCName, SDLPermissionItem *> *permissions;
 @property (strong, nonatomic) NSMutableArray<SDLPermissionFilter *> *filters;
 @property (copy, nonatomic, nullable) SDLHMILevel currentHMILevel;
+@property (assign, nonatomic) BOOL requiresEncryption;
```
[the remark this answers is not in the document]
NicoleYarroch: The logic for setting `requiresEncryption` should be the same as in the `SDLEncryptionLifecycleManager`. Perhaps you should create a utility class that handles the logic for both classes since they are exactly the same.

SatbirTanda:
@theresalech ready for review

SatbirTanda:
@theresalech ready for review

Issue #1304, opened 2019-06-17
[SDL 0237] Add Feature to Disable "Video Streaming Backgrounded String" Feature


theresalech:
Proposal: [Add Feature to Disable "Video Streaming Backgrounded String" Feature](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0237-add-feature-to-disable-video-streaming-backgrounded-string-feature.md)

> Add an option for app developer to disable the "Video Streaming Backgrounded String (SDL-0118)" feature.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/753

Steering Committee Decision:

> The Steering Committee voted to accept this proposal with the following revision: rename `sendVideoBackgroundedString` to `showVideoBackgroundDisplay`.

> The proposal .md file was updated to reflect these revisions on 6/17/19.

theresalech:
Proposal md file has been updated per [Accepted Evolution Proposal](https://github.com/smartdevicelink/sdl_evolution/issues/801).  Please find details of change on the review issue.

theresalech:
Closed via #1357

Issue #1269, opened 2019-05-20
[SDL 0224] Navigation Subscription Buttons


theresalech:
Proposal: [Navigation Subscription Buttons](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0224-navigation-subscription-buttons.md)

> This proposal adds support for new subscription buttons (like the existing Play/Pause, Next, Previous, etc.) specifically designed for navigation scenarios.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/712

Steering Committee Decision:

>The Steering Committee voted to accept this proposal with the following revisions:

- Incorporate [this comment](https://github.com/smartdevicelink/sdl_evolution/issues/712#issuecomment-487981666), with NAV_ROTATE including rotating left or right 90 degrees. 
- Add button press to the Navigation Service's handled RPCs with these new button types.

>Additionally, there was a question in the Steering Committee meeting about requiring applications to be compatible with this proposal.  It was advised that this requirement is out of the scope of SDL, and OEMs would need to ask application developers to implement this proposal once it's available in the library for developers to implement.

> The proposal .md file was updated to reflect these revisions on 5/20/19.

Issue #1083, opened 2018-09-26
[SDL 0199]-Adding-GPS-Shift-support 


jordynmackool:
Proposal: [Adding GPS Shift support](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0199-Adding-GPS-Shift-support.md)

> To add shifted param for GPS data struct so that HMI can communicate to app when GPS data is shifted using an algorithm

Review: https://github.com/smartdevicelink/sdl_evolution/issues/598

Steering Committee Decision:

> The Steering Committee fully agreed to accept this proposal on 2018-09-25.

theresalech:
Closed via #1299

Issue #1028, opened 2018-07-18
[SDL 0187] Restructure iOS Threading and Queueing


theresalech:
Proposal: [Restructure iOS Threading and Queueing](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0187-restructure-ios-threading.md)

> This change targets the iOS library and restructures the existing threading / asynchronicity into a more manageable, consistent, and understandable system.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/543

Steering Committee Decision:

> The Steering Committee fully agreed to accept this proposal. It was also confirmed during discussion that this proposal does not make any changes regarding minimum version.

PR #1343, opened 2019-07-12
v6.4.0 Release

joeljfischer:
### Risk

### Testing Plan
Unit tests and full smoke tests

### Summary
SDL iOS 6.4.0 brings features like widgets, a new multithreading architecture, and multiple remote control modules per type, along with a Smörgåsbord of other features and a plentitude of bug fixes.

It supports [SDL RPC Spec 6.0.0](https://github.com/smartdevicelink/rpc_spec/releases/tag/5.1.0) and [SDLProtocol Spec 5.2.0](https://github.com/smartdevicelink/protocol_spec/releases/tag/5.2.0).

##### Enhancements
* The autocomplete options for users in searchable choice sets and popup keyboards can now support more than one item (#790).
* The application can now be closed (put into HMI level NONE) at runtime programmatically (#801).
* Added the ability to open the main menu or directly to any sub-menu (#806).
* Added the ability for the module to turn on or off a gesture to dismiss the lock screen – this can also be disabled on the developer side (#932).
* Added the ability to display an icon on an alert (#995).
* Automatic video streaming now supports module capability parameters for scale (e.g. 1x, 2x), and pixel density (#1007).
* Added the ability to set a title for the displayed template layout (to name the current app screen) (#1031).
* Support for canceling popup menus, keyboards, alerts, scrollable messages, and sliders (#1055).
* `SDLGPSData` now has a `shifted` parameter to support proprietary shifts in certain locales (#1083).
* Additional remote control RADIO and CLIMATE module parameters have been added (#1143).
* The RPC session can now be encrypted when the head unit has the capability and desires it. This allows all RPCs, or a subset, to be encrypted over the wire similar to the current capability with video data (#1163).
  * A new encryption configuration has been added to configure security libraries.
    * A new delegate callback has been added to know when encryption succeeds or fails.
  * The ability to start encryption and mark RPC requests as requiring encryption has been added.
  * The library will automatically encrypt RPCs that the head unit requires to be encrypted.
* Add support for OEM-only custom (non-SDL-defined) vehicle data (#1184).
* The MEDIA app service now supports a media image (#1247).
* App services can now be unpublished and their manifests can now be updated (#1260).
* Additional subscription buttons designed for navigation applications are now available (#1269).
* Support multiple windows per display via widgets (#1270).
* Support multiple modules per remote control module type (#1272).
* A tile layout is now supported for the main menu and sub-menus (#1276).
* The background string that displays while video streaming when the app is put into the background can now be disabled (#1304).
* The lock screen can now be enabled at all times (#1367).
* The module logo on the lock screen can now be disabled (#1370).
* `SDLSoftButtonObject` can now be initialized without first initializing an `SDLSoftButtonState` (#1375).
* The secondary transport feature for video streaming can now be disabled (#1380).
* The `SDLSystemCapabilityManager` now has support for the new multi-window display capabilities with automatic backward-compatibility support (#1386).
* `SDLHMICapabilities` now supports checking app services (#1389).

##### Bug Fixes
* Fixed the permission manager processing permission updates incorrectly (#965).
* Multithreading throughout the iOS library has been re-implemented for simplicity and speed (#1028).
* `SDLSyncMsgVersion` is now `SDLMsgVersion` across the project (#1352).
* Fix the choice set manager not properly handling successes that also error (like warnings) (#1363).
* Internal operations now properly immediately when they're canceled before starting (#1379).
* The video streaming capability `bitrate` is now properly used by default (#1392).
* Fixed `SDLHMICapabilities` missing parameters (#1395, #1423).
* Fixed unit tests on Xcode 11 (#1406, #1407).
* Fix video streaming should not begin if `HMICapabilities.videoStreaming` is `NO` (#1411).
* Fix several RPC response parameters should be optional / nullable (#1417).
* Fixed iOS 13 making the lock screen dismissible (#1422).
* Fixed second phone connecting to head unit over iAP disconnecting the first phone (#1431).
* Add a check before popping a buffer to prevent crashes (#1433).
* More information is logged when asserts occur due to type mismatches (#1439).
* Fixed `SDLLightControlCapabilities` and `SDLHMISettingsControlCapabilities` not returning correctly due to mismatched types (#1441).
* Fixed wording in RPC and Protocol version blocking warning logs (#1450).
* Encrypted protocol messages are logged more clearly and with additional information (#1451).
* File Manager will still function if the `ListFiles` returns with a `DISALLOWED` `resultCode` to work around some production module bugs (#1454).

##### Example Apps
* Fixed the VR-only Perform Interaction always failing (#1353).
* Add a slider and a scrollable message example to the example apps (#1383).
* Obj-C and Swift example apps are now styled correctly on iOS 13 devices in dark mode (#1428).

### CLA


mrapitis:
@theresalech 
Ford is requesting the reversion of changes for issue #1385 for the 6.4.0 release.  We are currently seeing degradation of app connectivity with some of our head unit implementations.  

It is possible that removing the data session delay all together may have caused some timing issues that need further investigation.  We would like time to test across our various head unit implementations and provide sufficient feedback -- its also important to note that testing needs to occur with USB and BT scenarios as iAP impacts both.  

We would be open to including the bugfix in an upcoming release or hotfix release after additional testing is completed.

PR #1368, opened 2019-08-01
Fix ChoiceSetManager not calling back correctly if a WARNING occurs

### LINKED ISSUE #1363: The choice set manager does not handle a success result with an error
# Bug Report
The `SDLChoiceSetManager` is not setup to handle both an error and a success result for a `PerformInteraction` response, so a subscriber is not notified when a choice set item is selected.

##### Reproduction Steps
1. Connect the SDL_iOS Swift example app to sdl_core + sdl_hmi 
2. Click on the first PICS item in the menu

##### Expected Behavior
The subscriber should be notified that a choice set item was selected.

##### Observed Behavior
The subscriber is not notified that a choice set item was selected.

* iOS Version: n/a
* SDL iOS Version: 6.4
* Testing Against: sdl_core + sdl_hmi

Tested using the library's Swift example app

joeljfischer:
Fixes #1363 

### Risk

### Testing Plan
Tests performed against sdl_hmi

### Summary
This PR adds an additional check to `SDLPresentChoiceSetOperation` to only return an error if the RPC response `success` is `NO`.

##### Bug Fixes
* Presenting a choice set no longer fails in certain cases when it's successful.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLPresentChoiceSetOperation.m
+++ b/SmartDeviceLink/SDLPresentChoiceSetOperation.m
@@ -129,7 +129,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)sdl_presentChoiceSet {
     __weak typeof(self) weakself = self;
     [self.connectionManager sendConnectionRequest:self.performInteraction withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
-        if (!response.success.boolValue && error != nil) {
+        if (!response.success.boolValue || error != nil) {
             SDLLogE(@"Presenting choice set failed with response: %@, error: %@", response, error);
             weakself.internalError = error;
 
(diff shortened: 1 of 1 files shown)


NicoleYarroch:
This does not fix the issue I was having. Part of the issues is that the sdl_hmi has a bug where the `choiceID` is not returned when a choice set item is created with a static icon. I created an issue on the sdl_hmi repo with more detail: https://github.com/smartdevicelink/sdl_hmi/issues/222.

Perhaps the library should check if error is not `nil` OR if the response's `resultCode` is not `SUCCESS`.

SmartDeviceLink/SDLPresentChoiceSetOperation.m
```diff
@@ -129,7 +129,7 @@ - (void)sdl_updateKeyboardPropertiesWithCompletionHandler:(nullable void(^)(void
 - (void)sdl_presentChoiceSet {
     __weak typeof(self) weakself = self;
     [self.connectionManager sendConnectionRequest:self.performInteraction withResponseHandler:^(__kindof SDLRPCRequest * _Nullable request, __kindof SDLRPCResponse * _Nullable response, NSError * _Nullable error) {
-        if (error != nil) {
+        if (!response.success.boolValue || error != nil) {
```
NicoleYarroch: Checking `response.success.boolValue` does not work because `success` is `true` in the HMI bug. Rather  the `response.resultCode` should be checked to make sure that it is `SUCCESS`.

joeljfischer: Well, the HMI bug is fixed, so perhaps we should just close this issue unless it crops up elsewhere?

NicoleYarroch: This is the first time I've seen this HMI bug so I agree that we should close this issue and PR.

PR #1499, opened 2019-12-10
LockScreen bounds should always be full bounds of Main screen

### LINKED ISSUE #1492: Lockwindow got wrong size make my app lockScreen UI abnormal
### Bug Report
[Summary]
When car in moving mode, my app lock screen UI shows the lockScreen image on top left corner.
But when using SDL v6.3.1, no this issue.

##### The root cause may as below. If I remove the code like below, sometimes can work well, sometimes not work. If recover those code, lockScreen UI always not work.

@implementation SDLLockScreenViewController

//- (BOOL)shouldAutorotate {
//    if (self.presentingViewController != nil) {
//        return self.presentingViewController.shouldAutorotate;
//    } else {
//        return YES;
//    }
//}
//
//- (UIInterfaceOrientationMask)supportedInterfaceOrientations {
//    if (self.presentingViewController != nil) {
//        return self.presentingViewController.supportedInterfaceOrientations;
//    } else {
//        return UIInterfaceOrientationMaskAll;
//    }
//}

* iOS Version: [What version of iOS are you using when the bug occurred]
13.1.3
* SDL iOS Version: [What version of the library has this bug been seen on]
6.4.1

Here is the issue's screenshot link:
https://github.com/84798012/SDL_Window_Rotate_Bug_Reproduce_Demo/blob/master/WechatIMG1.jpeg

justingluck93:
Fixes #1492 

### Risk

### Testing Plan
- [n/a] I have run the unit tests with this PR

#### Unit Tests
There are no unit tests for this class 

#### Core Tests
Tested against TDK

### Summary
Made the lockscreen always be the full size of the phone screen when presented

#### Bug Fixes
Fixes the issue that sets the lockscreen to the size of the `appWindow`, in some cases the appWIndow may not be the full size of the screen. This can cause issues with how the lockscreen looks when presented.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLLockScreenPresenter.m
+++ b/SmartDeviceLink/SDLLockScreenPresenter.m
@@ -134,12 +134,8 @@ NS_ASSUME_NONNULL_BEGIN
     // We let ourselves know that the lockscreen will present, because we have to pause streaming video for that 0.3 seconds or else it will be very janky.
     [[NSNotificationCenter defaultCenter] postNotificationName:SDLLockScreenManagerWillPresentLockScreenViewController object:nil];
 
-    CGRect firstFrame = appWindow.frame;
-    firstFrame.origin.x = CGRectGetWidth(firstFrame);
-    appWindow.frame = firstFrame;
-
     // We then move the lockWindow to the original appWindow location.
-    self.lockWindow.frame = appWindow.bounds;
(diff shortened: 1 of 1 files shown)


joeljfischer:
Waiting for issue reporter

joeljfischer:
Closed in favor of #1509

PR #1507, opened 2019-12-12
LockScreen Dismiss animation 

### LINKED ISSUE #1504: Lock screen dismiss animation breaks
### Bug Report
The lock screen dismiss animation breaks the first time the lock screen is shown. 

##### Reproduction Steps
1. App is open on phone
1. TDK is set to "parked"
1. Connect app to TDK (SYNC 3 v3.0)

##### Expected Behavior
The lock screen should show and be dismissed since the vehicle is not moving. The dismissal of the lock screen should be animated. 

##### Observed Behavior
The lock screen should show and be dismissed since the vehicle is not moving. The lock screen dismissal is not animated. 

I believe this is because `dismiss` is called twice in quick succession
due to two driver distraction notifications coming in one right after another. We should probably check if the lock screen is already in the process of being dismissed before doing the dismissal.

* iOS Version: 13
* SDL iOS Version: 6.4.1
* Testing Against: TDK, SYNC 3, v3.0

justingluck93:
Fixes #1504 

### Risk

### Testing Plan
- [n/a] I have run the unit tests with this PR

#### Core Tests
Tested Lockscreen animation to make sure that dismiss is called once and the screen dismiss animation is shown

Core version / branch / commit hash / module tested against: TDK, SYNC 3, V3.0

### Summary
Made a change that makes sure the dismiss lock screen is only fired once so the animation can occur

##### Bug Fixes
dismissViewControllerAnimated was being called twice causing issues with the lockscreen animation

### CLA

### DIFF
--- a/SmartDeviceLink/SDLLockScreenPresenter.m
+++ b/SmartDeviceLink/SDLLockScreenPresenter.m
@@ -22,6 +22,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+BOOL isDissmissing = NO;
 
 @implementation SDLLockScreenPresenter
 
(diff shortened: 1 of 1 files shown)


joeljfischer:
Closed in favor of #1509

PR #1524, opened 2020-01-22
Fixing lock screen dismiss call if not being presented.

### LINKED ISSUE #1523: Dismiss LockViewController not always completed
### Bug Report
This is a regression since 6.5.0. With 6.5.0 the lock screen presenter has changed. The logic to dismiss a lock screen is triggered for many different reasons e.g. manager connects or app gets to foreground on the phone. See [SDLLockScreenPresenter.m:168](https://github.com/smartdevicelink/sdl_ios/blob/6.5.0/SmartDeviceLink/SDLLockScreenPresenter.m#L168). Every dismiss attempt sends a `SDLLockScreenManagerWillDismissLockScreenViewController` notification. However if the lock screen is not presented, the notification `SDLLockScreenManagerDidDismissLockScreenViewController` is missing. The latter one is important for `SDLCarWindow` to operate as it's tracking lockscreen transitions. The consequence is that mobile navigation apps don't project the UI as the car window is early returning in `syncFrame` see [SDLCarWindow.m:70](https://github.com/smartdevicelink/sdl_ios/blob/6.5.0/SmartDeviceLink/SDLCarWindow.m#L70).

To solve the issue the lock screen presenter should not send notifications if `LockViewController` is not presented. See below snippet that I tested right before the notification solved the problem for me.

```objc
    if (self.lockViewController.presentingViewController == nil) {
        SDLLogW(@"Attempted to dismiss lockscreen, but lockViewController is not presented");
        if (completionHandler == nil) { return; }
        return completionHandler();
    }
```

##### Reproduction Steps
1. Install mobile navigation app with sdl_ios 6.5.0
2. Connect iPhone to an IVI
3. Start the nav app on the phone
4. Start the nav app on the IVI

##### Expected Behavior
The nav app should stream the app UI to the IVI

##### Observed Behavior
The IVI is waiting for the app to start the stream
- The CarWindow is ignoring `syncFrame` calls

* iOS Version: 13.3 (reproduced on 13.2 as well)
* SDL iOS Version: 6.5.0
* Testing Against: Proprietary app. Can provide access upon request

kshala-ford:
Fixes #1523 

### Risk

##### Bug Fixes
* Fixes issue where LockScreenWillDismiss notification appears without LockScreenDidDismiss.

### Tasks Remaining:
- [x] [Add early return if lock view controller is not presented]
- [x] [Run existing unit tests successfully]

### CLA

### DIFF
--- a/SmartDeviceLink/SDLLockScreenPresenter.m
+++ b/SmartDeviceLink/SDLLockScreenPresenter.m
@@ -163,6 +163,12 @@ NS_ASSUME_NONNULL_BEGIN
         if (completionHandler == nil) { return; }
         return completionHandler();
     }
+    
+    if (self.lockViewController.presentingViewController == nil) {
+        SDLLogW(@"Attempted to dismiss lockscreen, but lockViewController is not presented");
+        if (completionHandler == nil) { return; }
+        return completionHandler();
+    }
 
     // Let ourselves know that the lockscreen will dismiss so we can pause video streaming for a few milliseconds - otherwise the animation to dismiss the lockscreen will be very janky.
     [[NSNotificationCenter defaultCenter] postNotificationName:SDLLockScreenManagerWillDismissLockScreenViewController object:nil];


joeljfischer:
Hi @kshala-ford. Unfortunately, I'm having difficulty reproducing this issue. I understand the general idea of it, but I have tested an example video streaming app (the Obj-C example app with `AppHMITypeNavigation` and insecure autostreaming of a blank `UIViewController`) on Sync 3.0 and Sync 3.4 and have not been able to reproduce it. `syncFrame` is seems to be called correctly for me whether I start it without a lock screen, with a lock screen, or change the lock screen.

I'm also having difficulty understanding this statement from the attached issue:

> Every dismiss attempt sends a `SDLLockScreenManagerWillDismissLockScreenViewController` notification. However if the lock screen is not presented, the notification `SDLLockScreenManagerDidDismissLockScreenViewController` is missing.

Should the second sentence be "the notification `SDLLockScreenManagerDidDismissLockScreenViewController` should not be sent."?

Could you please clarify the original issue and provide additional reproduction steps?

joeljfischer:
I was able to reproduce this and test the fix on iOS 12 devices.

PR #1385, opened 2019-08-14
Tweak IAP transport to only delay control sessions, not data sessions

joeljfischer:
Fixes #1240

This PR is **ready** for review.

### Risk
This PR makes **no** API changes.

### Testing Plan
Smoke tests were performed with the following scenarios:
1. 6 apps were forced to connect with this branch on the control protocol repeatedly.
2. 13 apps were forced to connect with this branch on the multisession protocol repeatedly.

### Summary
This PR updates the IAP transport to only delay apps from connecting with their retry delay when the control session is being used for connection, not when the multisession protocol (direct to data session) is being used.

##### Bug Fixes
* Connecting to newer head units using the multisession protocol is sped up.

### CLA
- [x] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

PR #1549, opened 2020-02-07
Remove the delay on data sessions

### LINKED ISSUE #1240: Don't Delay Apps Attempting to Connect over the MultiSession Protocol String
### Bug Report
Currently it appears that all apps are delayed in connecting to either a control or data session. However, since multisession apps directly connect to the multisession protocol string, we should not delay those apps from connecting.

See: `SDLIAPTransport sdl_accessoryConnected:` and `SDLIAPTransport sdl_connectAccessory:`

##### Reproduction Steps
1. Connect an app over the multisession protocol string

##### Expected Behavior
No delay exists in connecting the apps

##### Observed Behavior
Apps are delayed in connecting 1.5 - 9.5 seconds

* iOS Version: n/a
* SDL iOS Version: 6.2.0-develop
* Testing Against: Ford TDK Sync 3.0

joeljfischer:
Fixes #1240

Recreates #1385, which was reverted

### Risk

### Testing Plan
Smoke tests were performed with the following scenarios:
1. 6 apps were forced to connect with this branch on the control protocol repeatedly.
2. 13 apps were forced to connect with this branch on the multisession protocol repeatedly.

### Summary
This PR updates the IAP transport to only delay apps from connecting with their retry delay when the control session is being used for connection, not when the multisession protocol (direct to data session) is being used.

##### Bug Fixes
* Connecting to newer head units using the multisession protocol is sped up.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLIAPDataSession.m
+++ b/SmartDeviceLink/SDLIAPDataSession.m
@@ -110,6 +110,9 @@ NS_ASSUME_NONNULL_BEGIN
     [self sdl_isIOThreadCanceled:self.canceledSemaphore completionHandler:^(BOOL success) {
         if (success == NO) {
             SDLLogE(@"Destroying thread (IOStreamThread) for data session when I/O streams have not yet closed.");
+            
+            // FIX: Try to close the session if the canceledSemaphore is never triggered by the `sdl_accessoryEventLoop`
+            [self sdl_closeSession];
         }
         self.ioStreamThread = nil;
         [super cleanupClosedSession];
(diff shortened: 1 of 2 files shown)


joeljfischer:
@kshala-ford @mrapitis This PR removes the data session delay, which was planned for v6.4.0 but removed due to concerns. Please do additional testing with it if you can, it would be nice to add it to the v6.6.0 or v6.7.0 release this year.

kshala-ford:
@joeljfischer thank you. I was going to get in contact with you about this PR. My observations from testing were that the first versions with multisession suffer the most compared to the most recent builds to be practically not affected at all. We will do some more thorough testing to confirm. Eventually we can reduce the delay for head units running Core 4.x and immediately connect on Core 5.x (expecting that identifying Core is possible with EAAccessory data only).

PR #1568, opened 2020-02-21
Scale Workaround for legacy high-res displays

### LINKED ISSUE #1567: Scale Workaround for legacy high-res displays
### Bug Report

Requesting a workaround for two head units with high-resolution displays that don't support scale values. The reason is per user complaints but also (and more importantly) these displays won't meet driver distraction guidelines for SDL navigation apps. With "pixel density and scale" API the workaround is rather small allowing these head units to increase the appearance of the app UI elements.

##### Reproduction Steps
1. Start an IVI (see below affected 10" IVIs)
2. Register an SDL application with NAVIGATION HMI type
3. Start the application on the IVI 

##### Expected Behavior
- App UI should be of decent size.
- Button size should be big enough to hit.
- App performance should be decent.

##### Observed Behavior
- The app UI is very small. 
- Buttons are hard to hit.
- Framerate is very low

* iOS Version: Irrelevant
* SDL iOS Version: 6.5.0
* Testing Against: Ford Explorer 10" portrait or Lincoln Navigator 10" landscape

kshala-ford:
Fixes #1567 

### Risk

### Testing Plan

#### Core Tests
The change was tested primarily on a Lincoln 10" test bench to confirm the code serves the expected behavior. It was also tested against other OEM specific vehicles that have smaller displays to confirm the code does not cause unexpected changes. Finally it was tested against Core on a test bench that provides actual scaling values to confirm the actual scaling value is used instead of the fake value.

### Summary
Adds a condition for OEM specific head units that have larger screens than 800 pixels per edge. The faked scale value is preset upon RAI response and will be overwritten if the IVI provides actual video streaming capabilities.

##### Bug Fixes
* Fixes an issue on older high-res IVIs that were produced before pixel scale proposal.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLStreamingVideoLifecycleManager.m
+++ b/SmartDeviceLink/SDLStreamingVideoLifecycleManager.m
@@ -592,7 +592,7 @@ typedef void(^SDLVideoCapabilityResponseHandler)(SDLVideoStreamingCapability *_N
 #pragma clang diagnostic pop
     if (resolution != nil) {
         self.videoScaleManager.displayViewportResolution = CGSizeMake(resolution.resolutionWidth.floatValue,
-                                                                      resolution.resolutionHeight.floatValue);
+                                 resolution.resolutionHeight.floatValue);
         // workaround for older high resolution displays that don't support scaling
         NSString *make = registerResponse.vehicleType.make;
         CGSize resolution = self.videoScaleManager.displayViewportResolution;
--- a/SmartDeviceLink/SDLStreamingVideoLifecycleManager.m
+++ b/SmartDeviceLink/SDLStreamingVideoLifecycleManager.m
@@ -592,7 +592,13 @@ typedef void(^SDLVideoCapabilityResponseHandler)(SDLVideoStreamingCapability *_N
 #pragma clang diagnostic pop
     if (resolution != nil) {
         self.videoScaleManager.displayViewportResolution = CGSizeMake(resolution.resolutionWidth.floatValue,
-                                 resolution.resolutionHeight.floatValue);
+                                                                      resolution.resolutionHeight.floatValue);
+        // workaround for older high resolution displays that don't support scaling
+        NSString *make = registerResponse.vehicleType.make;
+        CGSize resolution = self.videoScaleManager.displayViewportResolution;
+        if (([make containsString:@"Ford"] || [make containsString:@"Lincoln"]) && (resolution.width > 800 || resolution.height > 800)) {
+            self.videoScaleManager.scale = 1.0f / 0.75f; // Scale by 1.333333
+        }
     }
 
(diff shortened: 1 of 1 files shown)


kshala-ford:
The PR is ready for review. Please let me know if you have any questions. I'm more than happy to provide more insight to the requested workaround.

SmartDeviceLink/SDLStreamingVideoLifecycleManager.m
```diff
@@ -593,6 +593,12 @@ - (void)sdl_didReceiveRegisterAppInterfaceResponse:(SDLRPCResponseNotification *
     if (resolution != nil) {
         self.videoScaleManager.displayViewportResolution = CGSizeMake(resolution.resolutionWidth.floatValue,
                                  resolution.resolutionHeight.floatValue);
+        // workaround for older high resolution displays that don't support scaling
```
joeljfischer:
```suggestion
        // HAX: Workaround for Ford and Lincoln displays with > 800 resolution width or height. They don't support scaling and if we don't do this workaround, they will not correctly scale the view.
```

SmartDeviceLink/SDLStreamingVideoLifecycleManager.m
```diff
@@ -593,6 +593,12 @@ - (void)sdl_didReceiveRegisterAppInterfaceResponse:(SDLRPCResponseNotification *
...
+        NSString *make = registerResponse.vehicleType.make;
+        CGSize resolution = self.videoScaleManager.displayViewportResolution;
+        if (([make containsString:@"Ford"] || [make containsString:@"Lincoln"]) && (resolution.width > 800 || resolution.height > 800)) {
```
joeljfischer: This will occur for _all_ Ford and Lincoln head units with resolution width or height > 800px. Any other head unit released above that, this code will be triggered. It should be overwritten later if the head unit supports the video streaming capability, are you banking on that?

ashwink11: Hi @joeljfischer ,
I will be the stand-in for Kujtim.
Regarding your query, this PR is just to address issues in legacy Ford IVIs. If IVI sends Video streaming capability, the new scale values should be used.

SmartDeviceLink/SDLStreamingVideoLifecycleManager.m
ashwink11: Hi @joeljfischer, I am not able to modify this PR. So, I created another one #1585 with suggested changes in the comment. Please do let me know if it requires any other update.

joeljfischer: Thanks for the update Ashwin, I will get back to that PR as soon as I can.

PR #1574, opened 2020-02-28
removed code from swift example that checks for first time hmiLevel i…

### LINKED ISSUE #1554: The Swift example app logic for checking the first time that the `hmiLevel` is `FULL` is broken
### Bug Report
The Swift example app code for finding the first time that the `hmiLevel` is `FULL` is broken as the following code never gets called in the `ProxyManager` class: 

```
if newLevel == .full && firstHMILevelState != .full {
    // This is our first time in a `FULL` state.
    firstHMILevelState = newLevel
}
```

This logic doesn't seem to be in the Obj-C example app so it either the logic needs to get fixed in both example apps or removed from the Swift example app.

##### Reproduction Steps
1. Connect Swift example app to a head unit.

##### Expected Behavior
The code above should be called the very first time the `hmiLevel` is `FULL`.

##### Observed Behavior
The code above is never called.

* iOS Version: n/a
* SDL iOS Version: 6.5
* Testing Against: SYNC 4

lapinskijw:
Fixes #1554 

### Risk

### Testing Plan

#### Unit Tests
All changes are in example app, unit tests not needed.

#### Core Tests
n/a Core unaffected by changes.

### Summary
Removed the check in the example Swift app that checks for first `hmiLevel` `FULL`.

##### Bug Fixes
* Removed unused method if statement.

### CLA

### DIFF
--- a/Example Apps/Example Swift/ProxyManager.swift
+++ b/Example Apps/Example Swift/ProxyManager.swift
@@ -173,11 +173,6 @@ extension ProxyManager: SDLManagerDelegate {
             vehicleDataManager.subscribeToVehicleOdometer()
         }
 
-        if newLevel == .full && firstHMILevelState != .full {
-            // This is our first time in a `FULL` state.
-            firstHMILevelState = newLevel
-        }
-
         switch newLevel {
         case .full:                // The SDL app is in the foreground
             // Always try to show the initial state to guard against some possible weird states. Duplicates will be ignored by Core.


Example Apps/Example Swift/ProxyManager.swift
```diff
@@ -173,11 +173,6 @@ extension ProxyManager: SDLManagerDelegate {
             vehicleDataManager.subscribeToVehicleOdometer()
         }
 
-        if newLevel == .full && firstHMILevelState != .full {
-            // This is our first time in a `FULL` state.
-            firstHMILevelState = newLevel
```
joeljfischer: The `firstHMILevelState` property needs to be removed, then re-test

lapinskijw: Looking at this again, won't we want to keep the `firstHMILevelState` property to recognize first time in a non-`NONE` state? Otherwise we will be sending commands and subscribing to vehicle data every time `newLevel` is non-`NONE` when `hmiLevel(oldLevel:didChangeToLevel:)` is called. Keeping this would also match the behavior of the Objective-C app.
https://github.com/smartdevicelink/sdl_ios/blob/e94851172e2b99408eb4a033a557438406d3cde1/Example%20Apps/Example%20Swift/ProxyManager.swift#L166-L174

PR #1576, opened 2020-03-05
OOM when the WiFi signal is extremely weak

### LINKED ISSUE #1529: App shutting down on Out-of-memory error when video streaming is started about 10 minutes if the WiFi signal is extremely weak.
### Bug Report
App shutting down on Out-of-memory error when video streaming is started about 10 minutes if the WiFi signal is extremely weak.

##### Preconditions
The WiFi signal is extremely weak.

##### Reproduction Steps
1. Connect the head unit to the phone via BT+WiFi
2. Register a NaviAPP
3. Activate a NaviAPP
4. Start video streaming
5. After about 10 minutes

##### Expected Behavior
App not shutting down on Out-of-memory error

##### Observed Behavior
App shutting down on Out-of-memory error

①The data of video streaming is saved into `SDLTcpTransport's sendDataQueue`,
then is sent to HU.

```
#pragma mark - Data Transmission

- (void)sendData:(NSData *)msgBytes {
    [self.sendDataQueue enqueueBuffer:msgBytes.mutableCopy];

    [self performSelector:@selector(sdl_writeToStream) onThread:self.ioThread withObject:nil waitUntilDone:NO];
}

- (void)sdl_writeToStream {
    NSAssert([[NSThread currentThread] isEqual:self.ioThread], @"sdl_writeToStream is called on a wrong thread!");

    if (!self.outputStreamHasSpace || [self.sendDataQueue count] == 0) {
        return;
    }

    NSMutableData *buffer = [self.sendDataQueue frontBuffer];
    NSUInteger bufferLen = buffer.length;

    NSInteger bytesWritten = [self.outputStream write:buffer.bytes maxLength:bufferLen];
    if (bytesWritten < 0) {
        SDLLogW(@"TCP transport write error: %@", self.outputStream.streamError);
        [self sdl_onStreamError:self.outputStream];
        return;
    } else if (bytesWritten == 0) {
        SDLLogD(@"TCP transport output stream closed");
        [self sdl_onStreamEnd:self.outputStream];
        return;
    }

    if (bytesWritten == bufferLen) {
        // we have consumed all of data in this buffer
        [self.sendDataQueue popBuffer];
    } else {
        [buffer replaceBytesInRange:NSMakeRange(0, (NSUInteger)bytesWritten) withBytes:NULL length:0];
    }

    // the output stream may still have some spaces, but let's wait for another NSStreamEventHasSpaceAvailable event before writing
    self.outputStreamHasSpace = NO;
}
```

②There is no limit to the size of the `sendDataQueue`，so the app can save streaming data into the queue without limit.
packetQueue' data can't be sent to HU via TCP immediately when the WiFi signal is extremely weak, so Out-of-memory error occurs.

* iOS Version: 10.2
* SDL iOS Version: 6.4.1
* Testing Against: SDL Core 6.0.1 + HMI running on our development environment.

##### Occurrence
100%

##### Recovery Method:
Kill the NaviAPP, connect BT+WIFI, restart the NaviAPP

zhouxin627:
Fixes #1529 

### Risk

### Testing Plan
Covered by Unit tests.

### Summary
**Reason** 
The data of video streaming and audio streaming is saved. There is no limit to the size of the Queue, so the app can save streaming data into the queue without limit. Queue's data can't be sent to HU via TCP immediately when the WiFi signal is extremely weak, so crash happens as OOM due to queue increases.

**Solution** 
Set a max value to the size of the Queue.                 
When the size of the Queue reaches the max value, clear all the data in the Queue.

max value is 200 (how to decide the max value, referred to as follows）

IOS   TestData | iPhone11pro(iOS13.2) | iPhoneXR(iOS13.3.1)
-- | -- | --
①nothing to do when VMP is   showing | size of Queue between 0 and 2 | the same as left
②touch playAudioStream | size of Queue between 0 and 3 | the same as left
③touch playAudioStream and   touch other button | size of Queue between 0 and 3 | the same as left
④repeat touch or move screen | size of Queue between 0 and 2 | the same as left
⑤when OOM | 500+ | the same as left

### CLA

### DIFF
--- a/SmartDeviceLink/SDLMutableDataQueue.m
+++ b/SmartDeviceLink/SDLMutableDataQueue.m
@@ -10,7 +10,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-NSInteger const MAXDATACOUNT = 200;
+//NSInteger const MAXDATACOUNT = 200;
 
 @interface SDLMutableDataQueue ()
 
@@ -35,15 +35,15 @@ NSInteger const MAXDATACOUNT = 200;
 - (void)enqueueBuffer:(NSMutableData *)data {
     // Since this method is being called from the main thread and the dequeue methods are being called from the data session stream thread, we need to put critical sections around the queue members. Use the @synchronized object level lock to do this.
     @synchronized(self) {
-        if ([_elements count] > MAXDATACOUNT) {
-            [_elements removeAllObjects];
-            [_elements addObject:data];
-            self.frontDequeued = NO;
-        }
-        else {
+//        if ([_elements count] > MAXDATACOUNT) {
+//            [_elements removeAllObjects];
+//            [_elements addObject:data];
+//            self.frontDequeued = NO;
+//        }
+//        else {
             [self.elements addObject:data];
             self.frontDequeued = NO;
-        }
+//        }
     }
 }
 
--- a/SmartDeviceLink/SDLMutableDataQueue.m
+++ b/SmartDeviceLink/SDLMutableDataQueue.m
@@ -10,6 +10,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+NSInteger const MAXDATACOUNT = 200;
+
 @interface SDLMutableDataQueue ()
 
 @property (nonatomic, strong) NSMutableArray *elements;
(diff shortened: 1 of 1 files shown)


theresalech:
See issue #1529  for closing rationale.

PR #1603, opened 2020-03-25
Made the choice set manager thread safe

### LINKED ISSUE #1584: Crash in SDLChoiceSetManager due to multiple threads accessing nonatomic NSSets
### Bug Report
Experienced a crash in `SDLChoiceSetManager` due to multiple threads accessing nonatomic `NSMutableSet`s `preloadedMutableChoices` and `pendingMutablePreloadChoices`.

##### Reproduction Steps
This bug is difficult to reproduce due to this being a race condition. I've only seen this bug once.
1. Create multiple choice sets using the Screen Manager API.
1. Connect app to head unit. 

##### Expected Behavior
The app does not crash

##### Observed Behavior
The app does crash due to multiple threads trying to access `preloadedMutableChoices` and `pendingMutablePreloadChoices`in the `SDLChoiceSetManager`.

* iOS Version: n/a
* SDL iOS Version: 6.5
* Testing Against: n/a

Testing with Livio music player.

NicoleYarroch:
Fixes #1584

### Risk

### Testing Plan

#### Unit Tests
Added test cases to the _SDLChoiceSetManagerSpec_ to verify that the list of pending and uploaded choice items is not updated after the choice set manager is shutdown. This can happen if the transport is destroyed while choice items are being preloaded, deleted or presented. 

#### Core Tests
1. Checked that destroying the transport while choices are being preloaded does not break the choice set manager when the transport is reestablished.
1. Checked that destroying the transport while choices are being presented does not break the choice set manager when the transport is reestablished.
1. Check that destroying the transport while the keyboard is being presented does not break the choice set manager when the transport is reestablished.  

Core version / branch / commit hash / module tested against: SYNC 3.0
HMI name / version / branch / commit hash / module tested against: SYNC 3.0

### Summary
Made the choice set manager thread safe.

##### Bug Fixes
* Fixed the library crashing due to multiple threads accessing `NSMutableSet`s at the same time.
* Made setting the choice set ids thread safe, so no two choices can be assigned the same `choiceId` or `cancelId`.
* Fixed the choice set manager failing to present a choice set due to an `INVALID_ID`. This was due to adding choice set items to the list of pending and presenting choice items after the transport was destroyed. Since the library thought those choice items were on the module, they were not uploaded, thus creating an error when trying to present the choice set items. 

### Tasks Remaining:
- [x] Unit tests
- [x] Smoke tests

### CLA

### DIFF
--- a/SmartDeviceLink/SDLChoiceSetManager.m
+++ b/SmartDeviceLink/SDLChoiceSetManager.m
@@ -129,7 +129,9 @@ UInt16 const ChoiceCellCancelIdMin = 1;
 }
 
 - (void)stop {
-    [self.stateMachine transitionToState:SDLChoiceManagerStateShutdown];
+    [self sdl_runSyncOnQueue:^{
+        [self.stateMachine transitionToState:SDLChoiceManagerStateShutdown];
+    }];
 }
 
 + (NSDictionary<SDLState *, SDLAllowableStateTransitions *> *)sdl_stateTransitionDictionary {
@@ -168,18 +170,17 @@ UInt16 const ChoiceCellCancelIdMin = 1;
 
 - (void)didEnterStateShutdown {
     SDLLogV(@"Manager shutting down");
+    _currentHMILevel = SDLHMILevelNone;
+
     [self.transactionQueue cancelAllOperations];
     self.transactionQueue = [self sdl_newTransactionQueue];
-    _vrOptional = YES;
-    _currentHMILevel = SDLHMILevelNone;
+    _preloadedMutableChoices = [NSMutableSet set];
+    _pendingMutablePreloadChoices = [NSMutableSet set];
     _pendingPresentationSet = nil;
 
-    [self sdl_runSyncOnQueue:^{
-        self->_preloadedMutableChoices = [NSMutableSet set];
-        self->_pendingMutablePreloadChoices = [NSMutableSet set];
-        self->_nextChoiceId = ChoiceCellIdMin;
-        self->_nextCancelId = ChoiceCellCancelIdMin;
-    }];
+    _vrOptional = YES;
+    _nextChoiceId = ChoiceCellIdMin;
+    _nextCancelId = ChoiceCellCancelIdMin;
 }
 
 - (void)didEnterStateCheckingVoiceOptional {
--- a/SmartDeviceLink/SDLChoiceSetManager.m
+++ b/SmartDeviceLink/SDLChoiceSetManager.m
@@ -64,6 +64,7 @@ typedef NSNumber * SDLChoiceId;
 
 @property (strong, nonatomic, readonly) SDLStateMachine *stateMachine;
 @property (strong, nonatomic) NSOperationQueue *transactionQueue;
+@property (copy, nonatomic) dispatch_queue_t readWriteQueue;
 
 @property (copy, nonatomic, nullable) SDLHMILevel currentHMILevel;
 @property (copy, nonatomic, nullable) SDLWindowCapability *currentWindowCapability;
@@ -96,6 +97,12 @@ UInt16 const ChoiceCellCancelIdMin = 1;
     _systemCapabilityManager = systemCapabilityManager;
     _stateMachine = [[SDLStateMachine alloc] initWithTarget:self initialState:SDLChoiceManagerStateShutdown states:[self.class sdl_stateTransitionDictionary]];
     _transactionQueue = [self sdl_newTransactionQueue];
+
+    if (@available(iOS 10.0, *)) {
+        _readWriteQueue = dispatch_queue_create_with_target("com.sdl.screenManager.choiceSetManager.readWriteQueue", DISPATCH_QUEUE_SERIAL, [SDLGlobals sharedGlobals].sdlProcessingQueue);
+    } else {
+        _readWriteQueue = [SDLGlobals sharedGlobals].sdlProcessingQueue;
+    }
 
     _preloadedMutableChoices = [NSMutableSet set];
     _pendingMutablePreloadChoices = [NSMutableSet set];
@@ -160,17 +167,20 @@ UInt16 const ChoiceCellCancelIdMin = 1;
 #pragma mark - State Management
 
 - (void)didEnterStateShutdown {
-    _currentHMILevel = SDLHMILevelNone;
-
     [self.transactionQueue cancelAllOperations];
     self.transactionQueue = [self sdl_newTransactionQueue];
-    _preloadedMutableChoices = [NSMutableSet set];
-    _pendingMutablePreloadChoices = [NSMutableSet set];
-    _pendingPresentationSet = nil;
 
-    _vrOptional = YES;
-    _nextChoiceId = ChoiceCellIdMin;
-    _nextCancelId = ChoiceCellCancelIdMin;
+    __weak typeof(self) weakself = self;
+    [self sdl_runSyncOnQueue:^{
+        __strong typeof(weakself) strongself = weakself;
+        strongself->_currentHMILevel = SDLHMILevelNone;
+        strongself->_preloadedMutableChoices = [NSMutableSet set];
+        strongself->_pendingMutablePreloadChoices = [NSMutableSet set];
+        strongself->_nextChoiceId = ChoiceCellIdMin;
+        strongself->_nextCancelId = ChoiceCellCancelIdMin;
+        strongself->_pendingPresentationSet = nil;
+        strongself->_vrOptional = YES;
+    }];
 }
 
 - (void)didEnterStateCheckingVoiceOptional {
@@ -334,10 +344,10 @@ UInt16 const ChoiceCellCancelIdMin = 1;
     SDLPresentChoiceSetOperation *presentOp = nil;
     if (delegate == nil) {
         // Non-searchable choice set
-        presentOp = [[SDLPresentChoiceSetOperation alloc] initWithConnectionManager:self.connectionManager choiceSet:self.pendingPresentationSet mode:mode keyboardProperties:nil keyboardDelegate:nil cancelID:self.nextCancelId++];
+        presentOp = [[SDLPresentChoiceSetOperation alloc] initWithConnectionManager:self.connectionManager choiceSet:self.pendingPresentationSet mode:mode keyboardProperties:nil keyboardDelegate:nil cancelID:self.nextCancelId];
     } else {
         // Searchable choice set
-        presentOp = [[SDLPresentChoiceSetOperation alloc] initWithConnectionManager:self.connectionManager choiceSet:self.pendingPresentationSet mode:mode keyboardProperties:self.keyboardConfiguration keyboardDelegate:delegate cancelID:self.nextCancelId++];
+        presentOp = [[SDLPresentChoiceSetOperation alloc] initWithConnectionManager:self.connectionManager choiceSet:self.pendingPresentationSet mode:mode keyboardProperties:self.keyboardConfiguration keyboardDelegate:delegate cancelID:self.nextCancelId];
     }
     self.pendingPresentOperation = presentOp;
 
@@ -374,7 +384,7 @@ UInt16 const ChoiceCellCancelIdMin = 1;
 
     SDLLogD(@"Presenting keyboard with initial text: %@", initialText);
     // Present a keyboard with the choice set that we used to test VR's optional state
-    UInt16 keyboardCancelId = self.nextCancelId++;
+    UInt16 keyboardCancelId = self.nextCancelId;
     self.pendingPresentOperation = [[SDLPresentKeyboardOperation alloc] initWithConnectionManager:self.connectionManager keyboardProperties:self.keyboardConfiguration initialText:initialText keyboardDelegate:delegate cancelID:keyboardCancelId];
     [self.transactionQueue addOperation:self.pendingPresentOperation];
     return @(keyboardCancelId);
@@ -395,14 +405,17 @@ UInt16 const ChoiceCellCancelIdMin = 1;
 
 #pragma mark - Choice Management Helpers
 
+/// Check if any of the choices already exist on the module and remove them from being preloaded.
+/// @param choices The choices to be uploaded
 - (NSSet<SDLChoiceCell *> *)sdl_choicesToBeUploadedWithArray:(NSArray<SDLChoiceCell *> *)choices {
-    // Check if any of the choices already exist on the head unit and remove them from being preloaded
     NSMutableSet<SDLChoiceCell *> *choicesSet = [NSMutableSet setWithArray:choices];
     [choicesSet minusSet:self.preloadedChoices];
 
     return [choicesSet copy];
 }
 
+/// Checks if any of the choices have not yet been uploaded to the module and removes them from being deleted.
+/// @param choices The choices to be deleted
 - (NSSet<SDLChoiceCell *> *)sdl_choicesToBeDeletedWithArray:(NSArray<SDLChoiceCell *> *)choices {
     NSMutableSet<SDLChoiceCell *> *choicesSet = [NSMutableSet setWithArray:choices];
     [choicesSet intersectSet:self.preloadedChoices];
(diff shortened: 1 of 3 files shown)


SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -288,16 +313,27 @@ - (void)deleteChoices:(NSArray<SDLChoiceCell *> *)choices {
...
         if (weakOp.error != nil) {
             SDLLogE(@"Failed to delete choices: %@", weakOp.error);
             return;
         }
 
-        SDLLogD(@"Finished deleting choices");
-        [weakself.preloadedMutableChoices minusSet:cellsToBeDeleted];
+        // Check if the manager has shutdown because the list of uploaded choices should not be updated
+        if ([strongSelf.currentState isEqualToString:SDLChoiceManagerStateShutdown]) {
+            SDLLogD(@"The manager has shutdown");
```
joeljfischer:
```suggestion
            SDLLogD(@"Cancelling deleting choices because manager is shut down");
```

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -239,13 +251,24 @@ - (void)preloadChoices:(NSArray<SDLChoiceCell *> *)choices withCompletionHandler
...
-        [weakSelf.preloadedMutableChoices unionSet:choicesToUpload];
-        [weakSelf.pendingMutablePreloadChoices minusSet:choicesToUpload];
 
         if (handler != nil) {
             handler(weakPreloadOp.error);
         }
+
+        // Check if the manager has shutdown because the list of uploaded and pending choices should not be updated
+        if ([strongSelf.currentState isEqualToString:SDLChoiceManagerStateShutdown]) {
+            SDLLogD(@"The manager has shutdown");
```
joeljfischer:
```suggestion
            SDLLogD(@"Cancelling preloading choices because the manager is shut down");
```

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -315,13 +351,14 @@ - (void)presentChoiceSet:(SDLChoiceSet *)choiceSet mode:(SDLInteractionMode)mode
         return;
     }
 
-    if (self.pendingPresentationSet != nil) {
+    if (self.pendingPresentationSet != nil && self.pendingPresentOperation.isFinished == NO) {
```
joeljfischer:
```suggestion
    if (self.pendingPresentationSet != nil && !self.pendingPresentOperation.isFinished) {
```

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -353,8 +391,15 @@ - (void)presentChoiceSet:(SDLChoiceSet *)choiceSet mode:(SDLInteractionMode)mode
...
-        weakself.pendingPresentOperation = nil;
+        strongSelf.pendingPresentOperation = nil;
+
+        // Check if the manager has shutdown because the list of pending choices should not be updated
+        if ([strongSelf.currentState isEqualToString:SDLChoiceManagerStateShutdown]) {
+            SDLLogD(@"The manager has shutdown");
+            return;
+        }
+
+        strongSelf.pendingPresentationSet = nil;
```
joeljfischer: But the `pendingPresentationSet` can be sent to nil anyway, right? So there's no reason to add the shutdown check? It doesn't seem to anything except skipping the `pendingPresentationSet`.

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -395,45 +440,54 @@ - (void)dismissKeyboardWithCancelID:(NSNumber<SDLInt> *)cancelID {
 
 #pragma mark - Choice Management Helpers
 
+/// Checks the passed list of choices to be uploaded and returns the items that have not yet been uploaded to the module.
+/// @param choices The choices to be uploaded
```
joeljfischer: Missing `@return` information

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -395,45 +440,54 @@ - (void)dismissKeyboardWithCancelID:(NSNumber<SDLInt> *)cancelID {
...
 - (NSSet<SDLChoiceCell *> *)sdl_choicesToBeUploadedWithArray:(NSArray<SDLChoiceCell *> *)choices {
-    // Check if any of the choices already exist on the head unit and remove them from being preloaded
     NSMutableSet<SDLChoiceCell *> *choicesSet = [NSMutableSet setWithArray:choices];
     [choicesSet minusSet:self.preloadedChoices];
 
     return [choicesSet copy];
 }
 
+/// Checks the passed list of choices to be deleted and returns the items that have been uploaded to the module.
+/// @param choices The choices to be deleted
```
joeljfischer: Missing `@return`

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -395,45 +440,54 @@ - (void)dismissKeyboardWithCancelID:(NSNumber<SDLInt> *)cancelID {
...
 - (NSSet<SDLChoiceCell *> *)sdl_choicesToBeDeletedWithArray:(NSArray<SDLChoiceCell *> *)choices {
     NSMutableSet<SDLChoiceCell *> *choicesSet = [NSMutableSet setWithArray:choices];
     [choicesSet intersectSet:self.preloadedChoices];
 
     return [choicesSet copy];
 }
 
+/// Checks the passed list of choices to be deleted and returns the items that are waiting to be uploaded to the module.
+/// @param choices The choices to be deleted
```
joeljfischer: Missing `@return`

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -395,45 +440,54 @@ - (void)dismissKeyboardWithCancelID:(NSNumber<SDLInt> *)cancelID {
...
 - (NSSet<SDLChoiceCell *> *)sdl_choicesToBeRemovedFromPendingWithArray:(NSArray<SDLChoiceCell *> *)choices {
     NSMutableSet<SDLChoiceCell *> *choicesSet = [NSMutableSet setWithArray:choices];
     [choicesSet intersectSet:self.pendingPreloadChoices];
 
     return [choicesSet copy];
 }
 
+/// Assigns a unique id to each choice item.
+/// @param choices An array of choices
```
joeljfischer: Missing `@return`

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -395,45 +440,54 @@ - (void)dismissKeyboardWithCancelID:(NSNumber<SDLInt> *)cancelID {
...
 - (void)sdl_updateIdsOnChoices:(NSSet<SDLChoiceCell *> *)choices {
     for (SDLChoiceCell *cell in choices) {
         cell.choiceId = self.nextChoiceId;
-        self.nextChoiceId++;
     }
 }
 
+/// Checks each choice item to find out if it has already been uploaded or if it is the the process of being uploaded. If so, the choice item is assigned the unique id of the uploaded item.
+/// @param choiceSet A set of choice items
```
joeljfischer: Missing `@return`

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -395,45 +440,54 @@ - (void)dismissKeyboardWithCancelID:(NSNumber<SDLInt> *)cancelID {
...
 - (void)sdl_findIdsOnChoiceSet:(SDLChoiceSet *)choiceSet {
-    return [self sdl_findIdsOnChoices:[NSSet setWithArray:choiceSet.choices]];
+    [self sdl_findIdsOnChoices:[NSSet setWithArray:choiceSet.choices]];
 }
 
+/// Checks each choice item to find out if it has already been uploaded or if it is the the process of being uploaded. If so, the choice item is assigned the unique id of the uploaded item.
+/// @param choices An array of choice items
```
joeljfischer: Missing `@return`

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -459,18 +513,79 @@ - (SDLKeyboardProperties *)sdl_defaultKeyboardConfiguration {
...
+}
+
+- (NSMutableSet<SDLChoiceCell *> *)pendingMutablePreloadChoices {
+    __block NSMutableSet<SDLChoiceCell *> *set = nil;
+    [self sdl_runSyncOnQueue:^{
+        set = self->_pendingMutablePreloadChoices;
+    }];
+
+    return set;
+}
```
joeljfischer: I'm not sure that this is necessary. Every place but one that access it is already synchronized. That might be a better approach.

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -459,18 +513,79 @@ - (SDLKeyboardProperties *)sdl_defaultKeyboardConfiguration {
...
 
 - (NSSet<SDLChoiceCell *> *)pendingPreloadChoices {
-    return [_pendingMutablePreloadChoices copy];
+    __block NSSet<SDLChoiceCell *> *set = nil;
+    [self sdl_runSyncOnQueue:^{
+        set = [self->_pendingMutablePreloadChoices copy];
+    }];
+
+    return set;
+}
```
joeljfischer: Same as above

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -459,18 +513,79 @@ - (SDLKeyboardProperties *)sdl_defaultKeyboardConfiguration {
...
+/// Checks if we are already on a serial queue. If so, the block is added to the queue; if not, the block is dispatched to the serial `readWrite` queue.
+/// @discussion Used to synchronize access to class properties.
+/// @param block The block to be executed.
+- (void)sdl_runSyncOnQueue:(void (^)(void))block {
+    if (dispatch_get_specific(SDLProcessingQueueName) != nil) {
+        block();
+    } else {
+        dispatch_sync(self.readWriteQueue, block);
+    }
+}
```
joeljfischer: We use this a number of places now, should be put it on `SDLGlobals` instead?

SmartDeviceLink/SDLChoiceSetManager.m
NicoleYarroch: This method returns `void`

SmartDeviceLink/SDLChoiceSetManager.m
NicoleYarroch: This method returns `void`

SmartDeviceLink/SDLChoiceSetManager.m
NicoleYarroch: This method returns `void`.

SmartDeviceLink/SDLChoiceSetManager.m
NicoleYarroch: Fixed

SmartDeviceLink/SDLChoiceSetManager.m
NicoleYarroch: Fixed

SmartDeviceLink/SDLChoiceSetManager.m
NicoleYarroch: Fixed

SmartDeviceLink/SDLChoiceSetManager.m
NicoleYarroch: Removed and added synchronization to read access to `pendingMutablePreloadChoices` and `preloadedMutableChoices`.

SmartDeviceLink/SDLChoiceSetManager.m
NicoleYarroch: Since the `pendingMutablePreloadChoices` synchronized getter was removed, this is still needed. My understanding is that a crash will happen if a thread attempts to read while another thread is writing, then a crash can occur.

SmartDeviceLink/SDLChoiceSetManager.m
NicoleYarroch: There are four other classes that use the same logic. It would be nice to move the logic to `SDLGlobals`. The only downside is that each class gives its "readWriteQueue" a custom label that makes it easier to debug issues. Perhaps it would be better to create a separate issue for this feature.

SmartDeviceLink/SDLChoiceSetManager.m
NicoleYarroch: Fixed. Removed the state check.

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -160,33 +167,35 @@ - (void)sdl_updateTransactionQueueSuspended {
 #pragma mark - State Management
 
 - (void)didEnterStateShutdown {
```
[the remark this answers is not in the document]
NicoleYarroch: Do you want me to fix this issue in other screen sub-managers or should I open another issue?

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -160,33 +167,35 @@ - (void)sdl_updateTransactionQueueSuspended {
...
-    _preloadedMutableChoices = [NSMutableSet set];
-    _pendingMutablePreloadChoices = [NSMutableSet set];
+    _vrOptional = YES;
+    _currentHMILevel = SDLHMILevelNone;
     _pendingPresentationSet = nil;
 
-    _vrOptional = YES;
-    _nextChoiceId = ChoiceCellIdMin;
-    _nextCancelId = ChoiceCellCancelIdMin;
+    [self sdl_runSyncOnQueue:^{
```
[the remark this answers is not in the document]
NicoleYarroch: Fixed.

SmartDeviceLink/SDLChoiceSetManager.m
joeljfischer: I agree, we can make this a separate issue.

NicoleYarroch: Created PR #1614

SmartDeviceLink/SDLChoiceSetManager.m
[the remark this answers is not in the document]
NicoleYarroch: I looked at the other sub managers and it don't think it is an issue in those managers. The problem in the choice set manager is that it can take a few seconds to send 100 choice items serially so if the transport is destroyed while the items are being uploaded, it is possible for the choice set upload to complete after the manager has been stopped.

There is a small, but unlikely chance that one of the class properties could be reset in the Text and Graphics manager after shutdown. Since there is no state machine in that class, I simply reset the properties in the `start` method.

SmartDeviceLink/SDLChoiceSetManager.m
```diff
@@ -288,16 +324,27 @@ - (void)deleteChoices:(NSArray<SDLChoiceCell *> *)choices {
...
+        __strong typeof(weakSelf) strongSelf = weakSelf;
         if (weakOp.error != nil) {
             SDLLogE(@"Failed to delete choices: %@", weakOp.error);
             return;
         }
 
-        SDLLogD(@"Finished deleting choices");
-        [weakself.preloadedMutableChoices minusSet:cellsToBeDeleted];
+        // Check if the manager has shutdown because the list of uploaded choices should not be updated
+        if ([strongSelf.currentState isEqualToString:SDLChoiceManagerStateShutdown]) {
```
NicoleYarroch, line 339: Instead of checking the state here and in other blocks, we could just "reset" all the properties in the `start` method.

joeljfischer: I'm not sure what you mean by that. Why would resetting properties in `start` help? They're already reset in `didEnterStateShutdown`.

NicoleYarroch: The problem is that the `didEnterStateShutdown` is called and all the properties reset before the completion block is called for the choice set upload. Then in the completion block, some of the properties are set.

joeljfischer: Ah I gotcha. I think this is a better way to do it. Better not to set the properties at all than to have to reset them later. If a dev checks them while it's shut down it'll show the wrong data.

PR #1617, opened 2020-04-03
Delay shutting down secondary transport when device app is backgrounded

### LINKED ISSUE #1560: Delay shutting down the TCP secondary transport when device app is backgrounded
### Bug Report
Due to the way iOS manages sockets, it is impossible to keep a socket alive when the app on the device is backgrounded or the device is locked with the app in the foreground. However, it is possible to create a background task when the app enters the background that will keep the socket alive for ~1 minute. This has two UX advantages:

1. Improve user experience by avoiding a reconnect delay if the user backgrounds the app and then foregrounds the app within the time the background task is kept alive (~1 minute) (https://forums.developer.apple.com/thread/97824).
1. Improve user experience by showing the `videoStreamBackgroundString` consistently when the app on the device is backgrounded. Currently, if the device locks while video is streaming, then the `videoStreamBackgroundString` is never shown due to the socket erroring right after the device locks. We can tell the developer to configure the app to not sleep while video is streaming but if the user locks the device then there is no indication why the SDL app closed. 

We should try to close the socket when the background task expires, however the socket could still be reclaimed by the OS before then. This means we will need to make sure to handle all errors that might occur when the socket errors out. 

##### Reproduction Steps
1. Connect a video streaming app with secondary transports enabled to a head unit that supports secondary transports. Launch the SDL app if necessary and wait for video to start streaming. 
1. While the app on the device is foregrounded, lock the device screen.

##### Expected Behavior
The user should see the `videoStreamBackgroundString` before the socket is closed. 

##### Observed Behavior
The video stream simply stops since the socket errors out. 

* iOS Version: 12 & 13
* SDL iOS Version: 6.5
* Testing Against: SYNC 4

NicoleYarroch:
Fixes #1560 

### Risk

### Testing Plan

#### Unit Tests
Unit tests were added to the **SecondaryTransportManagerSpec**. 

#### Core Tests
Tests were performed with a video streaming app using a TCP secondary transport. Tests were considered successful:
1. If the background task did not expire before the app was foregrounded then video resumed and audio continued without interruption.
1. If the the background task expired while the app was backgrounded, then when the app was brought back to the foreground video and audio resumed.

##### Tests Performed
1. Put app on device in the background and brought app on device back to foreground before background task expired. 
1. Put app on device in the background and brought app on device back to foreground after the background task expired.
1. Put app on device in the background and then locked the device. Unlocked the device before the background task expired and foregrounded the app. 
1. Put app on device in the background and then locked the device. Waited for the background task to expire. Unlocked the device and foregrounded the app. 
1. Put app on device in the background and then locked the device. Waited for the background task to expire. Attempted to launch the SDL app while the device was still locked, the attempting to connect alert shows up. Unlocked the device and foregrounded the app. 
1. Locked the device while the app on the device was still in foreground. Unlocked the device.
1. Put app on device in background, opened another SDL video streaming app, then switched back to the app and brought it back to the foreground.

Core version / branch / commit hash / module tested against: SYNC 4 (20016_DEVTEST)
HMI name / version / branch / commit hash / module tested against: SYNC 4 (20016_DEVTEST)

### Summary
When the app on the device is backgrounded the secondary transport is now closed when the background task expires instead of immediately. This improves user experience by:

1. Avoiding a reconnect delay if the user backgrounds the app and then foregrounds the app within the time the background task is kept alive (~1 minute).
1. Showing the `videoStreamBackgroundString` when the app on the device is backgrounded. Previously the warning message was either flashed for a moment or not shown all.

In order to ensure that the video, audio and tcp transports are all shutdown before the background task is ended, I moved calling the `videoServiceEndedCompletionHandler` and the `audioServiceEndedCompletionHandler` from the stop service ACK and NAK handler methods to the stopped state of the `SDLStreamingVideoLifecycleManager` and `SDLStreamingAudioLifecycleManager` to ensure that both managers' properties get reset before the background task ends when the app is backgrounded. This also reduced duplicate code since both the ACK/NAK handlers transition to the stopped state anyways. 

##### Enhancements
* When the app on the device is backgrounded the secondary transport is now closed when the background task expires instead of immediately.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLBackgroundTaskManager.h
+++ b/SmartDeviceLink/SDLBackgroundTaskManager.h
@@ -16,7 +16,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @interface SDLBackgroundTaskManager : NSObject
 
-/// Handler called when the background task has ended.
+/// Handler called when the background task is about to expire. Use this handler to perform some cleanup before the background task is destroyed. When you have finished cleanup, you must call the `endBackgroundTask` function so the background task can be destroyed. If you do not call `endBackgroundTask`, the system may kill the app.
 @property (copy, nonatomic, nullable) void (^taskEndedHandler)(void);
 
 - (instancetype)init NS_UNAVAILABLE;
@@ -29,17 +29,10 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (instancetype)initWithBackgroundTaskName:(NSString *)backgroundTaskName;
 
-/**
- *  Starts a background task that allows the app to establish a session while app is backgrounded. If the app is not currently backgrounded, the background task will remain dormant until the app moves to the background.
- */
+/// Starts a background task. If the app is not currently backgrounded, the background task will remain dormant until the app moves to the background.
 - (void)startBackgroundTask;
 
-/**
- *  Cleans up a background task when it is stopped. This should be called when:
- *
- *  1. The app has established a session.
- *  2. The system has called the `expirationHandler` for the background task. The system may kill the app if the background task is not ended when `expirationHandler` is called.
- */
+/// Destroys the background task.
 - (void)endBackgroundTask;
 
 @end
--- a/SmartDeviceLink/SDLBackgroundTaskManager.h
+++ b/SmartDeviceLink/SDLBackgroundTaskManager.h
@@ -36,7 +36,7 @@ NS_ASSUME_NONNULL_BEGIN
 /// Destroys the background task.
 - (void)endBackgroundTask;
 
-/// Destroys the background task after cleanup has finished. This should only be called if you have subscribed to the `taskExpiringHandler` in order to do some cleanup before the background task is destroyed.
+/// Destroys the background task after cleanup has finished.
 - (void)expiredTaskCleanupFinished;
 
 @end
--- a/SmartDeviceLink/SDLBackgroundTaskManager.h
+++ b/SmartDeviceLink/SDLBackgroundTaskManager.h
@@ -16,6 +16,9 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @interface SDLBackgroundTaskManager : NSObject
 
+/// Handler called when the background task has ended.
+@property (copy, nonatomic, nullable) void (^taskEndedHandler)(void);
+
 - (instancetype)init NS_UNAVAILABLE;
 
 /**
--- a/SmartDeviceLink/SDLBackgroundTaskManager.h
+++ b/SmartDeviceLink/SDLBackgroundTaskManager.h
@@ -17,7 +17,7 @@ NS_ASSUME_NONNULL_BEGIN
 @interface SDLBackgroundTaskManager : NSObject
 
 /// Handler called when the background task is about to expire. Use this handler to perform some cleanup before the background task is destroyed. When you have finished cleanup, you must call the `endBackgroundTask` function so the background task can be destroyed. If you do not call `endBackgroundTask`, the system may kill the app.
-@property (copy, nonatomic, nullable) void (^taskEndedHandler)(void);
+@property (copy, nonatomic, nullable) BOOL (^taskExpiringHandler)(void);
 
 - (instancetype)init NS_UNAVAILABLE;
 
@@ -34,6 +34,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// Destroys the background task.
 - (void)endBackgroundTask;
+
+/// Destroys the background task after cleanup has finished. This should only be called if you have subscribed to the `taskExpiringHandler` in order to do some cleanup before the background task is destroyed.
+- (void)expiredTaskCleanupFinished;
 
 @end
 
--- a/SmartDeviceLink/SDLBackgroundTaskManager.h
+++ b/SmartDeviceLink/SDLBackgroundTaskManager.h
@@ -17,7 +17,7 @@ NS_ASSUME_NONNULL_BEGIN
 @interface SDLBackgroundTaskManager : NSObject
 
 /// Handler called when the background task is about to expire. Use this handler to perform some cleanup before the background task is destroyed. When you have finished cleanup, you must call the `endBackgroundTask` function so the background task can be destroyed. If you do not call `endBackgroundTask`, the system may kill the app.
-/// @return Whether or to wait for the subscriber to cleanup. If NO, the background task will be killed immediately. If YES, the background task will not be destroyed until the `expiredTaskCleanupFinished` method is called.
+/// @return Whether or to wait for the subscriber to cleanup. If NO, the background task will be killed immediately. If YES, the background task will not be destroyed until the `endBackgroundTask` method is called.
 @property (copy, nonatomic, nullable) BOOL (^taskExpiringHandler)(void);
 
 - (instancetype)init NS_UNAVAILABLE;
@@ -35,9 +35,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// Destroys the background task.
 - (void)endBackgroundTask;
-
-/// Destroys the background task after cleanup has finished.
-- (void)expiredTaskCleanupFinished;
 
 @end
 
--- a/SmartDeviceLink/SDLBackgroundTaskManager.h
+++ b/SmartDeviceLink/SDLBackgroundTaskManager.h
@@ -17,6 +17,7 @@ NS_ASSUME_NONNULL_BEGIN
 @interface SDLBackgroundTaskManager : NSObject
 
 /// Handler called when the background task is about to expire. Use this handler to perform some cleanup before the background task is destroyed. When you have finished cleanup, you must call the `endBackgroundTask` function so the background task can be destroyed. If you do not call `endBackgroundTask`, the system may kill the app.
+/// @return Whether or to wait for the subscriber to cleanup. If NO, the background task will be killed immediately. If YES, the background task will not be destroyed until the `expiredTaskCleanupFinished` method is called.
 @property (copy, nonatomic, nullable) BOOL (^taskExpiringHandler)(void);
 
 - (instancetype)init NS_UNAVAILABLE;
--- a/SmartDeviceLink/SDLBackgroundTaskManager.h
+++ b/SmartDeviceLink/SDLBackgroundTaskManager.h
@@ -17,7 +17,7 @@ NS_ASSUME_NONNULL_BEGIN
 @interface SDLBackgroundTaskManager : NSObject
 
 /// Handler called when the background task is about to expire. Use this handler to perform some cleanup before the background task is destroyed. When you have finished cleanup, you must call the `endBackgroundTask` function so the background task can be destroyed. If you do not call `endBackgroundTask`, the system may kill the app.
-/// @return Whether or to wait for the subscriber to cleanup. If NO, the background task will be killed immediately. If YES, the background task will not be destroyed until the `endBackgroundTask` method is called.
+/// @return Whether or not to wait for the subscriber to cleanup. If NO, the background task will be killed immediately. If YES, the background task will not be destroyed until the `endBackgroundTask` method is called by the subscriber.
 @property (copy, nonatomic, nullable) BOOL (^taskExpiringHandler)(void);
 
 - (instancetype)init NS_UNAVAILABLE;
--- a/SmartDeviceLink/SDLBackgroundTaskManager.m
+++ b/SmartDeviceLink/SDLBackgroundTaskManager.m
@@ -22,7 +22,7 @@ NS_ASSUME_NONNULL_BEGIN
 @implementation SDLBackgroundTaskManager
 
 - (instancetype)initWithBackgroundTaskName:(NSString *)backgroundTaskName {
-    SDLLogV(@"Creating manager with name %@", backgroundTaskName);
+    SDLLogV(@"Creating background task manager with name %@", backgroundTaskName);
     self = [super init];
     if (!self) {
         return nil;
@@ -42,7 +42,7 @@ NS_ASSUME_NONNULL_BEGIN
     __weak typeof(self) weakSelf = self;
     self.currentBackgroundTaskId = [[UIApplication sharedApplication] beginBackgroundTaskWithName:self.backgroundTaskName expirationHandler:^{
         __strong typeof(weakSelf) strongSelf = weakSelf;
-        SDLLogD(@"The %@ background task expired", strongSelf.backgroundTaskName);
+        SDLLogD(@"The background task %@ expired", strongSelf.backgroundTaskName);
         [strongSelf endBackgroundTask];
     }];
 
--- a/SmartDeviceLink/SDLBackgroundTaskManager.m
+++ b/SmartDeviceLink/SDLBackgroundTaskManager.m
@@ -22,7 +22,7 @@ NS_ASSUME_NONNULL_BEGIN
 @implementation SDLBackgroundTaskManager
 
 - (instancetype)initWithBackgroundTaskName:(NSString *)backgroundTaskName {
-    SDLLogV(@"SDLBackgroundTaskManager init with name %@", backgroundTaskName);
+    SDLLogV(@"Creating manager with name %@", backgroundTaskName);
     self = [super init];
     if (!self) {
         return nil;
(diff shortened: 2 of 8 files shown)


SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -682,29 +682,50 @@ - (void)sdl_onAppStateUpdated:(NSNotification *)notification {
...
+/// Handles a notification that the background task has ended. If the app is still in the background, the TCP transport disconnects. If the app has re-entered the foreground or the manager has shutdown then the notification is ignored.
+/// @return A background task ended handler
+- (nullable void (^)(void))sdl_backgroundTaskEndedHandler {
+    __weak typeof(self) weakSelf = self;
+    return ^{
+        __strong typeof(self) strongSelf = weakSelf;
+        if (strongSelf.sdl_getAppState == UIApplicationStateActive) {
+            SDLLogD(@"App has been foregrounded. Ignoring notification that the background task ended.");
+        } else if ([strongSelf.stateMachine isCurrentState:SDLSecondaryTransportStateStopped]) {
+            SDLLogD(@"Manager has been stopped. Ignoring notification that the background task ended.");
```
joeljfischer, line 721: I don't know that these are necessary. I think they can either be verbose logs or removed.

SmartDeviceLink/SDLBackgroundTaskManager.m
```diff
@@ -39,27 +39,34 @@ - (void)startBackgroundTask {
...
     }];
 
     SDLLogD(@"The %@ background task started with id: %lu", self.backgroundTaskName, (unsigned long)self.currentBackgroundTaskId);
 }
 
 - (void)endBackgroundTask {
+    if (self.taskEndedHandler != nil) {
+        self.taskEndedHandler();
+        self.taskEndedHandler = nil;
+    }
```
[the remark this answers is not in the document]
NicoleYarroch: I looked it up and you have to do all your cleanup before ending the the background task. My code did have a bug where the background task could be destroyed before the video, audio, and TCP transports managers were shutdown so I added a fix for that issue.

SmartDeviceLink/SDLSecondaryTransportManager.m
NicoleYarroch: I changed to verbose logs

SmartDeviceLink/SDLBackgroundTaskManager.m
```diff
@@ -41,26 +41,34 @@ - (void)startBackgroundTask {
...
     __weak typeof(self) weakSelf = self;
     self.currentBackgroundTaskId = [[UIApplication sharedApplication] beginBackgroundTaskWithName:self.backgroundTaskName expirationHandler:^{
+        // We have ~1 second to do cleanup before ending the background task. If we take too long, the system will kill the app.
         __strong typeof(weakSelf) strongSelf = weakSelf;
-        SDLLogD(@"The %@ background task expired", strongSelf.backgroundTaskName);
-        [strongSelf endBackgroundTask];
+        SDLLogD(@"The background task %@ expired", strongSelf.backgroundTaskName);
+        if (strongSelf.taskEndedHandler != nil) {
+            SDLLogD(@"Waiting for cleanup to finish before ending the background task");
+            strongSelf.taskEndedHandler();
```
joeljfischer, line 49: Doesn't this cause `endBackgroundTask` to never be set? It's not called from the `taskEndedHandler` in the `SecondaryTransportManager` which makes me think `endBackgroundTask` will ever be called in the `SecondaryTransportManager` if the task expires. If you were to require that the `endBackgroundTask` method be called from the handler, you should pass the `BackgroundTaskManager` as a parameter of the function.

SmartDeviceLink/SDLBackgroundTaskManager.m
```diff
@@ -41,26 +41,34 @@ - (void)startBackgroundTask {
...
 }
 
 - (void)endBackgroundTask {
+    SDLLogV(@"Attempting to end background task %@", self.backgroundTaskName);
+
     if (self.taskEndedHandler != nil) {
-        self.taskEndedHandler();
+        SDLLogD(@"Background task %@ cleanup finished", self.backgroundTaskName);
         self.taskEndedHandler = nil;
     }
```
joeljfischer: If this is called from the `taskEndedHandler` as I have mentioned above, then `nil`ing it out here seems invalid and the code below won't be run.

SmartDeviceLink/SDLStreamingAudioLifecycleManager.m
```diff
@@ -155,6 +155,11 @@ - (void)didEnterStateAudioStreamStopped {
     _audioEncrypted = NO;
 
     [[NSNotificationCenter defaultCenter] postNotificationName:SDLAudioStreamDidStopNotification object:nil];
+
+    if (self.audioServiceEndedCompletionHandler != nil) {
+        self.audioServiceEndedCompletionHandler();
+        self.audioServiceEndedCompletionHandler = nil;
+    }
```
joeljfischer, line 162: Is this a part of this PR? If so, can you describe what this change is in the PR description / changelog?

SmartDeviceLink/SDLStreamingVideoLifecycleManager.m
```diff
@@ -364,6 +364,11 @@ - (void)didEnterStateVideoStreamStopped {
     [self sdl_disposeDisplayLink];
 
     [[NSNotificationCenter defaultCenter] postNotificationName:SDLVideoStreamDidStopNotification object:nil];
+
+    if (self.videoServiceEndedCompletionHandler != nil) {
+        self.videoServiceEndedCompletionHandler();
+        self.videoServiceEndedCompletionHandler = nil;
+    }
```
joeljfischer, line 371: Same as above

SmartDeviceLink/SDLBackgroundTaskManager.m
NicoleYarroch: `endBackgroundTask` does get called in the `SDLSecondaryTransportManager`'s `disconnectSecondaryTransport` method. We need to wait for Core to ACK/NAK the end video and audio services and disconnect the TCP transport streams before ending the background task.

SmartDeviceLink/SDLBackgroundTaskManager.m
NicoleYarroch: `endBackgroundTask` does not get called from the `taskEndedHandler`. It only gets called when cleanup has finished in the `SecondaryTransportManager`'s `disconnectSecondaryTransport` method. We could add a method called `cleanupCompleted` or something like that to the background task manager.

SmartDeviceLink/SDLStreamingAudioLifecycleManager.m
NicoleYarroch: Added

SmartDeviceLink/SDLStreamingVideoLifecycleManager.m
NicoleYarroch: Added

SmartDeviceLink/SDLBackgroundTaskManager.m
joeljfischer: If that's the case, then `disconnectSecondaryTransport` must be called every single time that the handler is called, right? And that definitely will not happen in the `if` and `else if` cases.

I'm also curious why the `else` goes to `configured` but the log says "Disconnecting TCP transport due to the background task ending." Configured doesn't disconnect the transport, it re-starts it, right? It should go to `stopped` or `reconnecting`? 

If any of my assumptions are wrong here, then the code is difficult to follow. I would request a refactor or clear comments describing what it supposed to be happening here.

NicoleYarroch: Just so we are on the same page does the following statement refer to the `if:else` in the expiration handler?  Or are you talking about the `if:else` in the `taskEndedHandler` block?
  
> If that's the case, then disconnectSecondaryTransport must be called every single time that the handler is called, right? And that definitely will not happen in the if and else if cases.

joeljfischer: Sorry, in `sdl_backgroundTaskEndedHandler`

NicoleYarroch: The issue should be fixed now.

NicoleYarroch: > I'm also curious why the else goes to configured but the log says "Disconnecting TCP transport due to the background task ending." Configured doesn't disconnect the transport, it re-starts it, right? It should go to stopped or reconnecting?

I updated the logs with more information. Basically it means that the manager is shutting down the secondary transport but it is still configured (i.e. has the ip and port) to restart the secondary transport.

SmartDeviceLink/SDLBackgroundTaskManager.h
```diff
@@ -17,7 +17,7 @@ NS_ASSUME_NONNULL_BEGIN
 @interface SDLBackgroundTaskManager : NSObject
 
 /// Handler called when the background task is about to expire. Use this handler to perform some cleanup before the background task is destroyed. When you have finished cleanup, you must call the `endBackgroundTask` function so the background task can be destroyed. If you do not call `endBackgroundTask`, the system may kill the app.
-@property (copy, nonatomic, nullable) void (^taskEndedHandler)(void);
+@property (copy, nonatomic, nullable) BOOL (^taskExpiringHandler)(void);
```
joeljfischer, line 20: Document the retval

SmartDeviceLink/SDLBackgroundTaskManager.m
```diff
@@ -41,27 +41,38 @@ - (void)startBackgroundTask {
...
-            SDLLogD(@"Waiting for cleanup to finish before ending the background task");
-            strongSelf.taskEndedHandler();
+        SDLLogD(@"The background task %@ is expiring.", strongSelf.backgroundTaskName);
+
+        // We have ~1 second to do cleanup before ending the background task. If we take too long, the system will kill the app.
+        if (strongSelf.taskExpiringHandler != nil) {
+            SDLLogD(@"Checking if subscriber wants to to perform some cleanup before ending the background task %@", strongSelf.backgroundTaskName);
+            BOOL waitForCleanupToFinish = strongSelf.taskExpiringHandler();
+            if (waitForCleanupToFinish) {
+                SDLLogD(@"Subscriber wants to clean up before ending the background task %@. Waiting...", self.backgroundTaskName);
```
joeljfischer, line 52:
```suggestion
                SDLLogD(@"The subscriber will end background task itself %@. Waiting...", self.backgroundTaskName);
```

SmartDeviceLink/SDLBackgroundTaskManager.m
```diff
@@ -41,27 +41,38 @@ - (void)startBackgroundTask {
...
+            // No subscriber. Just end the background task.
+            SDLLogV(@"Ending background task %@", strongSelf.backgroundTaskName);
             [strongSelf endBackgroundTask];
         }
     }];
 
     SDLLogD(@"The %@ background task started with id: %lu", self.backgroundTaskName, (unsigned long)self.currentBackgroundTaskId);
 }
 
+- (void)expiredTaskCleanupFinished {
```
joeljfischer, line 67: What is the utility of this as opposed to `endBackgroundTask`. The only difference is that it checks that the `taskExpiringHandler` isn't `nil` first. I'm struggling to understand what the utility of that is. Can't the subscriber just call `endBackgroundTask`?

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -711,17 +711,20 @@ - (void)sdl_onAppStateUpdated:(NSNotification *)notification {
...
+        if (strongSelf.sdl_getAppState == UIApplicationStateActive || [strongSelf.stateMachine isCurrentState:SDLSecondaryTransportStateStopped]) {
+            SDLLogV(@"No cleanup needed since app has been foregrounded.");
+            return NO;
         } else if ([strongSelf.stateMachine isCurrentState:SDLSecondaryTransportStateStopped]) {
-            SDLLogV(@"Manager has been stopped. Ignoring notification that the background task ended.");
+            SDLLogV(@"No cleanup needed since manager has been stopped.");
+            return NO;
         } else {
-            SDLLogD(@"Disconnecting TCP transport due to the background task ending.");
+            SDLLogD(@"Performing cleanup due to the background task expiring: disconnecting the TCP transport.");
```
joeljfischer, line 725:
```suggestion
            // `endBackgroundTask` will be called when the secondary transport disconnects.
            SDLLogD(@"Performing cleanup due to the background task expiring: disconnecting the TCP transport.");
```

SmartDeviceLink/SDLSecondaryTransportManager.m
```diff
@@ -438,7 +438,7 @@ - (BOOL)disconnectSecondaryTransport {
     self.secondaryTransport = nil;
     self.secondaryProtocol = nil;
 
-    [self.backgroundTaskManager endBackgroundTask];
+    [self.backgroundTaskManager expiredTaskCleanupFinished];
```
joeljfischer, line 441: The task needs to be cleaned up even if the `if` above is true, right?

SmartDeviceLink/SDLSecondaryTransportManager.m
joeljfischer: I'm trying to trace how `disconnectSecondaryTransport` will be called through here. Somehow it seems to be that the TCP transport is cut off, which causes the protocol to be ended, which causes the audio / video service to end, which updates the `SDLStreamingMediaManager`. Can you help me understand how, from here, `disconnectSecondaryTransport` will be called by adding a comment here? I'm just envisioning a but in the secondary transport in the future and the pain caused by having to track this down without an over-abundance of documentation since this is so complicated.

SmartDeviceLink/SDLBackgroundTaskManager.m
NicoleYarroch: Yes, that is the way I had it setup before but you seemed confused by how everything is interacting so I thought I would separate it out into its own function.

SmartDeviceLink/SDLBackgroundTaskManager.h
```diff
@@ -35,6 +35,9 @@ NS_ASSUME_NONNULL_BEGIN
 /// Destroys the background task.
 - (void)endBackgroundTask;
 
+/// Destroys the background task after cleanup has finished. This should only be called if you have subscribed to the `taskExpiringHandler` in order to do some cleanup before the background task is destroyed.
```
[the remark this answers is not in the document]
NicoleYarroch: nope

SmartDeviceLink/SDLBackgroundTaskManager.h
NicoleYarroch: Fixed.

NicoleYarroch:
> While testing, I saw a few more issues:
> 
> 1. I consistently saw a background task ended that was unknown:
> 
> ```
> 2020-04-14 14:44:03.388147-0400 SDL Example[1175:377941] Can't end BackgroundTask: no background task exists with identifier 3 (0x3), or it may have already been ended. Break in UIApplicationEndBackgroundTaskError() to debug.
> ```
> 
> It doesn't seem to be causing an issue, but it does make me a bit suspicious.

I saw the same notifications and it seems to be a known iOS 13 issue. I created an empty single-view Swift app and I get the exact same log `Can't end BackgroundTask: no background task exists with identifier 2 (0x2), or it may have already been ended. Break in UIApplicationEndBackgroundTaskError() to debug.` every time I foreground  the app after backgrounding it.

SmartDeviceLink/SDLSecondaryTransportManager.m
NicoleYarroch: Reverted to using `endBackgroundTask`. 

If the `if` statement is `true` then either the `disconnectSecondaryTransport` method was called twice or the `secondaryTransport` was never setup and thus a background task would not have been created as a background task is only created if an active TCP session is in progress.

SmartDeviceLink/SDLBackgroundTaskManager.m
NicoleYarroch: I reverted to using `endBackgroundTask`.

SmartDeviceLink/SDLBackgroundTaskManager.h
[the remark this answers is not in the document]
NicoleYarroch: Removed

SmartDeviceLink/SDLSecondaryTransportManager.m
joeljfischer: @NicoleYarroch This needs internal comment documentation. It's very confusing to try to trace even now. Trying to figure out what's going on a year from now would be very difficult. Please update comment documentation to explain what is happening here in detail.

PR #1646, opened 2020-05-01
Replace uses of NSTimer with SDLTimer

### LINKED ISSUE #1611: Replace NSTimer with SDLTimer
### Bug Report
The `SDLTCPTransport` and `SDLTouchManager` both use `NSTimer`, which should be replaced by `SDLTimer`. 

* iOS Version: n/a
* SDL iOS Version: 6.5
* Testing Against: n/a

joeljfischer:
Fixes #1611 

### Risk

### Testing Plan

#### Unit Tests
Unit tests were run, no changes were needed

#### Core Tests
Run against Manticore to test `TCPTransport` changes.

Core version / branch / commit hash / module tested against: v6.0.1 (Manticore)
HMI name / version / branch / commit hash / module tested against: Generic HMI v0.7.2 (Manticore)

### Summary
This PR removes all uses of `NSTimer` and uses the `SDLTimer` wrapper instead.

### Changelog
n/a - all internal best practice, no known bugs fixed.

### Tasks Remaining:
n/a

### CLA

### DIFF
--- a/SmartDeviceLink/SDLTouchManager.m
+++ b/SmartDeviceLink/SDLTouchManager.m
@@ -18,6 +18,7 @@
 #import "SDLProxyListener.h"
 #import "SDLRPCNotificationNotification.h"
 #import "SDLStreamingVideoScaleManager.h"
+#import "SDLTimer.h"
 #import "SDLTouch.h"
 #import "SDLTouchCoord.h"
 #import "SDLTouchEvent.h"
@@ -74,7 +75,7 @@ static NSUInteger const MaximumNumberOfTouches = 2;
  *  @abstract
  *      Timer used for distinguishing between single & double taps.
  */
-@property (nonatomic, strong, nullable) NSTimer *singleTapTimer;
+@property (nonatomic, strong, nullable) SDLTimer *singleTapTimer;
 
 /*!
  *  @abstract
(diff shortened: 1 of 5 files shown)


SmartDeviceLink/SDLTouchManager.m
```diff
@@ -447,19 +448,22 @@ - (void)sdl_initializeSingleTapTimerAtPoint:(CGPoint)point {
         [self sdl_cancelSingleTapTimer];
     }
 
-    self.singleTapTimer = [NSTimer timerWithTimeInterval:self.tapTimeThreshold target:self selector:@selector(sdl_singleTapTimerCallback:) userInfo:@{@"point": [NSValue valueWithCGPoint:point]} repeats:NO];
-    [[NSRunLoop mainRunLoop] addTimer:self.singleTapTimer forMode:NSRunLoopCommonModes];
+    __weak typeof(self) weakSelf = self;
+    self.singleTapTimer = [[SDLTimer alloc] initWithDuration:self.tapTimeThreshold];
```
NicoleYarroch, line 452: If `self.tapTimeThreshold` is set to `0` then the `elapsedBlock` never gets called. This means that the second tap is always the second tap in a double tap regardless of the time between the taps.

joeljfischer: I changed SDLTimer to assert if a duration of `<= 0` is used

NicoleYarroch: I thought setting a duration of 0 disabled the double tap gesture? That is the current behavior, which is why I noticed that it stopped working. 

Should probably update the `tapTimeThreshold` documentation to state that `tapTimeThreshold` can't be less than 0.

joeljfischer: Disabling the double tap gesture with a `tapTimeThreshold` of 0 was never the case. From the `NSTimer` documentation:

> The number of seconds between firings of the timer. If ti is less than or equal to 0.0, this method chooses the nonnegative value of 0.1 milliseconds instead.

So it was just so small that you could never double tap in that time span. I updated the documentation.

PR #1651, opened 2020-05-07
Video Background String Not Showing

### LINKED ISSUE #1620: Video stream background string not shown on some devices when device locks
### Bug Report
On some older devices (seen on iPhone 6, iOS 12), the video stream background string does not show when the device locks while the app is still foregrounded. This appears to be a bug with the iOS Video Toolbox's `VTCompressionSessionCreate` handler failing when the device is locked.

A possible hack would be to encode the frame with the background string when the video stream manager starts and keep a reference to it instead of waiting to encode the frame when the device has been locked. 

##### Reproduction Steps
1. Connect a video streaming app to a TDK via BT or USB. Launch the SDL app and wait for video to start streaming.
2. While the app on the phone is still in the foreground, lock the phone.

##### Expected Behavior
The video stream background string should show since the app is no longer in the foreground.

##### Observed Behavior
The video stream background string does not show. Logs show that encoding of the background frames is not occurring. 

* iOS Version: iOS 12
* SDL iOS Version:6.5
* Testing Against: n/a

lapinskijw:
Fixes #1620   

### Risk

### Testing Plan

#### Core Tests

Core version / branch / commit hash / module tested against:
HMI name / version / branch / commit hash / module tested against:

### Summary
`encodeFrame:presentationTimestamp:` will now check for a `kVTInvalidSessionErr`. If the error occurs, `VTCompressionSessionCompleteFrames` is called to force the completion of the current frames. We then reset the compression session.

##### Bug Fixes
The frames for the video background string screen will now be forced to be encoded when the phone is locked and an `kVTInvalidSessionErr` occurs, the compression session is also reset at this time. This fixes the issue on older devices where the background string frames were not being encoded when phone was locked.

### Tasks Remaining:
- [x] Run tests

### CLA

### DIFF
--- a/SmartDeviceLink/SDLH264VideoEncoder.m
+++ b/SmartDeviceLink/SDLH264VideoEncoder.m
@@ -187,7 +187,6 @@ static NSDictionary<NSString *, id>* _defaultVideoEncoderSettings;
 
 - (BOOL)encodeSavedFrame:(CVImageBufferRef)imageBuffer {
     return [self encodeSavedFrame:imageBuffer presentationTimestamp:kCMTimeInvalid];
-
 }
 
 - (BOOL)encodeSavedFrame:(CVImageBufferRef)imageBuffer presentationTimestamp:(CMTime)presentationTimestamp {
--- a/SmartDeviceLink/SDLH264VideoEncoder.m
+++ b/SmartDeviceLink/SDLH264VideoEncoder.m
@@ -159,6 +159,7 @@ static NSDictionary<NSString *, id>* _defaultVideoEncoderSettings;
 
     OSStatus status = VTCompressionSessionEncodeFrame(_compressionSession, imageBuffer, presentationTimestamp, kCMTimeInvalid, NULL, (__bridge void *)self, NULL);
 
+    // HAX: [#1620](https://github.com/smartdevicelink/sdl_ios/issues/1620) On some older iPhones VTCompressionSessionEncodeFrame fails with a kVTInvalidSessionErr when the device is locked. Attempt to fix this by recreating the compression session.
     if (status == kVTInvalidSessionErr) {
         VTCompressionSessionCompleteFrames(_compressionSession, presentationTimestamp);
         [self sdl_resetCompressionSession];
--- a/SmartDeviceLink/SDLH264VideoEncoder.m
+++ b/SmartDeviceLink/SDLH264VideoEncoder.m
@@ -142,52 +142,6 @@ static NSDictionary<NSString *, id>* _defaultVideoEncoderSettings;
     }
 }
 
-- (void)encodeSavedFrame:(CVImageBufferRef)imageBuffer {
-    [self encodeSavedFrame:imageBuffer presentationTimestamp:kCMTimeInvalid];
-}
-
-- (void)encodeSavedFrame:(CVImageBufferRef)imageBuffer presentationTimestamp:(CMTime)presentationTimestamp {
-    if (!CMTIME_IS_VALID(presentationTimestamp)) {
-        int32_t timeRate = 30;
-        if (self.videoEncoderSettings[(__bridge NSString *)kVTCompressionPropertyKey_ExpectedFrameRate] != nil) {
-            timeRate = ((NSNumber *)self.videoEncoderSettings[(__bridge NSString *)kVTCompressionPropertyKey_ExpectedFrameRate]).intValue;
-        }
-
-        presentationTimestamp = CMTimeMake((int64_t)self.currentFrameNumber, timeRate);
-    }
-    self.currentFrameNumber++;
-
-    VTCompressionSessionEncodeFrameWithOutputHandler(_compressionSession, imageBuffer, presentationTimestamp, kCMTimeInvalid, NULL, NULL, ^(OSStatus status, VTEncodeInfoFlags infoFlags, CMSampleBufferRef  _Nullable sampleBuffer) {
-        if (status != noErr) {
-            SDLLogW(@"Error encoding video frame: %d", (int)status);
-            return;
-        }
-
-        if (sampleBuffer == NULL) {
-            return;
-        }
-
-        NSArray *nalUnits = [self.class sdl_extractNalUnitsFromSampleBuffer:sampleBuffer];
-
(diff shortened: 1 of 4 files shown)


SmartDeviceLink/SDLH264VideoEncoder.m
```diff
@@ -158,6 +158,7 @@ - (BOOL)encodeFrame:(CVImageBufferRef)imageBuffer presentationTimestamp:(CMTime)
     self.currentFrameNumber++;
 
     OSStatus status = VTCompressionSessionEncodeFrame(_compressionSession, imageBuffer, presentationTimestamp, kCMTimeInvalid, NULL, (__bridge void *)self, NULL);
+    VTCompressionSessionCompleteFrames(_compressionSession, presentationTimestamp);
```
NicoleYarroch, line 161: Alas, trying `VTCompressionSessionCompleteFrames` did not work.

I printed out the `status` returned by `VTCompressionSessionEncodeFrame` in `encodeFrame:presentationTimestamp:` and I got [-12903](https://www.osstatus.com/search/results?platform=all&framework=all&search=-12903). I tried resetting the compression session when I get that error and it fixed the issue (i.e. the background string message shows up as expected).

```
- (BOOL)encodeFrame:(CVImageBufferRef)imageBuffer presentationTimestamp:(CMTime)presentationTimestamp {
    ....

    OSStatus status = VTCompressionSessionEncodeFrame(_compressionSession, imageBuffer, presentationTimestamp, kCMTimeInvalid, NULL, (__bridge void *)self, NULL);
    VTCompressionSessionCompleteFrames(_compressionSession, presentationTimestamp);

    // New code
    if (status == kVTInvalidSessionErr) {
        [self sdl_resetCompressionSession];
    }

    return (status == noErr);
}
```

SmartDeviceLink/SDLH264VideoEncoder.m
```diff
@@ -159,6 +159,11 @@ - (BOOL)encodeFrame:(CVImageBufferRef)imageBuffer presentationTimestamp:(CMTime)
 
     OSStatus status = VTCompressionSessionEncodeFrame(_compressionSession, imageBuffer, presentationTimestamp, kCMTimeInvalid, NULL, (__bridge void *)self, NULL);
 
+    if (status == kVTInvalidSessionErr) {
+        VTCompressionSessionCompleteFrames(_compressionSession, presentationTimestamp);
```
NicoleYarroch, line 163: You can use `kCMTimeInvalid` instead of the timestamp. That way you can move `VTCompressionSessionCompleteFrames` to right before the compression session is invalidated (here and an in the `stop` method - maybe create a new method called `invalidateCompressionSession` since it is duplicate code) .

SmartDeviceLink/SDLH264VideoEncoder.m
```diff
@@ -159,6 +159,11 @@ - (BOOL)encodeFrame:(CVImageBufferRef)imageBuffer presentationTimestamp:(CMTime)
 
     OSStatus status = VTCompressionSessionEncodeFrame(_compressionSession, imageBuffer, presentationTimestamp, kCMTimeInvalid, NULL, (__bridge void *)self, NULL);
 
+    if (status == kVTInvalidSessionErr) {
```
NicoleYarroch, line 162: Add a comment here to explain why this code was added since it is a hack to fix a Video Toolbox bug. Something like:
> // HAX: On some older iPhones `VTCompressionSessionEncodeFrame` fails with a `kVTInvalidSessionErr` as soon as the device is locked. Attempt to fix this by recreating the compression session.

SmartDeviceLink/SDLH264VideoEncoder.m
lapinskijw: > That way you can move VTCompressionSessionCompleteFrames to right before the compression session is invalidated

Could you clarify what you mean by this? Do you mean to add the `invalidateCompressionSession` method call in the `stop` method and `sdl_resetCompressionSession` method? For example:
```
- (void)sdl_invalidateCompressionSession {
    VTCompressionSessionCompleteFrames(self.compressionSession, kCMTimeInvalid);
    VTCompressionSessionInvalidate(self.compressionSession);
}
```
This would be called in `stop` and also in `sdl_resetCompressionSession` where we are already invalidating the session:
```- (BOOL)sdl_resetCompressionSession {
    // Destroy the current compression session before attempting to create a new one. Otherwise the attempt to create a new compression session sometimes fails.
    if (self.compressionSession != NULL) {
        [self sdl_invalidateCompressionSession];
        CFRelease(self.compressionSession);
        self.compressionSession = NULL;
    }

    OSStatus status = VTCompressionSessionCreate(NULL, (int32_t)self.dimensions.width, (int32_t)self.dimensions.height, kCMVideoCodecType_H264, NULL, self.sdl_pixelBufferOptions, NULL, &sdl_videoEncoderOutputCallback, (__bridge void *)self, &_compressionSession);
    return (status == noErr);
}
```

NicoleYarroch: Sorry, I was talking about consolidating the code that destroys the the compression session like so:

```
// TODO: add documentation for this method
- (void)sdl_invalidateCompressionSession {
    if (self.compressionSession == NULL) { return; }

    VTCompressionSessionCompleteFrames(self.compressionSession, kCMTimeInvalid);
    VTCompressionSessionInvalidate(self.compressionSession);
    CFRelease(self.compressionSession);
    self.compressionSession = NULL;
}
```
which would be called by the `stop` and `sdl_resetCompressionSession` methods like so:

```
- (void)stop {
    _currentFrameNumber = 0;
    _timestampOffset = 0.0;

    [self sdl_invalidateCompressionSession];
}
```

```
/// Attempts to create a new VTCompressionSession using the dimensions passed when the video encoder was created and returns whether or not creating the new compression session was created successfully.
- (BOOL)sdl_resetCompressionSession {
    // Destroy the current compression session before attempting to create a new one. Otherwise the attempt to create a new compression session sometimes fails.
    [self sdl_invalidateCompressionSession];

    OSStatus status = VTCompressionSessionCreate(NULL, (int32_t)self.dimensions.width, (int32_t)self.dimensions.height, kCMVideoCodecType_H264, NULL, self.sdl_pixelBufferOptions, NULL, &sdl_videoEncoderOutputCallback, (__bridge void *)self, &_compressionSession);
    return (status == noErr);
}
```

PR #1665, opened 2020-05-26
Added empty check for menu cell voice commands array

### LINKED ISSUE #1648: If an empty array is set for voice commands on the SDLMenuCell the AddCommand is rejected
### Bug Report
If an empty array is set for `voiceCommands` when creating a `SDLMenuCell` the resulting `AddCommand` is rejected by core because you can not send an empty array. 

The screen manager should not set the `vrCommands` parameter on the `AddCommand` if the `SDLMenuCell`'s `voiceCommands` array is empty. 

##### Reproduction Steps
1. Create a `SDLMenuCell` with an empty array set for the `voiceCommands` parameter and use the screen manager to send the menu.
##### Expected Behavior
The menu item should show up on the HMI

##### Observed Behavior
The menu item is rejected by Core because of the empty VR command list

* iOS Version: n/a
* SDL iOS Version: 6.6
* Testing Against: n/a

lapinskijw:
Fixes #1648 

### Risk

### Testing Plan

#### Core Tests
1. Set the first `SDLMenuCell` in the SmartDeviceLink-Example-Swift app to have an empty `voiceCommands` array. 
2. Connect app to Core/Manticore
3. Observe that the `SDLMenuCell` appears in menu

Core version / branch / commit hash / module tested against: SYNC 3.0 Build 17276_DEVTEST
HMI name / version / branch / commit hash / module tested against: Manticore v0.7.2

### Summary
Added empty check to `SDLMenuCell.m` for the `voiceCommands` being passed in. If the array is empty, the `voiceCommands` will be set to nil.

##### Bug Fixes
* Fixes `SDLMenuCell` being rejected due to empty `voiceCommands` array.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLMenuCell.m
+++ b/SmartDeviceLink/SDLMenuCell.m
@@ -27,7 +27,7 @@ NS_ASSUME_NONNULL_BEGIN
 
     _title = title;
     _icon = icon;
-    _voiceCommands = voiceCommands;
+    _voiceCommands = ([voiceCommands count] == 0) ? nil : voiceCommands;
     _handler = handler;
 
     _cellId = UINT32_MAX;
--- a/SmartDeviceLink/SDLMenuCell.m
+++ b/SmartDeviceLink/SDLMenuCell.m
@@ -27,7 +27,7 @@ NS_ASSUME_NONNULL_BEGIN
 
     _title = title;
     _icon = icon;
-    _voiceCommands = ([voiceCommands count] == 0) ? nil : voiceCommands;
+    _voiceCommands = voiceCommands;
     _handler = handler;
 
     _cellId = UINT32_MAX;
--- a/SmartDeviceLink/SDLMenuManager.m
+++ b/SmartDeviceLink/SDLMenuManager.m
@@ -642,7 +642,7 @@ UInt32 const MenuCellIdMin = 1;
     params.position = @(position);
 
     command.menuParams = params;
-    command.vrCommands = cell.voiceCommands.count == 0 ? nil : cell.voiceCommands;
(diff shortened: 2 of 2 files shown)


SmartDeviceLink/SDLMenuCell.m
```diff
@@ -27,7 +27,7 @@ - (instancetype)initWithTitle:(NSString *)title icon:(nullable SDLArtwork *)icon
 
     _title = title;
     _icon = icon;
-    _voiceCommands = voiceCommands;
+    _voiceCommands = ([voiceCommands count] == 0) ? nil : voiceCommands;
```
joeljfischer, line 30: I think it would be better to do this in `SDLMenuManager sdl_commandForMenuCell:withArtwork:position`. That way if the developer sets an empty array and later assumes it won't be nil, they won't be surprised. That's very much an edge case, but I think it's better to handle this awkward case internally rather than externally facing to the developer.

Also dot-syntax is preferred for property accessors, so `[voiceCommands count]` can just be `voiceCommands.count`.

SmartDeviceLink/SDLMenuManager.m
```diff
@@ -642,7 +642,7 @@ - (SDLAddCommand *)sdl_commandForMenuCell:(SDLMenuCell *)cell withArtwork:(BOOL)
     params.position = @(position);
 
     command.menuParams = params;
-    command.vrCommands = cell.voiceCommands;
+    command.vrCommands = cell.voiceCommands.count == 0 ? nil : cell.voiceCommands;
```
joeljfischer, line 645:
```suggestion
    command.vrCommands = (cell.voiceCommands.count == 0) ? nil : cell.voiceCommands;
```

PR #1712, opened 2020-07-14
Fix a deadlock when restarting the lifecycle manager from managerDidDisconnect

### LINKED ISSUE #1710: Restarting the sdlManager crashes due to a race condition in the lifecycle manager
### Bug Report
When restarting the sdlManager, it crashes due to a race condition in the lifecycle manager.

##### Reproduction Steps
1. Connect the SDL app to sdl_core + sdl_hmi using a TCP connection. Launch the SDL app.
1. Open the menu and select "Exit UNAUTHORIZED_TRANSPORT_REGISTRATION". When the `SDLManager's` `managerDidDisconnect` delegate method is called, restart the `sdlManager`. 

##### Expected Behavior
The SDL app restarts.

##### Observed Behavior
The SDL app crashes due to going from state `Stopped` to state `Connected`.

The `SDLManager` is started and the `SDLLifecycleManager` transitions to state `started`. However, before the transition can complete, the `SDLRPCServiceDidConnect` notification comes in. Since the lifecycle state is still `stopped`, the app crashes in the `sdl_rpcServiceDidConnect` method because it attempts to transition to state `connected`, which is an invalid state transition.   

* iOS Version: n/a/
* SDL iOS Version: 6.6 (branch develop)
* Testing Against:  sdl_core (tag 5.1.3) and sdl_hmi (branch develop).

joeljfischer:
Fixes #1710 

### Risk

### Testing Plan

#### Unit Tests
Ran the unit tests, no new tests were added

#### Core Tests
Tested hard transport disconnect with a reconnection from `managerDidDisconnect`.

Core version / branch / commit hash / module tested against: 5.1.0
HMI name / version / branch / commit hash / module tested against: SDL HMI 5.1.0

### Summary
Changes startup queuing to only dispatch sync to the lifecycle queue if we're not already on the lifecycle queue.

##### Bug Fixes
* Fixed deadlock when restarting the `SDLManager` from the `managerDidDisconnect` callback.

### CLA

### DIFF
--- a/SmartDeviceLink/SDLLifecycleManager.m
+++ b/SmartDeviceLink/SDLLifecycleManager.m
@@ -190,10 +190,10 @@ NSString *const BackgroundTaskTransportName = @"com.sdl.transport.backgroundTask
     _mobileHMIStateHandler = [[SDLLifecycleMobileHMIStateHandler alloc] initWithConnectionManager:self];
 
     // Notifications
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_rpcServiceDidConnect) name:SDLRPCServiceDidConnect object:nil];
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_transportDidDisconnect) name:SDLTransportDidDisconnect object:nil];
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(hmiStatusDidChange:) name:SDLDidChangeHMIStatusNotification object:nil];
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(remoteHardwareDidUnregister:) name:SDLDidReceiveAppUnregisteredNotification object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sdl_rpcServiceDidConnect) name:SDLRPCServiceDidConnect object:_notificationDispatcher];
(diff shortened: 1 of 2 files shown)


SmartDeviceLink/SDLLifecycleManager.m
```diff
@@ -201,9 +201,9 @@ - (instancetype)initWithConfiguration:(SDLConfiguration *)configuration delegate
 }
 
 - (void)startWithReadyHandler:(SDLManagerReadyBlock)readyHandler {
-    dispatch_sync(_lifecycleQueue, ^{
+    [self sdl_runOnProcessingQueue:^{
```
NicoleYarroch, line 204: It turns out that I was not re-starting the exiting `SDLManager` object but rather creating a new `SDLManager` object. This creates a new lifecycle manager but the old lifecycle manager is still around. Both lifecycle managers get the `SDLRPCServiceDidConnect` notification but only one is in the `started` state, hence the crash. Perhaps it would be better to add the notifications in when the `SDLManager` is started and remove the notifications when the `SDLManager` is stopped?

joeljfischer: We don't currently support multiple `SDLManager`'s existing at the same time, and I think your "bug" is basically attempting to do that. You created a new manager and didn't deallocate the old one. I changed the lifecycle manager to listen only to its own notification dispatcher, but not everything in the manager will necessarily work.

I think your suggestion could potentially lead to other issues where notifications that should be received are not. We cannot just remove the notification handlers when the developer says `stop` because we'll miss notifications. We would have to add it into `sdl_stopManagersAndRestart:`, but only when the manager won't be reconnected. So that wouldn't cover your use-case if the old manager is restarted. I think that what you did is just unsupported.

PR #1753, opened 2020-08-13
Refactor Text & Graphic Manager to use queues

### LINKED ISSUE #1749: Refactor Text & Graphic Manager
### Best Practice
Currently, queuing of T&G updates is done in a very rudimentary way that is subject to race conditions and spaghetti code. For maintainability and future update reasons (such as template management), it is necessary to improve our handling of T&G updates.

joeljfischer:
Fixes #1749 

### Risk

### Testing Plan

#### Unit Tests
Updated manager unit tests and wrote tests for the operation

#### Core Tests
Tested with Obj-C and Swift example apps by turning on / off text as fast as possible on Manticore and Sync 3.4 TDK.

Core version / branch / commit hash / module tested against:
- Manticore (Core v6.1.1)
- Ford Sync 3.4 TDK (Build 19252_DEVTEST)

HMI name / version / branch / commit hash / module tested against:
- Manticore (Generic HMI v0.8.1)
- Ford Sync 3.4 TDK (Build 19252_DEVTEST)

### Summary
This PR refactors the Text and Graphic sub-manager of the screen manager to use queues like the soft button and choice set managers.

##### Enhancements
* Improved internal template text and graphic handling.

### Tasks Remaining:
N/A

### CLA


SmartDeviceLinkTests/DevAPISpecs/SDLTextAndGraphicUpdateOperationSpec.m
```diff
@@ -0,0 +1,929 @@
...
+                    expect(testOp.currentScreenData.secondaryGraphic).to(beNil());
+                    expect(testOp.currentScreenData.metadataTags.mainField1).toNot(beNil());
+                    expect(testOp.currentScreenData.metadataTags.mainField2).toNot(beNil());
+                    expect(testOp.currentScreenData.metadataTags.mainField3).toNot(beNil());
+                    expect(testOp.currentScreenData.metadataTags.mainField4).to(beNil());
+                });
+            });
+        });
+
+        // with four lines available
```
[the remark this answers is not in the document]
joeljfischer: The reason I put all of these comments is because I found that the best way to navigate a spec is to fold the blocks in Xcode. But, then you can't see the name of the block. So I started putting comments so that I can fold the blocks and still see what the block is. If you agree that that's a good method of navigating specs, I think we should start doing that as much as we can, at least in complex specs.

SmartDeviceLink/SDLTextAndGraphicManager.m
```diff
@@ -720,6 +336,7 @@ - (nullable SDLArtwork *)blankArtwork {
 - (void)sdl_displayCapabilityDidUpdate:(SDLSystemCapability *)systemCapability {
     // we won't use the object in the parameter but the convenience method of the system capability manager
     self.windowCapability = self.systemCapabilityManager.defaultMainWindowCapability;
+    [self sdl_updateTransactionQueueSuspended];
     
     // Auto-send an updated show
     if ([self sdl_hasData]) {
```
joeljfischer, line 342: I think that this may need to call `sdl_updateWithCompletionHandler` or else it won't be updated if it's not "dirty".

SmartDeviceLink/SDLTextAndGraphicManager.m
```diff
@@ -736,6 +353,7 @@ - (void)sdl_hmiStatusNotification:(SDLRPCNotificationNotification *)notification
 
     SDLHMILevel oldLevel = self.currentLevel;
     self.currentLevel = hmiStatus.hmiLevel;
+    [self sdl_updateTransactionQueueSuspended];
 
     // Auto-send an updated show if we were in NONE and now we are not
     if ([oldLevel isEqualToString:SDLHMILevelNone] && ![self.currentLevel isEqualToString:SDLHMILevelNone] && self.waitingOnHMILevelUpdateToUpdate) {
```
joeljfischer, line 359: This `if` is no longer necessary.

SmartDeviceLink/SDLTextAndGraphicUpdateOperation.m
```diff
@@ -0,0 +1,493 @@
...
+
+    return (templateSupportsPrimaryArtwork && !graphicMatchesExisting && graphicExists);
+}
+
+- (BOOL)sdl_shouldUpdateSecondaryImage {
+    BOOL templateSupportsSecondaryArtwork = [self.currentCapabilities hasImageFieldOfName:SDLImageFieldNameSecondaryGraphic];
+    BOOL graphicMatchesExisting = [self.currentScreenData.secondaryGraphic.value isEqualToString:self.updatedState.secondaryGraphic.name];
+    BOOL graphicExists = (self.updatedState.secondaryGraphic != nil);
+
+    // Cannot detect if there is a secondary image, so we'll just try to detect if there's a primary image and allow it if there is.
```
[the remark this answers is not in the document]
joeljfischer: Fixed to make the comment more accurate.

SmartDeviceLink/SDLTextAndGraphicUpdateOperation.m
```diff
@@ -0,0 +1,493 @@
...
+        return;
+    }
+
+    // Build a show with everything from `self.newState`, we'll pull things out later if we can.
+    SDLShow *fullShow = [[SDLShow alloc] init];
+    fullShow.alignment = self.updatedState.alignment;
+    fullShow = [self sdl_assembleShowText:fullShow];
+    fullShow = [self sdl_assembleShowImages:fullShow];
+
+    __weak typeof(self) weakSelf = self;
```
NicoleYarroch, line 73: If I change the current template and the display capabilities change, then a `Show` request is unnecessary sent. 

For example:
1. Set the template to `NON-MEDIA`.
1. Set the four text-fields: 
```
{
    alignment = CENTERED;
    mainField1 = "textfield 1";
    mainField2 = "textfield 2";
    mainField3 = "textfield 3";
    mainField4 = "textfield 4";
    mediaTrack = "";
    metadataTags = "{\n    mainField1 =     (\n    );\n    mainField2 =     (\n    );\n    mainField3 =     (\n    );\n    mainField4 =     (\n    );\n}";
    templateTitle = "";
}
```
3. Set the template to `DOUBLE_GRAPHIC_WITH_SOFTBUTTONS`. Since the template updated, an automatic update occurs in the `SDLTextAndGraphicManager`'s `sdl_displayCapabilityDidUpdate`. I would expect that a `Show` request would not get sent because no information has changed. However, a `Show` request gets sent resetting all the textfields:
```
{
    alignment = CENTERED;
    mainField1 = "";
    mainField2 = "";
    mainField3 = "";
    mainField4 = "";
    mediaTrack = "";
    metadataTags = "{\n}";
    templateTitle = "";
}
```

NicoleYarroch: In addition, if I set the template to `DOUBLE_GRAPHIC_WITH_SOFTBUTTONS` and then try to set `textField1` to a string, then I would expect that a `Show` request would not be sent because the template does not support textFields). However, this `Show` does get sent:
```
{
    alignment = CENTERED;
    mainField1 = "";
    mainField2 = "";
    mainField3 = "";
    mainField4 = "";
    mediaTrack = "";
    metadataTags = "{\n}";
    templateTitle = "";
}
```

joeljfischer: In regards to the first issue you note, this is correct behavior. Think about the reverse, going from double graphic to NON_MEDIA. If the developer set several text fields (which would then not be shown), then the user switched to the `NON_MEDIA` template, those text fields should show up. But they won't if the change you requested is made.

joeljfischer: Regarding the second issue, I think that this is legitimate, but I do worry about the complexity of the fix. Basically we would need to check if the developer set data in the `currentState` but the show that will be sent is the same as the current screen data. That will actually get pretty complicated and is fairly error prone. I think it's better to send the few bytes of data in this case rather than to try to do all that calculation.

We can't just check if it's blank, because that only works in some cases, and because sometimes a blank Show is intended to clear the screen.

PR #1758, opened 2020-08-25
RPC spec generator no longer cut text after @TODO

### LINKED ISSUE #1744: RPC spec generator strips text after @TODO
### Bug Report
The RPC spec generator is designed to strip any `@TODO`s from the documentation, however if the `@TODO` comes in the middle of a sentence, then the rest of text after the `@TODO` is not preserved.

The code to strip `@TODO`s is in the **common_producer.py** script in the `extract_description` method.

##### Reproduction Steps
1. Run the generator script in the "generator" folder of the library.

##### Expected Behavior
The `@TODO` should be stripped from the sentence. 

For example, in the RPC Spec for the `RegisterAppInterface` request, the `ttsName` description has a `@TODO`. 

```
<param name="ttsName" type="TTSChunk" minsize="1" maxsize="100" array="true" mandatory="false" since="2.0">
    <description>
        TTS string for VR recognition of the mobile application name, e.g. "My S D L App".
        Meant to overcome any failing on speech engine in properly pronouncing / understanding app name.
        Needs to be unique over all applications from the same device.
        May not be empty.
        May not start with a new line character.
        Only characters from char set [@TODO: Create char set (character/hex value) for each ACM and refer to] are supported.
    </description>
</param>
```

It should generate:
```
/**
 * TTS string for VR recognition of the mobile application name, e.g. "My S D L App". Meant to overcome any failing on speech engine in properly pronouncing / understanding app name. Needs to be unique over all applications from the same device. May not be empty. May not start with a new line character. Only characters from char set are supported
 *
 * @since SDL 2.0.0
 */
```

##### Observed Behavior
The rest of the sentence is stripped after `@TODO`

```
/**
 * TTS string for VR recognition of the mobile application name, e.g. "My S D L App". Meant to overcome any failing on speech engine in properly pronouncing / understanding app name. Needs to be unique over all applications from the same device. May not be empty. May not start with a new line character. Only characters from char set
 *
 * @since SDL 2.0.0
 */
```

* iOS Version: n/a
* SDL iOS Version: 6.7
* Testing Against: n/a

FrankElias77:
Fixes #1744

### Risk

### Testing Plan

#### Unit Tests
[N/A]

#### Core Tests
[N/A]

Core version / branch / commit hash / module tested against: [N/A]
HMI name / version / branch / commit hash / module tested against: [N/A]

### Summary
Bug fixes - Updated the generator to change the extract description method to no longer cut off after @TODO in RPC description.

##### Bug Fixes
Generated RPC descriptions no longer cut off after @TODO.

### Tasks Remaining:
- N/A

### CLA

### DIFF
--- a/generator/transformers/common_producer.py
+++ b/generator/transformers/common_producer.py
@@ -337,7 +337,7 @@ class InterfaceProducerCommon(ABC):
                 'mandatory': param.is_mandatory,
                 'deprecated': json.loads(param.deprecated.lower()) if param.deprecated else False,
                 'modifier': 'strong'}
-        if isinstance(param.param_type, (Integer, Float, String, Array, Boolean)):
+        if isinstance(param.param_type, (Integer, Float, String, Array)):
             data['description'].append(self.create_param_descriptor(param.param_type, OrderedDict()))
 
         data.update(self.extract_type(param))
--- a/generator/transformers/common_producer.py
+++ b/generator/transformers/common_producer.py
@@ -156,9 +156,10 @@ class InterfaceProducerCommon(ABC):
         """
         if not data:
             return []
-        if isinstance(data, list):
-            data = ' '.join(data)
-        return textwrap.wrap(re.sub(r'(\s{2,}|\n|\[@TODO.+)', ' ', data).strip(), length)
+#        if isinstance(data, list):
+#            data = ' '.join(data)
+#        return textwrap.wrap(re.sub(r'(\s{2,}|\n|\[@TODO.+)', ' ', data).strip(), length)
+        return data
 
     @staticmethod
     def nullable(type_native: str, mandatory: bool) -> str:
--- a/generator/transformers/common_producer.py
+++ b/generator/transformers/common_producer.py
@@ -156,9 +156,6 @@ class InterfaceProducerCommon(ABC):
         """
         if not data:
             return []
-#        if isinstance(data, list):
-#            data = ' '.join(data)
-#        return textwrap.wrap(re.sub(r'(\s{2,}|\n|\[@TODO.+)', ' ', data).strip(), length)
         return data
 
     @staticmethod
--- a/generator/transformers/common_producer.py
+++ b/generator/transformers/common_producer.py
@@ -156,7 +156,9 @@ class InterfaceProducerCommon(ABC):
         """
         if not data:
             return []
-        return data
+        if isinstance(data, list):
+            data = ' '.join(data)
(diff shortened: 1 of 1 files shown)


generator/transformers/common_producer.py
```diff
@@ -156,9 +156,7 @@ def extract_description(data, length: int = 2048) -> list:
         """
         if not data:
             return []
-        if isinstance(data, list):
-            data = ' '.join(data)
-        return textwrap.wrap(re.sub(r'(\s{2,}|\n|\[@TODO.+)', ' ', data).strip(), length)
```
NicoleYarroch, line 161: I think we still need to keep the `re.sub` method because it cleans up extra whitespace and newlines. `\s{2,}` strips extra whitespace, and `\n` strips newlines. I think we can just remove `\[@TODO.+`.

For example in the `GPSData` struct, the `shifted` parameter description has multiple lines of text separated by a newline. This PR is generating documentation with a lot of extra spaces: 

```
/**
 * 
                True, if GPS lat/long, time, and altitude have been purposefully shifted (requires a proprietary algorithm to unshift).
                False, if the GPS data is raw and un-shifted.
                If not provided, then value is assumed False.
            
 *
 * @since SDL 6.0.0
 */
```

When we want: 
```
/**
 * True, if GPS lat/long, time, and altitude have been purposefully shifted (requires a proprietary algorithm to unshift). False, if the GPS data is raw and un-shifted. If not provided, then value is assumed False.
 *
 * @since SDL 6.0.0
 */
```

generator/transformers/common_producer.py
```diff
@@ -156,9 +156,7 @@ def extract_description(data, length: int = 2048) -> list:
         """
         if not data:
             return []
-        if isinstance(data, list):
-            data = ' '.join(data)
```
NicoleYarroch, line 160: Is `.join(data)` not being used?

generator/transformers/common_producer.py
```diff
@@ -337,7 +337,7 @@ def extract_param(self, param: Param, item_name: str):
                 'mandatory': param.is_mandatory,
                 'deprecated': json.loads(param.deprecated.lower()) if param.deprecated else False,
                 'modifier': 'strong'}
-        if isinstance(param.param_type, (Integer, Float, String, Array)):
+        if isinstance(param.param_type, (Integer, Float, String, Array, Boolean)):
```
joeljfischer, line 340: What is the purpose of this change? The only thing it seems to do is add a `{}` to the `shifted` parameter, but since it has no data it doesn't seem to be helpful without additional work. It also doesn't seem to be related to this change, so I think it can be reverted.

PR #1766, opened 2020-09-02
Fixed generator history/deprecated documentation

### LINKED ISSUE #1761: RPC Generator should create correct history documentation
### Bug Report
Currently the generator for the RPC Spec will add the below documentation when the since attribute is available.
`@since SmartDeviceLink X.X`

When something is deprecated in the RPC spec this is still the case. This can be confusing because it could be interpreted and something being added and deprecated within the same version. We would like to change this behavior in the generator.
If the deprecated attribute is true and the since attribute is available the documentation should instead include:
`@deprecated in SmartDeviceLink X.X`

When a parameter is added, the documentation should state:
`@added in SmartDeviceLink X.X`

RPC Spec Mobile API Versioning Proposal: https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0089-mobile-api-versioning.md

##### Reproduction Steps
1. Run the generator
2. Look at `RegisterAppInterface(Response).displayCapabilities` documentation

##### Expected Behavior
Should state `@added in SmartDeviceLink 2.0` and `@deprecated in SmartDeviceLink 6.0`.

##### Observed Behavior
Does not state the above.

* iOS Version: n/a
* SDL iOS Version: n/a
* Testing Against: n/a

n/a

FrankElias77:
Fixes #1761

### Risk

### Testing Plan

#### Unit Tests
Generator only changes.

#### Core Tests
Generator only changes.

### Summary
Fixed the since label referring to deprecated instead of to when the rpc was added.

##### Breaking Changes
* N/A

##### Enhancements
* N/A

##### Bug Fixes
* Fixed the generator since label to when rpc was deprecated instead to when an rpc is added

### Tasks Remaining:
- N/A

### CLA


generator/templates/description.jinja2
```diff
@@ -8,14 +8,17 @@
...
+ {%- if since %}
+ * @added in SmartDeviceLink {{ since }}
  {%- endif %}
- {%- if history %}
- * @history SDL {{ history }}
  {%- endif %}
- {%- if since %}
- * @since SDL {{ since }}
+ {%- if deprecated is sameas true and history %}
+ * @added in SmartDeviceLink {{ history[0].since }}
```
NicoleYarroch, line 21: The `history[0].since` value does not appear to get printed here if a struct or enum was deprecated. 

### Reproduce
In `SDLDisplayType` I see:
```
 * @deprecated in SmartDeviceLink 5.0.0
 * @added in SmartDeviceLink 
```
1.0.0 is missing after `@added in SmartDeviceLink`

This is the RPC spec entry for `DisplayType`:
```
<enum name="DisplayType" deprecated="true" since="5.0">
    <description>See DAES for further infos regarding the displays</description>
    <element name="CID"/>
    <element name="TYPE2" />
    <element name="TYPE5" />
    <element name="NGN" />
    <element name="GEN2_8_DMA" since="3.0" />
    <element name="GEN2_6_DMA" since="3.0" />
    <element name="MFD3" since="2.0" />
    <element name="MFD4" since="2.0" />
    <element name="MFD5" since="2.0" />
    <element name="GEN3_8-INCH" internal_name="GEN3_8_INCH" since="3.0" />
    <element name="SDL_GENERIC" since="4.0" />
    <history>
        <enum name="DisplayType" since="1.0" until="5.0"/>
    </history>
</enum>
```

NicoleYarroch (changes requested):
One small fix requested.

NicoleYarroch (approved):

PR #1777, opened 2020-09-10
Rpc generator fix to generate optional floats

### LINKED ISSUE #1763: RPC generator does not properly generate optional float initializer values as NSNumbers
### Bug Report
When generating a file like `SDLGetVehicleDataResponse`, the initializer is not properly generating `float` values as optional `NSNumber<SDLFloat> *`.

FrankElias77:
Fixes #1763

### Risk

### Testing Plan

#### Unit Tests
N/A

#### Core Tests
N/A

### Summary
Fixed the generator init function that was not properly generating float optional values.

##### Breaking Changes
* N/A

##### Enhancements
* N/A

##### Bug Fixes
* Fixed the generator init function that was not properly generating float optional values.

### Tasks Remaining:
N/A

### CLA

### DIFF
--- a/generator/transformers/common_producer.py
+++ b/generator/transformers/common_producer.py
@@ -168,7 +168,7 @@ class InterfaceProducerCommon(ABC):
         :param mandatory: is parameter mandatory
         :return: string with modificator
         """
-        if mandatory or re.match(r'BOOL|double', type_native):
+        if mandatory or re.match(r'BOOL', type_native):
             return ''
         return 'nullable '
 
@@ -307,9 +307,7 @@ class InterfaceProducerCommon(ABC):
         else:
             data = self.evaluate_type(param.param_type)
 
-        if not param.is_mandatory and re.match(r'\w*Int\d*|BOOL', data['type_native']):
-            data['type_native'] = data['type_sdl']
-        elif not param.is_mandatory and re.match(r'\w*float\d*', data['type_native']):
+        if not param.is_mandatory and re.match(r'\w*Int\d*|BOOL|\w*float\d*', data['type_native']):
             data['type_native'] = data['type_sdl']
 
         return data
--- a/generator/transformers/common_producer.py
+++ b/generator/transformers/common_producer.py
@@ -168,7 +168,7 @@ class InterfaceProducerCommon(ABC):
         :param mandatory: is parameter mandatory
         :return: string with modificator
         """
-        if mandatory or re.match(r'BOOL|float|double', type_native):
+        if mandatory or re.match(r'BOOL|double', type_native):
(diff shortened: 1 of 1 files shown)


generator/transformers/common_producer.py
```diff
@@ -309,6 +309,8 @@ def extract_type(self, param: Param) -> dict:
 
         if not param.is_mandatory and re.match(r'\w*Int\d*|BOOL', data['type_native']):
             data['type_native'] = data['type_sdl']
+        elif not param.is_mandatory and re.match(r'\w*float\d*', data['type_native']):
```
NicoleYarroch, line 312: Can this `elif` be combined into one statement with the above `if` like so: `re.match(r'\w*Int\d*|BOOL|\w*float\d*', data['type_native'])`?

generator/transformers/common_producer.py
```diff
@@ -168,7 +168,7 @@ def nullable(type_native: str, mandatory: bool) -> str:
         :param mandatory: is parameter mandatory
         :return: string with modificator
         """
-        if mandatory or re.match(r'BOOL|float|double', type_native):
+        if mandatory or re.match(r'BOOL|double', type_native):
```
NicoleYarroch, line 171: I think we can also get rid of the `double` since it appears we only have `SDLFloat` in **NSNumber+NumberType.h**.

generator/transformers/common_producer.py
NicoleYarroch, line 312: Can this `elif` be combined into one statement with the above `if` like so: `re.match(r'\w*Int\d*|BOOL|\w*float\d*', data['type_native'])`?

NicoleYarroch (changes requested):
Left some comments

generator/transformers/common_producer.py
```diff
@@ -307,7 +307,7 @@ def extract_type(self, param: Param) -> dict:
         else:
             data = self.evaluate_type(param.param_type)
 
-        if not param.is_mandatory and re.match(r'\w*Int\d*|BOOL', data['type_native']):
+        if not param.is_mandatory and re.match(r'\w*Int\d*|BOOL|\w*float\d*', data['type_native']):
```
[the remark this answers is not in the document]
joeljfischer: Is that what we do for ints as well? I thought we took regular floats / ints instead of it's required.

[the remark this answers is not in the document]
NicoleYarroch: Oh, I get it now, we use `NSNumber` for optionals because primitives can't be set to `null`. Ignore my previous comment.

NicoleYarroch (approved):

PR #1806, opened 2020-10-06
[WIP] Fix: Possible infinity loop in iAP data session

### LINKED ISSUE #1799: High CPU usage in Energy Organizer, iOS 14+
### Bug Report

image
image
image

##### Reproduction Steps
TBD

### LINKED ISSUE #1809: Possible infinity loop with iAP run loop
### Bug Report
There is a possible infinity loop in the [SDLIAPDataSession sdl_accessoryEventLoop](https://github.com/smartdevicelink/sdl_ios/blob/7.0.0-rc.2/SmartDeviceLink/private/SDLIAPDataSession.m#L322). There is no confirmed bug to this issue. However other reports show apps freaking out causing high CPU load possible due to this or another infinity loop.

A planned fix should add more checks to prevent a possible infinity loop. Should be no harm to regular use case but cancel the loop if the iAP session is cancelled.

* iOS Version: iOS 14.0
* SDL iOS Version: 7.0.0 RC2

kshala-ford:
Fixes #1809 

### Risk

### Testing Plan

I have not run the tests against Core/Manticore but against a set of different Ford SYNC systems to properly test with iAP.

#### Unit Tests
Unit tests to simulate the situation are planned. Feasibility must be confirmed because lot of mock classes might need to be created.

### Summary
Adds a break to cancel a possible infinity loop if the run loop is empty. Logs out details to the input and output stream and also the accessory.

##### Bug Fixes
* Adds a break to cancel a possible infinity loop.
* Adds setting the thread property to nil if it is being cancelled.

### Tasks Remaining:
- [x] [Test]
- [ ] [Test further]

### CLA

### DIFF
--- a/SmartDeviceLink/private/SDLIAPDataSession.m
+++ b/SmartDeviceLink/private/SDLIAPDataSession.m
@@ -102,6 +102,8 @@ NS_ASSUME_NONNULL_BEGIN
         BOOL cancelledSuccessfully = [strongSelf sdl_isIOThreadCancelled];
         if (!cancelledSuccessfully) {
             SDLLogE(@"The I/O streams were not shut down successfully. We might not be able to create a new session with an accessory during the same app session. If this happens, only force quitting and restarting the app will allow new sessions.");
+        } else {
+            strongSelf.ioStreamThread = nil;
         }
 
         [strongSelf.sendDataQueue removeAllObjects];
@@ -333,7 +335,11 @@ NS_ASSUME_NONNULL_BEGIN
 
         while (self.ioStreamThread != nil && !self.ioStreamThread.cancelled) {
             // Enqueued data will be written to and read from the streams in the runloop
-            [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.25f]];
+            BOOL result = [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.25f]];
+            if (!result) {
+                SDLLogE(@"The run loop returned immediately althoug two streams supposed to be scheduled.\nInput stream: %@\nOutput stream: %@\nAccessory: %@", self.eaSession.inputStream, self.eaSession.outputStream, self.accessory);
+                break;
+            }
         }
 
         SDLLogD(@"Closing the accessory event loop on thread: %@", NSThread.currentThread.name);
--- a/SmartDeviceLink/private/SDLIAPDataSession.m
+++ b/SmartDeviceLink/private/SDLIAPDataSession.m
@@ -6,6 +6,8 @@
 //  Copyright © 2019 smartdevicelink. All rights reserved.
 //
 
+#import <UIKit/UIDevice.h>
+
 #import "SDLIAPDataSession.h"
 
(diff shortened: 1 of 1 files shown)


SmartDeviceLink/private/SDLIAPDataSession.m
```diff
@@ -102,6 +104,8 @@ - (void)destroySessionWithCompletionHandler:(void (^)(void))disconnectCompletion
         BOOL cancelledSuccessfully = [strongSelf sdl_isIOThreadCancelled];
         if (!cancelledSuccessfully) {
             SDLLogE(@"The I/O streams were not shut down successfully. We might not be able to create a new session with an accessory during the same app session. If this happens, only force quitting and restarting the app will allow new sessions.");
+        } else {
+            strongSelf.ioStreamThread = nil;
```
thedavidharris, line 108: This is another thing I was looking for but couldn't figure out, the runloop logic is checking for if the ioStreamThread isn't nil but didn't ever see it getting set to nil here, just cancelled. We've been working on an assumption here that this is triggered when there are two connections going on, and the two cause a deadlock. Unsure on the trigger, but if https://github.com/smartdevicelink/sdl_ios/pull/1806/files#diff-f127f5f9119210b4566551d521031b0aL71 were called from different threads I don't think that operation is atomic either which would allow for it if I'm reading it right?

SmartDeviceLink/private/SDLIAPDataSession.m
```diff
@@ -330,10 +334,19 @@ - (void)sdl_accessoryEventLoop {
...
+        
         while (self.ioStreamThread != nil && !self.ioStreamThread.cancelled) {
             // Enqueued data will be written to and read from the streams in the runloop
-            [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.25f]];
+            BOOL result = [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.25f]];
+            if (!result && !runModeFailedOnce) {
+                // log error once on any iOS version but only break if on iOS 14 or up.
+                runModeFailedOnce = YES;
+                SDLLogE(@"The run loop returned immediately although two streams supposed to be scheduled.\nInput stream: %@\nOutput stream: %@\nAccessory: %@", self.eaSession.inputStream, self.eaSession.outputStream, self.accessory);
+                if (SDL_SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"14")) {
```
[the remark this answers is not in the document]
kshala-ford: I would like to remove this check altogether but it should make sense to focus to the group which we believe is affected. I can change that to 13.7 if you like.

SmartDeviceLink/private/SDLIAPDataSession.m
kshala-ford: Setting to nil makes sense because it might have impact loop's conditions. Also it fixes misleading error messages if "cancelldSuccessfully" is set to NO because the thread is already cancelled hence cannot send the semaphore signal.

I'm not so sure if we have invalidated this assumption in the separate discussion we had. Two connections should  lead to two EA sessions with separate streams and separate threads. A deadlock should not be possible with the SDL architecture. Please let me know if it's still valid.

kshala-ford:
This PR could be closed in favor of #1815

PR #1807, opened 2020-10-07
Change SDLLockScreenViewController modalPresentationStyle

### LINKED ISSUE #1794: Update SDLLockScreenViewController to be `modalPresentationStyle` == `UIModalPresentationFullScreen` by default
### Bug Report
The default presentation style, automatic, is dismissible in iOS 13 / 14, so we need to use `.fullscreen` by default for custom lock screens.

##### Reproduction Steps
1. Create a subclass of `SDLLockScreenViewController` and set it to be the custom lock screen
2. Present the lock screen

##### Expected Behavior
The lock screen is not dismissible and appears full-screen.

##### Observed Behavior
The lock screen is dismissible and appears non-full-screen

* iOS Version: 14.0.1
* SDL iOS Version: 7.0.0-develop
* Testing Against: n/a

FrankElias77:
Fixes #1794 

### Risk

### Testing Plan

#### Unit Tests
N/A

#### Core Tests
N/A

### Summary
Changed SDLLockScreenViewController modalPresentationStyle that was dismissed by default in case the user haven't set one.

##### Breaking Changes
* N/A

##### Enhancements
* N/A

##### Bug Fixes
* Changed SDLLockScreenViewController modalPresentationStyle that was dismissed by default in case the user haven't set one.

### Tasks Remaining:
N/A

### CLA

### DIFF
--- a/SmartDeviceLink/public/SDLLockScreenViewController.h
+++ b/SmartDeviceLink/public/SDLLockScreenViewController.h
@@ -17,6 +17,11 @@ NS_ASSUME_NONNULL_BEGIN
 typedef void (^SwipeGestureCallbackBlock)(void);
 
 /**
+ *  The SDL Lock Screen's presentation style.
+ */
+@property(nonatomic, assign) UIModalPresentationStyle modalPresentationStyle;
+
+/**
  *  The app's icon. This will be set by the lock screen configuration.
  */
 @property (copy, nonatomic, nullable) UIImage *appIcon;
--- a/SmartDeviceLink/public/SDLLockScreenViewController.h
+++ b/SmartDeviceLink/public/SDLLockScreenViewController.h
@@ -17,11 +17,6 @@ NS_ASSUME_NONNULL_BEGIN
 typedef void (^SwipeGestureCallbackBlock)(void);
 
 /**
- *  The SDL Lock Screen's presentation style.
- */
-@property(nonatomic, assign) UIModalPresentationStyle modalPresentationStyle;
-
-/**
  *  The app's icon. This will be set by the lock screen configuration.
  */
 @property (copy, nonatomic, nullable) UIImage *appIcon;
--- a/SmartDeviceLink/public/SDLLockScreenViewController.m
+++ b/SmartDeviceLink/public/SDLLockScreenViewController.m
@@ -32,11 +32,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 
 @implementation SDLLockScreenViewController
-
-- (void)awakeFromNib {
-    [super awakeFromNib];
-    self.modalPresentationStyle = UIModalPresentationFullScreen;
-}
+@synthesize modalPresentationStyle = _modalPresentationStyle;
 
 - (void)viewWillAppear:(BOOL)animated {
     [super viewWillAppear:animated];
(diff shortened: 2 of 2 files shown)


SmartDeviceLink/public/SDLLockScreenViewController.m
```diff
@@ -33,6 +33,11 @@ @interface SDLLockScreenViewController ()
 
 @implementation SDLLockScreenViewController
 
+- (void)awakeFromNib {
+    [super awakeFromNib];
+    self.modalPresentationStyle = UIModalPresentationFullScreen;
+}
```
joeljfischer, line 39: Override the `modalPresentationStyle` method instead of this

joeljfischer (changes requested):

SmartDeviceLink/public/SDLLockScreenViewController.h
```diff
@@ -16,6 +16,11 @@ NS_ASSUME_NONNULL_BEGIN
 /// A block that can be used to close the lockscreen when the user swipes on the lockscreen. Override this in your own custom view controllers if you build a custom lock screen.
 typedef void (^SwipeGestureCallbackBlock)(void);
 
+/**
+ *  The SDL Lock Screen's presentation style.
+ */
+@property(nonatomic, assign) UIModalPresentationStyle modalPresentationStyle;
+
```
joeljfischer, line 23: This is not correct. You are not re-defining a property, you are overriding it. Did you test this? I don't think this would work.

```suggestion
```

SmartDeviceLink/public/SDLLockScreenViewController.m
```diff
@@ -32,6 +32,7 @@ @interface SDLLockScreenViewController ()
 
 
 @implementation SDLLockScreenViewController
+@synthesize modalPresentationStyle = _modalPresentationStyle;
```
joeljfischer, line 35: This is not correct. Did you test this? I don't think it would work

```suggestion
```

SmartDeviceLink/public/SDLLockScreenViewController.m
```diff
@@ -61,6 +62,10 @@ - (UIStatusBarStyle)preferredStatusBarStyle {
     return useWhiteIcon ? UIStatusBarStyleLightContent : UIStatusBarStyleDefault;
 }
 
+- (UIModalPresentationStyle)modalPresentationStyle {
+    return _modalPresentationStyle;
+}
```
joeljfischer, line 67: This is not correct. You are not trying to re-define it, you're trying to override it.

```suggestion
- (UIModalPresentationStyle)modalPresentationStyle {
    return UIModalPresentationFullScreen;
}
```

SmartDeviceLink/public/SDLLockScreenViewController.m
```diff
@@ -87,6 +92,18 @@ - (void)setLockedLabelText:(NSString *_Nullable)lockedLabelText {
...
+        if (modalPresentationStyle == UIModalPresentationAutomatic) {
+            _modalPresentationStyle = UIModalPresentationFullScreen;
+        } else {
+            _modalPresentationStyle = modalPresentationStyle;
+        }
+    } else {
+        _modalPresentationStyle = modalPresentationStyle;
+    }
+}
+
```
joeljfischer, line 106: This seems unnecessary to me. We just override it to be `UIModalPresentationFullScreen`. The dev should be subclassing this and can override it if they want something different.

```suggestion
```

joeljfischer (changes requested):
You're still doing a lot of unnecessary work. Please just do what I mentioned here.

SmartDeviceLink/public/SDLLockScreenViewController.h
joeljfischer, line 23: This is not correct. You are not re-defining a property, you should be overriding it.

```suggestion
```

joeljfischer (changes requested):

joeljfischer (approved):

PR #1815, opened 2020-10-16
Bugfix/battery drain issue

### LINKED ISSUE #1799: High CPU usage in Energy Organizer, iOS 14+

kshala-ford:
Fixes #1799 and #1809

### Risk

### Testing Plan
- [ ] I have verified that I have not introduced new warnings in this PR (or explain why below)

#### Unit Tests
TBD

Note: Unit tests to simulate the EA streams are planned. Feasibility must be confirmed because lot of mock classes might need to be created.

#### Core Tests
Only tested against various Ford SYNC3 systems.

### Summary
This PR is a mirror of a Ford internal fork. The goal is to improve stability within the SDL library regarding EA session and stream lifecycle. 

##### Bug Fixes
- Adds a force close for EA sessions upon EA disconnect notifications.
- Adds retry mechanism if IO thread could not be cancelled.
- Fixes a possible infinity loop that could cause a 100% load on the IO thread.
- Prevents a spam of EA connect notifications to execute further.
- Ends existing background task identifier upon EA disconnect notifications.

### Tasks Remaining:
- [ ] [Code review]
- [ ] [Write unit tests when applicable]
- [ ] [Test against other EA devices and make sure no regression is introduced]

### CLA

### DIFF
--- a/SmartDeviceLink/private/SDLIAPDataSession.m
+++ b/SmartDeviceLink/private/SDLIAPDataSession.m
@@ -16,10 +16,14 @@
 #import "SDLIAPSession.h"
 #import "SDLLogMacros.h"
 #import "SDLMutableDataQueue.h"
+#import "SDLLifecycleManager.h"
 
 NSString *const IOStreamThreadName = @"com.smartdevicelink.iostream";
 NSTimeInterval const IOStreamThreadCanceledSemaphoreWaitSecs = 1.0;
 NSTimeInterval const IOStreamThreadRetryWaitSecs = 10.0;
+int const IOSStreamCancelRetryCountMax = 10;
+int IOSStreamCancelRetryCount = 0;
+
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -101,7 +105,7 @@ NS_ASSUME_NONNULL_BEGIN
         return disconnectCompletionHandler();
     }
     
-    dispatch_block_t cancelBlock = ^{
+    dispatch_block_t __block cancelBlock = ^{
         // Attempt to cancel the ioStreamThread. Once the thread realizes it has been cancelled, it will cleanup the I/O streams. Make sure to wake up the run loop in case there is no current I/O event running on the ioThread.
         [self.ioStreamThread cancel];
         [self performSelector:@selector(sdl_doNothing) onThread:self.ioStreamThread withObject:nil waitUntilDone:NO];
@@ -110,8 +114,13 @@ NS_ASSUME_NONNULL_BEGIN
         BOOL cancelledSuccessfully = [self sdl_isIOThreadCancelled];
         if (!cancelledSuccessfully) {
             SDLLogE(@"The I/O streams were not shut down successfully. We might not be able to create a new session with an accessory during the same app session. If this happens, only force quitting and restarting the app will allow new sessions.");
+            if (IOSStreamCancelRetryCount >= IOSStreamCancelRetryCountMax) {
+                return disconnectCompletionHandler();
+            }
+            IOSStreamCancelRetryCount += 1;
+
             // this will retry the block after the given number of seconds.
-            dispatch_after(dispatch_time(DISPATCH_TIME_NOW, IOStreamThreadRetryWaitSecs * NSEC_PER_SEC), dispatch_get_main_queue(), cancelBlock);
+            dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(IOStreamThreadRetryWaitSecs * NSEC_PER_SEC)), dispatch_get_main_queue(), cancelBlock);
         } else {
             self.ioStreamThread = nil;
         }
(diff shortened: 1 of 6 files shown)


SmartDeviceLink/private/SDLIAPDataSession.m
```diff
@@ -125,6 +144,34 @@ - (BOOL)sdl_isIOThreadCancelled {
...
+            [strongSelf performSelector:@selector(sdl_retryThreadCancel) withObject:nil afterDelay:IOStreamThreadRetryWaitSecs];
+            [strongSelf sdl_retryThreadCancel];
+        } else {
+            strongSelf.ioStreamThread = nil;
+        }
+
+        [strongSelf.sendDataQueue removeAllObjects];
+    });
+}
+
```
kshala-ford, line 174: As already mentioned for `destroySessionWithCompletionHandler` this method is not necessary. It's a copy and instead the destroy method should be called.
```suggestion
```

SmartDeviceLink/private/SDLIAPDataSession.m
```diff
@@ -331,9 +378,20 @@ - (void)sdl_accessoryEventLoop {
...
         while (self.ioStreamThread != nil && !self.ioStreamThread.cancelled) {
+            SDLLogD(@"In the Looper!");
             // Enqueued data will be written to and read from the streams in the runloop
-            [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.25f]];
+            BOOL result = [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.25f]];
+            if (!result && !runModeFailedOnce) {
+                // log error once on any iOS version but only break if on iOS 14 or up.
+                runModeFailedOnce = YES;
+                SDLLogE(@"The run loop returned immediately although two streams supposed to be scheduled.\nInput stream: %@\nOutput stream: %@\nAccessory: %@", self.eaSession.inputStream, self.eaSession.outputStream, self.accessory);
+                if (SDL_SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"14")) {
```
kshala-ford, line 391: We should consider to abort the loop for any iOS version.

SmartDeviceLink/private/SDLIAPTransport.m
```diff
@@ -92,7 +96,13 @@ - (void)sdl_stopEventListening {
...
 - (void)sdl_accessoryConnected:(NSNotification *)notification {
     EAAccessory *newAccessory = notification.userInfo[EAAccessoryKey];
-
+    double timeNow = NSTimeIntervalSince1970;
+    double timeSinceLastConnectionAttempt = timeNow - _eaAccessoryConnectedTime;
+    if (timeSinceLastConnectionAttempt < TimeRequiredBetweenConnectionAttempts) {
+        SDLLogD(@"Accessory Connecting too soon after previous attempt");
+        return;
+    }
+    _eaAccessoryConnectedTime = timeNow;
```
kshala-ford, line 105: I can't remember any rationale for this change. Ignoring a notification can be very dangerous leading to no app connecting. Can you give us a reason why this code is needed?

SmartDeviceLink/private/SDLLifecycleManager.m
```diff
@@ -192,9 +192,29 @@ - (instancetype)initWithConfiguration:(SDLConfiguration *)configuration delegate
...
+        SDLLogD(@"Accessory with serial number: %@, and connectionID: %lu disconnecting.", accessory.serialNumber, (unsigned long)accessory.connectionID);
+        if (self.backgroundTaskManager != nil) {
+            [self.backgroundTaskManager endBackgroundTask];
+        }
+        if (self.secondaryTransportManager != nil) {
+            [_secondaryTransportManager endBackgroundTask];
+        }
+    }
+}
+
```
kshala-ford, line 217: I have not looked into the code if ackground tasks are ended elsewhere. I have the impression the iOS bug might be caused by task expiration not being handled properly. So the idea to end them when not needed anymore is good. However ending them without being in sync with the disconnect process is dangerous as the thread might survive and run at a 100% whenever the app's process comes back to life.

kshala-ford (commented):
I left some comments to the PR. Please let me know your thoughts.

joeljfischer (changes requested):

SmartDeviceLink/private/SDLIAPDataSession.m
```diff
@@ -83,33 +85,50 @@ - (void)startSession {
...
-    __weak typeof(self) weakSelf = self;
     dispatch_async(dispatch_get_main_queue(), ^{
-        __strong typeof(weakSelf) strongSelf = weakSelf;
 
         // Attempt to cancel the ioStreamThread. Once the thread realizes it has been cancelled, it will cleanup the I/O streams. Make sure to wake up the run loop in case there is no current I/O event running on the ioThread.
-        [strongSelf.ioStreamThread cancel];
-        [strongSelf performSelector:@selector(sdl_doNothing) onThread:self.ioStreamThread withObject:nil waitUntilDone:NO];
+        [self.ioStreamThread cancel];
+        [self performSelector:@selector(sdl_doNothing) onThread:self.ioStreamThread withObject:nil waitUntilDone:NO];
```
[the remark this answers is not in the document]
kshala-ford: Running Xcode analysis doesn't show any issue with this attempt. Looks like it's pretty safe to call self in a block immediately dispatched asynchronously.

SmartDeviceLink/private/SDLIAPDataSession.m
kshala-ford: OK. I removed the version check

joeljfischer:
Fixed by #1910

PR #1831, opened 2020-10-27
Fix for template change

### LINKED ISSUE #1827: Example app template reset when HMI level changes
### Bug Report
In both the Swift and Obj-C example apps, if the user changed the template via the menu **Menu > Change Template** the template is reset back to `nonMedia` when the `hmiLevel` changes from `FULL`->`BACKGROUND`->`FULL`.

##### Reproduction Steps
1. Launch the app. The template will be `nonMedia`. 
1. Go to **Menu > Change Template** and select **Graphic With Text**. The menu will close and the template will update.
1. Close the app (do not kill the app). The app's `hmiLevel` will change to `BACKGROUND`. 
1. Open the app.  The app's `hmiLevel` will change to `FULL`.

##### Expected Behavior
The app should still be using the `graphicWithText` template.

##### Observed Behavior
The app is reset to use the `nonMedia` template.

* iOS Version: n/a
* SDL iOS Version: 7.0
* Testing Against: n/a

FrankElias77:
Fixes #1827 

### Risk

### Testing Plan

#### Unit Tests
N/A

#### Core Tests
N/A

### Summary
Fixed template issue, when the user switched HMI level from full to background and the back to full, the template was being reset to the default one(nonMedia)

##### Breaking Changes
* N/A

##### Enhancements
* N/A

##### Bug Fixes
* Fixed template issue, when the user switched HMI level from full to background and the back to full, the template was being reset to the default one(nonMedia)

### Tasks Remaining:
N/A

### CLA

### DIFF
--- a/Example Apps/Example Swift/MenuManager.swift
+++ b/Example Apps/Example Swift/MenuManager.swift
@@ -15,6 +15,7 @@ typealias RefreshTemplateHandler = ((SDLPredefinedLayout) -> Void)
 class MenuManager: NSObject {
     fileprivate var refreshTemplateHandler: RefreshTemplateHandler?
 
+    /// Setting current selected template
     public fileprivate(set) var currentTemplate: SDLPredefinedLayout {
         didSet {
             guard let refreshTemplateHandler = refreshTemplateHandler else { return }
--- a/Example Apps/Example Swift/MenuManager.swift
+++ b/Example Apps/Example Swift/MenuManager.swift
@@ -10,15 +10,28 @@ import Foundation
 import SmartDeviceLink
 import SmartDeviceLinkSwift
 
-class MenuManager: NSObject {
+typealias RefreshTemplateHandler = ((SDLPredefinedLayout) -> Void)
 
-    static var currentTemplate : SDLPredefinedLayout = .nonMedia
+class MenuManager: NSObject {
+    fileprivate var refreshTemplateHandler: RefreshTemplateHandler?
+
+    public fileprivate(set) var currentTemplate: SDLPredefinedLayout {
+        didSet {
+            guard let refreshTemplateHandler = refreshTemplateHandler else { return }
+            refreshTemplateHandler(currentTemplate)
+        }
+    }
+
+    init(updateTemplate: RefreshTemplateHandler? = nil) {
+        self.refreshTemplateHandler = updateTemplate
+        currentTemplate = .nonMedia
+    }
 
     /// Creates and returns the menu items
     ///
     /// - Parameter manager: The SDL Manager
     /// - Returns: An array of SDLAddCommand objects
-    class func allMenuItems(with manager: SDLManager, choiceSetManager: PerformInteractionManager) -> [SDLMenuCell] {
+    func allMenuItems(with manager: SDLManager, choiceSetManager: PerformInteractionManager) -> [SDLMenuCell] {
         return [menuCellSpeakName(with: manager),
                 menuCellGetAllVehicleData(with: manager),
                 menuCellShowPerformInteraction(with: manager, choiceSetManager: choiceSetManager),
@@ -34,7 +47,7 @@ class MenuManager: NSObject {
     ///
     /// - Parameter manager: The SDL Manager
     /// - Returns: An array of SDLVoiceCommand objects
-    class func allVoiceMenuItems(with manager: SDLManager) -> [SDLVoiceCommand] {
+    func allVoiceMenuItems(with manager: SDLManager) -> [SDLVoiceCommand] {
         guard manager.systemCapabilityManager.vrCapability else {
             SDLLog.e("The head unit does not support voice recognition")
             return []
@@ -50,7 +63,7 @@ private extension MenuManager {
     ///
     /// - Parameter manager: The SDL Manager
     /// - Returns: A SDLMenuCell object
-    class func menuCellSpeakName(with manager: SDLManager) -> SDLMenuCell {
+    func menuCellSpeakName(with manager: SDLManager) -> SDLMenuCell {
         return SDLMenuCell(title: ACSpeakAppNameMenuName, icon: SDLArtwork(image: UIImage(named: SpeakBWIconImageName)!.withRenderingMode(.alwaysTemplate), persistent: true, as: .PNG), voiceCommands: [ACSpeakAppNameMenuName], handler: { _ in
             manager.send(request: SDLSpeak(tts: ExampleAppNameTTS), responseHandler: { (_, response, error) in
                 guard response?.resultCode == .success else { return }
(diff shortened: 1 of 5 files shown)


Example Apps/Example Swift/MenuManager.swift
```diff
@@ -11,6 +11,9 @@ import SmartDeviceLink
 import SmartDeviceLinkSwift
 
 class MenuManager: NSObject {
+
+    static var currentTemplate : SDLPredefinedLayout = .nonMedia
```
NicoleYarroch, line 15: I wonder if it might be better to pass a `refreshTemplateHandler` that returns the selected template to the `MenuManager` similar to how we pass a `refreshUIHandler` to the `ButtonManager`. That way we can remove setting the template in the menu manager and consolidate the code for setting the template in the `ProxyManager`.

Then the handler is triggered when the `currentTemplate` is set:
```
public fileprivate(set) var currentTemplate: SDLPredefinedLayout {
    didSet {
        guard let refreshTemplateHandler = refreshTemplateHandler else { return }
        refreshTemplateHandler(currentTemplate)
    }
}
```

NicoleYarroch (changes requested):
Left an idea for streamlining the logic flow of how the template gets set on the Swift ex. app. Let me know if you think it works or have questions

Example Apps/Example Swift/MenuManager.swift
```diff
@@ -115,32 +132,21 @@ private extension MenuManager {
...
         let submenuTitleNonMedia = "Non - Media (Default)"
-        submenuItems.append(SDLMenuCell(title: submenuTitleNonMedia, icon: nil, voiceCommands: nil, handler: { (triggerSource) in
-            manager.screenManager.changeLayout(SDLTemplateConfiguration(predefinedLayout: .nonMedia)) { err in
-                if err != nil {
-                    AlertManager.sendAlert(textField1: errorMessage, sdlManager: manager)
-                    return
-                }
-            }
+        submenuItems.append(SDLMenuCell(title: submenuTitleNonMedia, icon: nil, voiceCommands: nil, handler: { [weak self] (triggerSource) in
+            self?.currentTemplate = .nonMedia
```
NicoleYarroch, line 143: I don't think we need to save the current template on both the menu manager and the proxy manager. If you want to keep the current template state on the proxy manager, then I would remove the `currentTemplate` variable and just call the `refreshTemplateHandler` handler here.

Example Apps/Example Swift/MenuManager.swift
```diff
@@ -10,12 +10,29 @@ import Foundation
...
+
+    /// Setting current selected template
+    public fileprivate(set) var currentTemplate: SDLPredefinedLayout {
+        didSet {
+            guard let refreshTemplateHandler = refreshTemplateHandler else { return }
+            refreshTemplateHandler(currentTemplate)
+        }
+    }
+
+    init(updateTemplate: RefreshTemplateHandler? = nil) {
```
NicoleYarroch, line 26:
```suggestion
    init(sdlManager: SDLManager, refreshTemplateHandler: RefreshTemplateHandler? = nil, choiceSetManager: PerformInteractionManager) {
```

NicoleYarroch: If the `sdlManager` is passed in the `init`, we can get rid of all the passed `sdlManager`s for each menu item.

Example Apps/Example Swift/MenuManager.swift
```diff
@@ -10,12 +10,29 @@ import Foundation
...
+        self.refreshTemplateHandler = updateTemplate
+        currentTemplate = .nonMedia
+    }
+
     /// Creates and returns the menu items
     ///
     /// - Parameter manager: The SDL Manager
     /// - Returns: An array of SDLAddCommand objects
-    class func allMenuItems(with manager: SDLManager, choiceSetManager: PerformInteractionManager) -> [SDLMenuCell] {
+    func allMenuItems(with manager: SDLManager, choiceSetManager: PerformInteractionManager) -> [SDLMenuCell] {
```
NicoleYarroch, line 35:
```suggestion
    var allVoiceMenuItems: [SDLVoiceCommand] {
```

NicoleYarroch: Since we can remove the parameters after passing the `sdlManager` and the `choiceSetManager` in the `init`, I think it makes more sense for each of the menu item functions to be refactored into variables.

Example Apps/Example Swift/ProxyManager.swift
```diff
@@ -248,11 +252,25 @@ private extension ProxyManager {
...
+            }
+        }
+    }
+
     /// Set the template and create the UI
     func showInitialData() {
         guard sdlManager.hmiLevel == .full else { return }
 
-        sdlManager.screenManager.changeLayout(SDLTemplateConfiguration(predefinedLayout: .nonMedia), withCompletionHandler: nil)
+        sdlManager.screenManager.changeLayout(SDLTemplateConfiguration(predefinedLayout: currentTemplate), withCompletionHandler: nil)
```
NicoleYarroch, line 273:
```suggestion
        refreshTemplateHandler(currentTemplate)
```

Example Apps/Example Swift/ProxyManager.swift
```diff
@@ -248,11 +252,25 @@ private extension ProxyManager {
         }
     }
 
+    /// Handler for refreshing the template
+    var refreshTemplateHandler: RefreshTemplateHandler {
+        return { [weak self] template in
+            let errorMessage = "Changing the template failed"
```
NicoleYarroch, line 258:
```suggestion
            guard let self = self else { return }
            let errorMessage = "Changing the template failed"
```

NicoleYarroch: You can remove all the `?` after `self` once you add this check

NicoleYarroch (changes requested):
I left some suggestions on the Swift Example app

joeljfischer:
@NicoleYarroch @FrankElias77 , I think this should be like 5 lines of changes, not 500. Why are you doing all of this unnecessary work? All you have to do is ensure that the first non_media is set the one time it goes from NONE->FULL. Can you explain why that tiny change isn't sufficient?

NicoleYarroch:
I prefer the logic of keeping the `ProxyManager` in charge of all screen and template updates but I guess we could add back the flag to keep track of the first time the `hmiLevel` goes to full and use it to set the template once and then have the menu manager set the template as needed.

joeljfischer:
When I start to look at the hundreds of lines of changes, callbacks, etc that are added here it starts to seem over-engineered to me. To fix this bug only needs a few lines of code to ensure that it only happens the first time it reaches HMI FULL. This is similar to what I did for #1818 

@FrankElias I would recommend starting over with a new branch and PR. This issue should only take 20 minutes.

FrankElias77:
Closed in favor of PR #1854

PR #1833, opened 2020-10-30
Put SDL queues at the highest priority

### LINKED ISSUE #1778: Softbutton Delay
### Bug Report 
With the latest sdk (6.7.0), softbuttons are not updating as quickly. I believe it may be due to the addition of the softbutton manager.

##### Reproduction Steps
1.  Load app and update softbutton text every couple seconds using the screenmanager

##### Expected Behavior
Softbuttons should update on SYNC with the new text

##### Observed Behavior
Softbuttons are not updating as fast. ( issue was not seen sdk 6.5.0)

* iOS Version: 13.5
* SDL iOS Version: 6.7.0
* Testing Against: App Partner beta app
* SYNC 3

joeljfischer:
Fixes #1778

### Risk

### Testing Plan

#### Unit Tests
No unit tests were needed

#### Core Tests
Verified lots of screen manager behavior with SDL Core. Slight improvements are possible, but it's not super easy to tell, especially if testing against Sync.

Core version / branch / commit hash / module tested against: Sync Gen 3.4
HMI name / version / branch / commit hash / module tested against: Sync Gen 3.4

### Summary
This update moves SDL queues to be at the highest priority level instead of the second-highest priority level, and additionally changes the soft button operations to always send `Show.mainField1` even if it should be blank.

##### Bug Fixes
* SDL code now runs at the highest priority level. Note that this may delay calls to the main thread, which may impact UI performance. However, because SDL code only runs when an SDL app is connected, this should be okay and should improve SDL performance and lag.
* SDL soft button operations now always send `Show.mainField1`, which works around a bug on Ford Sync head units.

### Tasks Remaining:
- [x] Additional testing with video streaming apps
- [x] See if any metrics can be generated

### CLA

### DIFF
--- a/SmartDeviceLink/private/SDLChoiceSetManager.m
+++ b/SmartDeviceLink/private/SDLChoiceSetManager.m
@@ -147,6 +147,7 @@ UInt16 const ChoiceCellCancelIdMin = 1;
     NSOperationQueue *queue = [[NSOperationQueue alloc] init];
     queue.name = @"com.sdl.screenManager.choiceSetManager.transactionQueue";
     queue.maxConcurrentOperationCount = 1;
+    queue.qualityOfService = NSQualityOfServiceUserInteractive;
     queue.underlyingQueue = [SDLGlobals sharedGlobals].sdlConcurrentQueue;
     queue.suspended = YES;
 
--- a/SmartDeviceLink/private/SDLGlobals.m
+++ b/SmartDeviceLink/private/SDLGlobals.m
@@ -60,8 +60,8 @@ typedef NSNumber *MTUBox;
     _rpcVersion = [[SDLVersion alloc] initWithString:@"1.0.0"];
     _dynamicMTUDict = [NSMutableDictionary dictionary];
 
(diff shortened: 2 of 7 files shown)


NicoleYarroch:
This PR did not break anything, however, the responsiveness of the soft buttons on SYNC 3.0 did not improve for me. 

I noticed in the `SDLSoftButtonManager` header that the `currentMainField1` has this documentation **HAX: This is necessary due to a Ford Sync 3 bug that doesn't like Show requests without a main field being set (it will accept them, but with a GENERIC_ERROR, and 10-15 seconds late...)**. In the library's Swift and Obj-C example apps, if the `Show`'s `mainField1` is cleared I get the `GENERIC_ERROR` 10 second delay because `currentMainField1` is set to `nil` in the `SDLSoftButtonManager`. I tried intercepting the `nil` values for `currentMainField1` and set an empty string `@""`, and I no longer experienced soft button update delays. Perhaps this is the issue the bug reporter was experiencing?

NicoleYarroch (changes requested):
I did some investigating and I think I might have found the issue...

NicoleYarroch (approved):

NicoleYarroch:
@rbright55 Can you please try this PR to see if it resolves your issue (#1778)? Thanks!

PR #1848, opened 2020-11-10
Bugfix 1847 library stops working if startservice times out

### LINKED ISSUE #1847: Library stops working if StartSession(RPC) times out
### Bug Report
In some cases the head unit takes more time to respond to StartService for RPC request and the library hits the timeout. See https://github.com/smartdevicelink/sdl_ios/blob/7.0.0/SmartDeviceLink/private/SDLLifecycleProtocolHandler.m#L86

This will cause the Protocol to close and end the transport. 
https://github.com/smartdevicelink/sdl_ios/blob/7.0.0/SmartDeviceLink/private/SDLProtocol.m#L89
https://github.com/smartdevicelink/sdl_ios/blob/7.0.0/SmartDeviceLink/private/SDLIAPTransport.m#L218

In the transport class following code will be executed:

```objc
- (void)sdl_stopEventListening {
    SDLLogV(@"SDLIAPTransport stopped listening for events");
    [[NSNotificationCenter defaultCenter] removeObserver:self];
}
```

Because the disconnect/stop was called in lower levels but not in the manager, there is no restart method for the transport object. No component of sdl_ios is now listening to EAConnect notifications. As a result the app will never connect to any head unit until it's killed and restarted on the phone.

##### Reproduction Steps
1. Install SDL Example on an iPhone
2. Connect iPhone to a head unit
3. Wait for the StartService timeout to occur
4. Disconnect iPhone
5. Reconnect iPhone

##### Expected Behavior
The app should try to connect again.

##### Observed Behavior
The app removed observer to EAConnect. It will never be able to connect to the head unit.

* iOS Version: Any
* SDL iOS Version: 7.0.0
* Testing Against: SDL Example

The issue is more visible on SYNC3.0 if >20 apps are installed. It's possible to stop SDL Core on the SYNC3.0 to make it easier to reproduce/test. Details can be shared separately.

kshala-ford:
Fixes #1847 

### Risk

### Testing Plan

### Summary
Changes the moment when a transport object removes observers to EA(dis)connect.

### Tasks Remaining:
- [ ] [Perform unit tests]
- [ ] [Test with and without secondary transport]
- [ ] [Verify the transport is deallocated properly when a new one is created]

### CLA

### DIFF
--- a/SmartDeviceLink/private/SDLIAPTransport.m
+++ b/SmartDeviceLink/private/SDLIAPTransport.m
@@ -57,6 +57,10 @@ int const CreateSessionRetries = 3;
     return self;
 }
 
+- (void)dealloc {
+    [self sdl_stopEventListening];
+}
+
 #pragma mark - Notifications
 
 /**
@@ -217,8 +221,6 @@ int const CreateSessionRetries = 3;
  */
 - (void)disconnectWithCompletionHandler:(void (^)(void))disconnectCompletionHandler {
     SDLLogD(@"Disconnecting IAP transport");
-    // Stop event listening here so that even if the transport is disconnected by `SDLProxy` when there is a start session timeout, the class unregisters for accessory notifications
-    [self sdl_stopEventListening];
 
     self.retryCounter = 0;
     self.sessionSetupInProgress = NO;


joeljfischer:
Closed in favor of #1951

PR #1869, opened 2020-12-07
Update to only overwrite non-persistent files on RPC v4.3.1 or less

### LINKED ISSUE #827: Work around module ListFiles bug
### Bug Report
On previous version of Core, there is a bug that causes `ListFiles` to include previously uploaded files that no longer exist on the system. https://github.com/smartdevicelink/sdl_core/issues/1393

We should work around this bug in the `SDLFileManager`. Likely we need a local array of uploaded files. If an ephemeral file is not in the local list but is in the remote list, the `overwrite` flag should be automatically toggled and the file uploaded.

##### Reproduction Steps
1. Connect to an affected version of SDLCore
2. Upload several ephemeral files
3. Disconnect and power cycle SDLCore
4. Reconnect
5. Send a ListFiles RPC
6. Upload the same ephemeral files

##### Expected Behavior
The previously uploaded files are not listed, new uploads succeed.

##### Observed Behavior
The previously uploaded files are listed, new uploads fail because they "already exist" when they do not.

* iOS Version: n/a
* SDL iOS Version: n/a
* Testing Against: Ford TDK, Core v4.3 and below

joeljfischer:
Fixes #827 

### Risk

### Testing Plan

#### Unit Tests
Unit tests were run, but no new tests were added.

#### Core Tests
Tested displaying a non-persistent artwork on a v4.2 head unit and a v6.0 head unit

Core version / branch / commit hash / module tested against: Manticore (6.0), Sync 3.4 (4.2)
HMI name / version / branch / commit hash / module tested against: Manticore (Generic HMI 0.8.2), Sync 3.4

### Summary
This PR alters a check that was added in SDL iOS v5.2 which changes non-persistent files to always overwrite existing data. This check was added due to a bug in Core v4.3 and below that doesn't properly check the cache for a list files, causing these files to not update properly unless they attempt an overwrite. Because this bug was fixed in SDL Core v4.3.1, we can add a version check to only apply the check on RPC versions v4.3.0 and below (which corresponds to Core v4.3.1 and below).

##### Bug Fixes
* Don't auto-overwrite non-persistent files when not necessary on newer SDL Core versions.

### Tasks Remaining:
- [x] Test against newer Core versions to ensure that no longer overwriting non-persistent files works properly.

### CLA

### DIFF
--- a/SmartDeviceLink/private/SDLLifecycleManager.m
+++ b/SmartDeviceLink/private/SDLLifecycleManager.m
@@ -395,7 +395,7 @@ NSString *const BackgroundTaskTransportName = @"com.sdl.transport.backgroundTask
 
         weakSelf.registerResponse = (SDLRegisterAppInterfaceResponse *)response;
         [SDLGlobals sharedGlobals].rpcVersion = [SDLVersion versionWithSDLMsgVersion:weakSelf.registerResponse.sdlMsgVersion];
-        SDLLogD(@"Did register app; RPC version: %@, SDL version: %@, starting languages: (VR) %@, (HMI) %@, vehicle type: %@", weakSelf.registerResponse.sdlMsgVersion, weakSelf.registerResponse.sdlVersion, weakSelf.registerResponse.language, weakSelf.registerResponse.hmiDisplayLanguage, weakSelf.registerResponse.vehicleType);
+        SDLLogD(@"Did register app; RPC version: %@, SDL version: %@, starting languages: (VR) %@, (HMI) %@, vehicle type: %@", weakSelf.registerResponse.sdlMsgVersion, (weakSelf.registerResponse.sdlVersion ?: @"unavailable"), weakSelf.registerResponse.language, weakSelf.registerResponse.hmiDisplayLanguage, weakSelf.registerResponse.vehicleType);
 
         [weakSelf sdl_transitionToState:SDLLifecycleStateRegistered];
     }];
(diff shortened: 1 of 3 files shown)


SmartDeviceLink/private/SDLLifecycleManager.m
```diff
@@ -395,6 +395,8 @@ - (void)didEnterStateConnected {
 
         weakSelf.registerResponse = (SDLRegisterAppInterfaceResponse *)response;
         [SDLGlobals sharedGlobals].rpcVersion = [SDLVersion versionWithSDLMsgVersion:weakSelf.registerResponse.sdlMsgVersion];
+        SDLLogD(@"Did register app; RPC version: %@, SDL version: %@, starting languages: (VR) %@, (HMI) %@, vehicle type: %@", weakSelf.registerResponse.sdlMsgVersion, weakSelf.registerResponse.sdlVersion, weakSelf.registerResponse.language, weakSelf.registerResponse.hmiDisplayLanguage, weakSelf.registerResponse.vehicleType);
```
NicoleYarroch, line 398: When connecting to SYNC 3 I get **Did register app; RPC version: 4.2.(null)...**. Is it possible not to print the patch version if it is `null`?

NicoleYarroch (changes requested):

SmartDeviceLink/private/SDLLifecycleManager.m
joeljfischer: Fixed

NicoleYarroch (approved):

PR #1906, opened 2021-01-29
Issue 799 prevent incorrect background processing

kmicha19-ford:
Fixes #1799 (when used in conjunction with PR #https://github.com/kmicha19-ford/sdl_ios/pull/1)

This PR is ready for review.

### Risk
This PR makes no API changes.

### Testing Plan
-  I have verified that I have not introduced new warnings in this PR (or explain why below)
- I have tested this PR against Core and verified behavior 

#### Unit Tests

No changes to any unit tests are required.

#### Core Tests

- Built the Obj-C example app and verified that all interactions work
   - iAP disconnect / reconnect
- Built the Swift example app and verified that all interactions work
   - iAP disconnect / reconnect
   
### Summary

The iOS External Åccessory framework manages app background execution on connect and disconnect.  The LifeCycle

##### Breaking Changes
* [Breaking change info]

##### Enhancements
* [Enhancement info]

##### Bug Fixes
* [Bug Fix Info]

### Tasks Remaining:
- [ ] [Task 1]
- [ ] [Task 2]

### CLA
- [ ] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

PR #1909, opened 2021-02-03
Prevent incorrect background processing

### LINKED ISSUE #1799: High CPU usage in Energy Organizer, iOS 14+

kmicha19-ford:
Fixes #1799  (when used in conjunction with EASession left open PR #https://github.com/kmicha19-ford/sdl_ios/pull/1)

This PR is ready for review.

### Risk

This PR makes no  API changes.

### Testing Plan

- I have verified that I have not introduced new warnings in this PR (or explain why below)
- I have tested this PR against Core and verified behavior.

#### Unit Tests

- Does not affect any unit tests.

#### Core Tests

- Built the Obj-C example app and verified that all interactions work
   - iAP disconnect / reconnect
   - LockScreen on/off rapidly
- Built the Swift example app and verified that all interactions work
   - iAP disconnect / reconnect
   - LockScreen on/off rapidly

### Summary

This fixes the bug where an app could run on and on unnecessarily even when backgrounded and disconnected. 

##### Breaking Changes
* None

##### Enhancements

* SDLifeCycleManager no longer uses the SDLBackgroundTaskManager.
* SDLBackgroundTaskManager uses the iOS instance method beginBackgroundTask(withName taskName: String?, 
    expirationHandler handler: (() -> Void)? = nil) -> UIBackgroundTaskIdentifier
* beginBackgroundTask is for requesting additional processing time when an app is backgrounded and before being suspended so an app can finish tasks, such as saving data, before being suspended.  This method is not meant to keep an app running in the background.  After 30 seconds iOS may or may not terminate the app
* Using beginBackgroundTask in this manner may conflict with EAAccessory Connect / Disconnect app management and keep the app running in the background when it should be suspended. This may cause undo battery drain. 
* It should be noted that SDLSecondaryTransportManager correctly uses SDLBackgroundTaskManager.

### Tasks Remaining:
- None

### CLA
- I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

### DIFF
--- a/SmartDeviceLink/private/SDLLifecycleManager.m
+++ b/SmartDeviceLink/private/SDLLifecycleManager.m
@@ -14,7 +14,6 @@
 #import "SDLLifecycleRPCAdapter.h"
 #import "SDLAsynchronousRPCOperation.h"
 #import "SDLAsynchronousRPCRequestOperation.h"
-#import "SDLBackgroundTaskManager.h"
 #import "SDLChangeRegistration.h"
 #import "SDLConfiguration.h"
 #import "SDLConnectionManagerType.h"
@@ -107,7 +106,6 @@ NSString *const BackgroundTaskTransportName = @"com.sdl.transport.backgroundTask
 @property (copy, nonatomic) SDLManagerReadyBlock readyHandler;
 @property (copy, nonatomic) dispatch_queue_t lifecycleQueue;
 @property (assign, nonatomic) int32_t lastCorrelationId;
-@property (copy, nonatomic) SDLBackgroundTaskManager *backgroundTaskManager;
 @property (strong, nonatomic) SDLLanguage currentVRLanguage;
 
 // RPC Handlers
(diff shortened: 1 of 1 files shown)


joeljfischer:
@makinney I apologize for not noticing earlier, but is this a duplicate of #1906? Is an additional PR intended to be submitted sometime soon as well?

kmicha19-ford:
will recreate

kmicha19-ford:
Hi Joel,

These are new as of today.

https://github.com/smartdevicelink/sdl_ios/pull/1910
https://github.com/smartdevicelink/sdl_ios/pull/1911

Hopefully these are correct.  Let me know otherwise. Sorry for any delays.  I never done any PRs on public repos before (can you tell?)

Mike
________________________________

PR #1912, opened 2021-02-04
Fix choice set default timeout behavior

### LINKED ISSUE #1891: Fix SDLChoiceSet default timeout
### Bug Report
Currently, it's possible for `SDLChoiceSet` default timeouts to be used incorrectly.

##### Reproduction Steps
1. Create a `SDLChoiceSet` without the timeout set
2. Update the default timeout of `SDLChoiceSet`

##### Expected Behavior
The choice set created before the update uses the new default choice set time

##### Observed Behavior
The choice set uses the old default choice set time.

* iOS Version: n/a
* SDL iOS Version: 7.1.0-develop
* Testing Against: n/a

n/a

NicoleYarroch:
Fixes #1891

### Risk

### Testing Plan

#### Unit Tests
Test cases for setting and getting the `SDLChoiceSet`'s `defaultTimeout` and `timeout` properties were added to the **SDLChoiceSetSpec**.

#### Core Tests
Tested with Manticore.

Core version / branch / commit hash / module tested against: Manticore (SDL Core v7.0.0)
HMI name / version / branch / commit hash / module tested against: Manticore (Generic HMI v0.9.0)

### Summary
Fixed the behavior of the`SDLChoiceSet`'s `defaultTimeout` and `timeout` properties:
1. If the `timeout` value is set to `0`, getting the `timeout` value now returns the `defaultTimeout` value instead of `0`.
2. The `timeout` value is now capped to a 5-100 second range. If set to a value less than 5, it will be capped at 5. If set to a value greater than 100, it will be capped to 100. 
1. The `timeout` value can now be set to any value in the `SDLChoiceSet` inits. Previously the object would not be created if the `timeout` was not set to a value in the 5-100 range. 
1. If the `SDLChoiceSet.defaultTimeout` is changed after the `SDLChoiceSet` object is created without the `timeout` value set, the new `SDLChoiceSet.defaultTimeout` is used to set the `timeout` value instead of the old `SDLChoiceSet.defaultTimeout ` value. 

##### Bug Fixes
* Fixed issues with getting the `SDLChoiceSet`'s `defaultTimeout` and `timeout` properties.

### CLA


SmartDeviceLink/public/SDLChoiceSet.m
```diff
@@ -57,7 +58,9 @@ - (instancetype)initWithTitle:(NSString *)title delegate:(id<SDLChoiceSetDelegat
         return nil;
     }
 
-    if (timeout < 5 || timeout > 100) {
+    if (timeout == 0) {
+        SDLLogV(@"Creating a choice set with a 0 second timeout; the default timeout will be used: %f seconds", SDLChoiceSet.defaultTimeout);
```
joeljfischer, line 62:
```suggestion
        SDLLogV(@"Creating a choice set with a 0 second timeout; the default timeout will be used: %.01f seconds", SDLChoiceSet.defaultTimeout);
```

This probably needs to reduce the number of decimals.

SmartDeviceLink/public/SDLChoiceSet.m
```diff
@@ -57,7 +58,9 @@ - (instancetype)initWithTitle:(NSString *)title delegate:(id<SDLChoiceSetDelegat
...
+    } else if (timeout < 5 || timeout > 100) {
```
joeljfischer, line 63: This isn't capped this way in `AlertView`. We want the same system across both.

SmartDeviceLink/public/SDLChoiceSet.m
```diff
@@ -144,6 +147,14 @@ - (void)setHelpList:(nullable NSArray<SDLVRHelpItem *> *)helpList {
     }
 }
 
+- (NSTimeInterval)timeout {
+    if (_timeout == TimeoutDefault) {
+        return SDLChoiceSet.defaultTimeout;
+    }
+
+    return _timeout;
+}
```
joeljfischer, line 156: This doesn't look the same as `SDLAlertView` (neither does `+(NSTimeInterval)defaultTimeout`

joeljfischer (changes requested):

SmartDeviceLink/public/SDLChoiceSet.m
NicoleYarroch: This log was removed

SmartDeviceLink/public/SDLChoiceSet.m
NicoleYarroch: Updated to cap the timeout range between 5 and 100 seconds.

SmartDeviceLink/public/SDLChoiceSet.m
NicoleYarroch: Updated to cap the timeout range between 5 and 100 seconds.

joeljfischer (approved):
This will need to be aligned before merge

PR #1935, opened 2021-03-04
Fix size of bulk data in PutFile

### LINKED ISSUE #1927: Fix the size of the bulk data that can be set in a PutFile
In the `SDLUploadFileOperation` class, if the file data exceeds the max size allowed for a packet, the file data is broken into chunks and sent in separate `SDLPutFile`s. However, there is a bug in the algorithm that breaks the file data into chunks because it assumes that the max allowed data chunk size is the max `mtu` size for the RPC service. This is wrong because the max `mtu` size is for the entire packet, which also includes the frame header, binary header, and JSON data along with the bulk data. 

Since the file data is broken into chunks equal to the max `mtu` size, each of the messages generated for the `SDLPutFile`s will exceed the allowed `mtu` size. This means that in the protocol layer, the messages will be sent as multi-frame packets instead of as one single-frame packet. This defeats the purpose of breaking the file data into chunks and sending it in multiple `SDLPutFile`s.  

##### Reproduction Steps
1. Use the file manager to upload a very large file that must be broken into separate `SDLPutFile`s.

##### Expected Behavior
All of the packets created for the `SDLPutFiles` generated by the `SDLUploadFileOperation` should be less than or equal to the `mtu` size allowed by the RPC service.  

##### Observed Behavior
All of the packets (except the last packet) created for the `SDLPutFiles` generated by the `SDLUploadFileOperation` are greater than the `mtu` size allowed by the RPC service.  

* iOS Version: n/a
* SDL iOS Version: 7.1
* Testing Against: n/a

NicoleYarroch:
Fixes #1927

### Risk

### Testing Plan

#### Unit Tests
Unit tests fixed in the **SDLUploadFileOperationSpec**.

#### Core Tests
* Sent a large image file to Manticore emulator and SYNC 3.0 and made sure each generated `SDLPutFile` was sent as a single frame message. 

##### Core version / branch / commit hash / module tested against:
* Manticore: SDL Core v7.0.0
* SYNC 3.0 (BUILD 17276_DEVTEST)
 
##### HMI name / version / branch / commit hash / module tested against:
* Manticore: Generic HMI v0.9.0
* SYNC 3.0 (BUILD 17276_DEVTEST) HMI
    
### Summary
Fixed the size of the bulk data set in `SDLPutFile` so the message is not broken into multiple frames if the size of the total message exceeds the max `mtu` size allowed for RPC connection. 

##### Bug Fixes
* Fixed the max size of the bulk data set in `SDLPutFile`.

### Tasks Remaining:
- [x] Test with SYNC 3.0
- [x] Test with Manticore

### CLA


joeljfischer (changes requested):
A few small comments

joeljfischer (approved):

Issue #1971, opened 2021-04-15
Encrypted RPCs that fail because encryption isn't set up don't call back


joeljfischer:
### Bug Report
If RPC service encryption is set up, attempting to send an RPC request before encryption is set up will result in the handler never being called. This can cause managers to hang when setting up if, for example, `PutFile` is encrypted. This results in the app being totally inoperable.

##### Reproduction Steps
1. Set up RPC encryption on Core and make `PutFile` require encryption
2. Connect an app that supports RPC encryption

##### Expected Behavior
The app successfully sets up, or at least is operable after failing some part

##### Observed Behavior
The app is entirely broken due to `SDLFileManager` hanging when attempting to send the app icon. This fails due to RPC encryption not being set up yet and the handler is never called, leaving the operation and manager to hang.

##### OS & Version Information
* iOS Version: n/a
* SDL iOS Version: v7.1.0
* Testing Against: Core v7.1.0 / Generic HMI v0.10.0

##### Test Case, Sample Code, and / or Example App
n/a

PR #1972, opened 2021-04-15
Fix a crash when RPC encryption starts and callback when RPCs fail due to RPC encryption

### LINKED ISSUE #1966: Post-Connection RPC StartService w/ Encryption Causes a Crash
### Bug Report
When connecting an encrypted connection, there's a persistent crash.

##### Reproduction Steps
1. Create an encrypted connection to Core
2. Connect

##### Expected Behavior
No crash

##### Observed Behavior
Crash in Lifecycle Manager when the second RPC StartServiceACK is received.

* iOS Version: n/a
* SDL iOS Version: 7.1-develop
* Testing Against: Core v7.1-develop

joeljfischer:
Fixes #1966, #1971

### Risk

### Testing Plan

#### Unit Tests
Unit tests were updated to ensure that the callback is handled correctly.

#### Core Tests
Tested an RPC encrypted app with `PutFile` encrypted. Ensured that no crashes occurred and the app set up correctly.

Core version / branch / commit hash / module tested against: v7.1.0-RC.1 (0df66080512cbe76ee5d4872910a54adf3ca1f36)
HMI name / version / branch / commit hash / module tested against: sdl_hmi v5.5.0

### Summary
1. `SDLLifecycleManager` now ignores RPC `StartServiceACK` if we're already connected instead of crashing.
1. Updates `SDLProtocol sendRPC:` to pass back an error and return a success BOOL so that the handler can be called in the `SDLLifecycleManager` if the RPC fails to send in the protocol layer.

##### Bug Fixes
* Fixed RPC encryption causing a crash when the second RPC `StartServiceACK` is received.
* Fixed not calling an RPC request handler if the message failed to send in the protocol layer of the library.

### Tasks Remaining:
- [x] Add a unit test for the RPC failing to send in the protocol layer should call the handler.
- [x] Smoke test on Core v7.1.0
- [x] Smoke test on Sync 3.4

### CLA


SmartDeviceLink/private/SDLProtocol.h
```diff
@@ -136,8 +136,9 @@ extern NSString *const SDLProtocolSecurityErrorDomain;
  *  Sends an unencrypted RPC to Core
  *
  *  @param message A SDLRPCMessage message
+ *  @returns YES if the RPC was sent, NO if it was not
```
NicoleYarroch, line 139: Can you add docs for the `error` param?

NicoleYarroch (changes requested):
I small documentation fix requested.

NicoleYarroch (approved):

PR #2019, opened 2021-07-06
Refactor Choice Set Manager to Fix Several Bugs

### LINKED ISSUE #2018: Back-to-Back Choice Sets Can Fail
### Bug Report
I'm seeing a bug in the choice set manager where back-to-back choice sets can fail. Because the `SDLPresentChoiceSetOperation` uses the standard `completionBlock`, which runs after the operation completes and the next one starts, if two present operations are next to each other, the second should run, but I've seen cases where that does not happen. This is because the `completionBlock` clears the `pendingPresentationSet`, which may no longer be the running presentation set.

Instead, we should do as we do in the alert manager and add pending presentations to the queue and cancel previous non-executing ones if a new one is added.

##### Reproduction Steps
1. Create a choice set
2. Before an option is pressed, present a new choice set

##### Expected Behavior
The second choice set appears

##### Observed Behavior
The second choice set fails to appear because the choice set is `nil` in the `PresentChoiceSetOperation`. This is because it's set to `nil` in `sdl_presentChoiceSetWithMode` in the `completionBlock`.

* iOS Version: iOS 14.6
* SDL iOS Version: v7.2.0
* Testing Against: Manticore and Sync 3.4

n/a

joeljfischer:
Fixes #2018, #2020, #2025

### Risk

### Testing Plan

#### Unit Tests
Unit tests updated and rewritten

#### Core Tests
* Tested sending a presentation, then sending another presentation when the first has a button pressed.
* Tested sending a presentation on various systems in both swift and obj-c

Core version / branch / commit hash / module tested against: Manticore (SDL v7.1.1), Sync Gen 3.4
HMI name / version / branch / commit hash / module tested against: Manticore (Generic HMI v0.10.0), Sync Gen 3.4

### Summary
1. This PR rewrites the choice set manager to use a full queuing system rather than the partial queueing system of the past. 
2. This PR rewrites the preload / present operations into one operation to solve some queueing problems. The operation will now figure out what it has to do based on how it is initialized.
3. This PR changes cancelation behavior. Before if you sent a present, then another present, the manager would try to cancel previous ones. Now they will be queued and the dev can cancel previous ones if desired.
4. This PR fixes strange deletion behavior. Before if a delete came in while a preload / present was happening, the manager would try to delete choices from the upload. Now it will wait until the delete operation comes around and delete at that time. This allows proper behavior around preload -> present -> delete situations.

##### Bug Fixes
* Fixed unpredictable choice set behavior when quickly loading / presenting / deleting choices. They now will happen in the order that you call the methods.

### Tasks Remaining:
Done

### CLA


FrankElias77 (commented):
1. Preload choice cells
2. Present choice cells
3. Select any cell
After step num 3, selected cells do not trigger `didSelectChoice` delegate method of `SDLChoiceSetDelegate`

In addition, I was unable to find the voice commands associated with the choice cells after presenting choice set

joeljfischer:
@FrankElias77 The first issue is fixed. The second is definitely being sent, but it may not function on certain head units?

FrankElias77 (dismissed):

SmartDeviceLink/private/SDLChoiceSetManager.m
```diff
@@ -204,164 +188,78 @@ - (void)didEnterStateStartupError {
...
 
+    [self.transactionQueue addOperation:deleteOp];
+}
+
+#pragma mark Upload / Present
+
+- (void)preloadChoices:(NSArray<SDLChoiceCell *> *)choices withCompletionHandler:(nullable SDLPreloadChoiceCompletionHandler)handler {
+    SDLLogV(@"Request to preload choices: %@", choices);
+    if (choices.count == 0) { return; }
+    if ([self.currentState isEqualToString:SDLChoiceManagerStateShutdown]) {
```
joeljfischer, line 226: Change to `!Ready` instead

SmartDeviceLink/private/SDLPreloadPresentChoicesOperation.m
```diff
@@ -0,0 +1,616 @@
...
+    }
+
+    // Remove cells that are already loaded, add ids to cells needing upload, then make the cells to upload unique
+    [self.cellsToUpload minusSet:self.loadedCells];
+    [SDLPreloadPresentChoicesOperationUtilities assignIdsToCells:self.cellsToUpload loadedCells:self.loadedCells];
+    [SDLPreloadPresentChoicesOperationUtilities makeCellsToUploadUnique:self.cellsToUpload basedOnLoadedCells:self.mutableLoadedCells windowCapability:self.windowCapability];
+
+    if ((self.loadedCells.count == UINT16_MAX) && (self.cellsToUpload.count > 0)) {
+        return [self finishOperation:[NSError sdl_choiceSetManager_noIdsAvailable]];
+    }
```
joeljfischer, line 178: Move before L173

FrankElias77 (approved):

FrankElias77 (dismissed):

SmartDeviceLink/private/SDLPreloadPresentChoicesOperation.m
```diff
@@ -290,6 +281,26 @@ - (void)sdl_uploadCellsWithCompletionHandler:(void(^)(NSError *_Nullable error))
     }];
 }
 
+#pragma mark - Artwork
+
+/// Get an array of artwork that needs to be uploaded form a list of menu cells
+/// @param cells The menu cells to get artwork from
```
FrankElias77, line 287: Please add missing descriptions for the remaining params

FrankElias77 (changes requested):

FrankElias77 (dismissed):

PR #2040, opened 2021-09-16
Fix file operations check file system at queue time instead of run time

### LINKED ISSUE #2034: File Manager will upload the same file multiple times
### Bug Report
When trying to load multiple of the same artwork and the file is not uploaded yet, the file will be uploaded multiple times. This is because we create multiple of the same upload operation and they all execute.

We should probably implement the upload / delete operation to check if the file needs to be uploaded at the time it would run instead of the manager figuring out if the operation should be created at all.

##### Reproduction Steps
1. Create 99 `SDLArtwork`s that are exactly the same
2. Attempt to upload all 99 in a loop

##### Expected Behavior
The one artwork will be uploaded 99 times, slowing down the head unit like crazy.

##### Observed Behavior
The one artwork should only be uploaded 1 time.

* iOS Version: [What version of iOS are you using when the bug occurred]
* SDL iOS Version: [What version of the library has this bug been seen on]
* Testing Against: [What you tested with to observe this behavior]

[Paste a link to a PR, gist, or other code that exemplifies this behavior]

joeljfischer:
Fix #2034

### Risk

### Testing Plan

#### Unit Tests
Unit tests added to handle the change to operations checking remote files

#### Core Tests
* Tested existing example app code
* Tested creating multiple of the same file in a loop doesn't upload the file multiple times

Core version / branch / commit hash / module tested against: Manticore (Core v7.1.1)
HMI name / version / branch / commit hash / module tested against: Manticore (Generic HMI v0.10.0)

### Summary
This PR updates the File Manager to check if a file is already on the system when an operation starts instead of when an operation is queued. This fixes issues if there is a change between when a file upload or delete is queued rather than when it's run.

##### Bug Fixes
* Fixed issues where file operations could fail if file state changes between when the operation is queued and when it runs. For example, attempting to upload 99 of the same file will cause 99 of that file to actually be uploaded if the file is not on the system when those 99 operations are queued. Now the system will only upload the file once.

### Tasks Remaining:
n/a

### CLA


FrankElias77 (commented):
Additional changes needed to support older SYNC versions

FrankElias77 (approved):

SmartDeviceLink/public/SDLFileManager.m
```diff
@@ -262,18 +258,21 @@ - (void)deleteRemoteFilesWithNames:(NSArray<SDLFileName *> *)names completionHan
...
     dispatch_group_enter(deleteFilesTask);
-    for(NSString *name in names) {
+    __weak typeof(self) weakself = self;
+    for (NSString *name in names) {
         dispatch_group_enter(deleteFilesTask);
         [self deleteRemoteFileWithName:name completionHandler:^(BOOL success, NSUInteger bytesAvailable, NSError * _Nullable error) {
-            if(!success) {
+            if (!success) {
                 failedDeletes[name] = error;
+            } else {
```
JulianKast, line 267: is this redundant, because you already set it in `deleteRemoteFileWithName` ?

PR #2104, opened 2022-08-03
Bugfix/issue 1633 open sub menu

### LINKED ISSUE #1633: Default cellID value passed in during openSubMenu method call
##### Bug Report
If you recreate and pass in a `SDLMenuCell` with submenu that already exists in the current default SDL `menu` to the `SDLMenuManager`'s `openSubMenu:` method, the request will fail as it will use the default `cellID` value of `UINT32_MAX` (4294967295). We should add a check for the default `cellID` value and return `NO` with an error message when this happens.

##### Reproduction Steps
1. Connect app to sdl_core and launch the SDL app. Send a `SDLMenuCell` with submenu to the built-in SDL menu.
2. Recreate the same `SDLMenuCell` with submenu and pass it into a `openSubMenu:` call.
3. Observe behavior.

##### Expected Behavior
`openSubMenu` should catch error and return `NO`

##### Observed Behavior
sdl_core (debug mode) shuts down.

* iOS Version: n/a
* SDL iOS Version: 6.6.0rc-1
* Testing Against: sdl_core (commit: d2331b066ebb30305f1a5df644511f0f72a2381f (HEAD -> release/6.1.0)) and sdl_hmi (commit cd3da85cdb7240a7b5f88d016dea7604158642b1 (HEAD -> Master)

jshivabeharry:
Fixes #1633 

### Risk

### Testing Plan

#### Unit Tests
Created a unit test to check when a submenu item is added to the menu and the same recreated cell is opened with openSubmenu it performs a successful transaction and is able to send an RPC.

#### Core Tests
In manticore I created a function that returns a menu with a submenu. I Added this menu into the screen managers menu property. I then called the function again to recreate the same menu cell inside a ```openSubMenu``` call. Before, it would fail with an error due to the default cell ID set to ```UINT32_MAX```, after changes it would successfully open the menu cell.

Core version / branch / commit hash / module tested against: sdl_core 8.1.0 (manticore)
HMI name / version / branch / commit hash / module tested against: generic_hmi 0.12.0 (manticore)

### Summary
- Added check for a clone cell in ```openMenu```
- Added additional unit test in ```SDLMenuManagerSpec.m```

##### Breaking Changes
* NA

##### Enhancements
* NA

##### Bug Fixes
* Fixes ```openMenu``` and ```openSubMenu``` calls when trying to open a copied cell

### CLA

### DIFF
--- a/SmartDeviceLink/private/SDLMenuManager.m
+++ b/SmartDeviceLink/private/SDLMenuManager.m
@@ -227,6 +227,16 @@ NS_ASSUME_NONNULL_BEGIN
         return NO;
     }
 
+    // Check for cloned cell
+    if (cell != nil) {
+        for(id clonedCell in self.menuCells) {
+            if ([cell isEqual:clonedCell]) {
+                cell = clonedCell;
+                break;
+            }
+        }
+    }
+    
     // Create the operation
     SDLMenuShowOperation *showMenuOp = [[SDLMenuShowOperation alloc] initWithConnectionManager:self.connectionManager toMenuCell:cell completionHandler:^(NSError * _Nullable error) {
         if (error != nil) {
--- a/SmartDeviceLink/private/SDLMenuManager.m
+++ b/SmartDeviceLink/private/SDLMenuManager.m
@@ -228,7 +228,7 @@ NS_ASSUME_NONNULL_BEGIN
     }
 
     // Check if a passed cell is a "re-created" cell without a cellID. If it is, then try to find the equivalent cell and use it instead
-    if (cell != nil) {
+    if (cell != nil && cell.cellId == UINT32_MAX) {
         for(id clonedCell in self.menuCells) {
             if ([cell isEqual:clonedCell]) {
                 cell = clonedCell;
--- a/SmartDeviceLink/private/SDLMenuManager.m
+++ b/SmartDeviceLink/private/SDLMenuManager.m
@@ -229,7 +229,7 @@ NS_ASSUME_NONNULL_BEGIN
 
     // Check if a passed cell is a "re-created" cell without a cellID. If it is, then try to find the equivalent cell and use it instead
     if (cell != nil && cell.cellId == UINT32_MAX) {
-        for(id clonedCell in self.menuCells) {
+        for (SDLMenuCell *headUnitCell in self.menuCells) {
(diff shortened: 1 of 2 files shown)


SmartDeviceLink/private/SDLMenuManager.m
```diff
@@ -227,6 +227,16 @@ - (BOOL)openMenu:(nullable SDLMenuCell *)cell {
         return NO;
     }
 
+    // Check for cloned cell
```
joeljfischer, line 230:
```suggestion
    // Check if a passed cell is a "re-created" cell without a cellID. If it is, then try to find the equivalent cell and use it instead
```

SmartDeviceLink/private/SDLMenuManager.m
```diff
@@ -227,6 +227,16 @@ - (BOOL)openMenu:(nullable SDLMenuCell *)cell {
...
+    if (cell != nil) {
```
joeljfischer, line 231:
```suggestion
    if (cell != nil && cell.cellId == UINT32_MAX) {
```

SmartDeviceLink/private/SDLMenuManager.m
```diff
@@ -227,6 +227,16 @@ - (BOOL)openMenu:(nullable SDLMenuCell *)cell {
...
+        for(id clonedCell in self.menuCells) {
```
joeljfischer, line 232:
```suggestion
        for (SDLMenuCell *headUnitCell in self.menuCells) {
```

Make sure to be as specific as possible

SmartDeviceLinkTests/DevAPISpecs/SDLMenuManagerSpec.m
```diff
@@ -269,6 +269,17 @@ @interface SDLMenuManager()
                     expect(canSendRPC).to(equal(YES));
                 });
 
+                // should queue an open menu operation for a copied submenu cell
+                it(@"should queue an open menu operation for a copied submenu cell", ^ {
```
joeljfischer, line 273: Tests that are added for a bug fix like this should fail before the fix is applied. This one passes.

joeljfischer (changes requested):

joeljfischer (approved):

PR #2107, opened 2022-08-29
Bugfix/issue 1781 good text failing

### LINKED ISSUE #1781: Setting bad data in one T&G field then good data quickly in another can lead to the good data failing.
### Bug Report
Based on how the Text & Graphic manager currently works, it is possible to have good data fail. A possible solution is that when a `Show` fails and the `currentScreenData` gets updated to also update the `updateState` of the T&G operation as well. This will require knowing which fields the T&G operation is updating so that they don't also get reset. 

With a similar method we could potentially also stop sending all Show data and only send updated Show data.

##### Reproduction Steps
1. Set "bad" data (data that will get rejected) into `TextField1` and start the operation.
2. Set "good" data (data that will not get rejected) into `TextField2` and start the operation _before_ the first operation gets a `Show` response.

##### Expected Behavior
`TextField1` should remain in its original state and `TextField2` should show the new data.

##### Observed Behavior
`TextField1` and `TextField2` will show the original state data because `TextField2`'s operation includes the `TextField1` data as well.

* iOS Version: n/a
* SDL iOS Version: 7.0.0-develop
* Testing Against: n/a

This was not verified on device, but by looking at the code.

None

jshivabeharry:
Fixes #1781 

### Risk

### Testing Plan

#### Unit Tests
Created two unit tests in `SDLTextAndGraphicManagerSpec`. One to test without batching if invidual operations are created and not cancelled. Another, to test the currentDataUpdatedHandler when given the error state.

#### Core Tests
In manticore verified that properties in screen manager when set would not fail when one attribute was given bad data.

Core version / branch / commit hash / module tested against: sdl_core 8.1.0 (manticore)
HMI name / version / branch / commit hash / module tested against: generic_hmi 0.12.0 (manticore)

### Summary
- Removed cancellation of mutliple `SDLTextAndGraphicUpdateOperation`.
- In `sdl_updatePendingOperationsWithNewScreenData` update pending operations updated state property based on previously failed updates.
- Add errorScreenData parameter to `currentScreenDataUpdatedHandler`.
- In `SDLTextAndGraphicUpdateOperation` add method to updated teh state with error data.
- Added additional unit test.

##### Enhancements
* Update screen data even if other screen properties were given bad data.

##### Bug Fixes
* Update screen data even if other screen properties were given bad data.

### Tasks Remaining:
- [x] [Create Fix]
- [x] [Add unit test
- [ ] [Code review]

### CLA

### DIFF
--- a/SmartDeviceLink/private/SDLTextAndGraphicManager.m
+++ b/SmartDeviceLink/private/SDLTextAndGraphicManager.m
@@ -178,7 +178,10 @@ NS_ASSUME_NONNULL_BEGIN
         } else if (error != nil) {
             // Invalidate data that's different from our current screen data if a Show or SetDisplayLayout fails. This will prevent subsequent `Show`s from failing if the request failed due to the developer setting invalid data or subsequent `SetDisplayLayout`s from failing if the template is not supported on the module. 
             [strongSelf sdl_resetFieldsToCurrentScreenData];
-            [strongSelf sdl_updatePendingOperationsWithFailedScreenState:error.userInfo[SDLTextAndGraphicFailedScreenStateErrorKey]];
+            SDLTextAndGraphicState *errorState = [error.userInfo objectForKey:SDLTextAndGraphicFailedScreenStateErrorKey];
+            if (errorState) {
+                [strongSelf sdl_updatePendingOperationsWithFailedScreenState:errorState];
+            }
         }
     } updateCompletionHandler:handler];
 
--- a/SmartDeviceLink/private/SDLTextAndGraphicManager.m
+++ b/SmartDeviceLink/private/SDLTextAndGraphicManager.m
@@ -173,16 +173,16 @@ NS_ASSUME_NONNULL_BEGIN
     }
 
     __weak typeof(self) weakSelf = self;
-    SDLTextAndGraphicUpdateOperation *updateOperation = [[SDLTextAndGraphicUpdateOperation alloc] initWithConnectionManager:self.connectionManager fileManager:self.fileManager currentCapabilities:self.windowCapability currentScreenData:self.currentScreenData newState:[self currentState] currentScreenDataUpdatedHandler:^(SDLTextAndGraphicState *_Nullable newScreenData, NSError *_Nullable error) {
+    SDLTextAndGraphicUpdateOperation *updateOperation = [[SDLTextAndGraphicUpdateOperation alloc] initWithConnectionManager:self.connectionManager fileManager:self.fileManager currentCapabilities:self.windowCapability currentScreenData:self.currentScreenData newState:[self currentState] currentScreenDataUpdatedHandler:^(SDLTextAndGraphicState *_Nullable newScreenData, NSError *_Nullable error, SDLTextAndGraphicState *_Nullable errorScreenData) {
         __strong typeof(weakSelf) strongSelf = weakSelf;
         if (newScreenData != nil) {
             // Update our current screen data
             strongSelf.currentScreenData = newScreenData;
-            [strongSelf sdl_updatePendingOperationsWithNewScreenData:newScreenData];
+            [strongSelf sdl_updatePendingOperationsWithNewScreenData:newScreenData errorState:nil];
         } else if (error != nil) {
             // Invalidate data that's different from our current screen data if a Show or SetDisplayLayout fails. This will prevent subsequent `Show`s from failing if the request failed due to the developer setting invalid data or subsequent `SetDisplayLayout`s from failing if the template is not supported on the module. 
             [strongSelf sdl_resetFieldsToCurrentScreenData];
-            [strongSelf sdl_updatePendingOperationsWithNewScreenData:strongSelf.currentScreenData];
+            [strongSelf sdl_updatePendingOperationsWithNewScreenData:strongSelf.currentScreenData errorState:errorScreenData];
         }
     } updateCompletionHandler:handler];
 
@@ -195,12 +195,13 @@ NS_ASSUME_NONNULL_BEGIN
     [self.transactionQueue addOperation:updateOperation];
 }
 
-- (void)sdl_updatePendingOperationsWithNewScreenData:(SDLTextAndGraphicState *)newScreenData {
+- (void)sdl_updatePendingOperationsWithNewScreenData:(SDLTextAndGraphicState *)newScreenData errorState:(SDLTextAndGraphicState *_Nullable)errorData {
     for (NSOperation *operation in self.transactionQueue.operations) {
         if (operation.isExecuting) { continue; }
         SDLTextAndGraphicUpdateOperation *updateOp = (SDLTextAndGraphicUpdateOperation *)operation;
 
         updateOp.currentScreenData = newScreenData;
+        [updateOp updateStateDataWithErrorData:errorData];
     }
 }
 
--- a/SmartDeviceLink/private/SDLTextAndGraphicManager.m
+++ b/SmartDeviceLink/private/SDLTextAndGraphicManager.m
@@ -197,7 +197,10 @@ NS_ASSUME_NONNULL_BEGIN
         SDLTextAndGraphicUpdateOperation *updateOp = (SDLTextAndGraphicUpdateOperation *)operation;
 
         updateOp.currentScreenData = newScreenData;
-        [updateOp updateStateDataWithErrorData:errorData];
+
+        if (errorData != nil) {
+            [updateOp updateStateDataWithErrorData:errorData];
+        }
     }
 }
 
--- a/SmartDeviceLink/private/SDLTextAndGraphicManager.m
+++ b/SmartDeviceLink/private/SDLTextAndGraphicManager.m
@@ -205,7 +205,7 @@ NS_ASSUME_NONNULL_BEGIN
         if (operation.isExecuting) { continue; }
         SDLTextAndGraphicUpdateOperation *updateOp = (SDLTextAndGraphicUpdateOperation *)operation;
 
-        [updateOp updateStateDataWithErrorData:errorState];
+        [updateOp updateTargetStateWithErrorState:errorState];
     }
 }
 
--- a/SmartDeviceLink/private/SDLTextAndGraphicManager.m
+++ b/SmartDeviceLink/private/SDLTextAndGraphicManager.m
@@ -178,7 +178,7 @@ NS_ASSUME_NONNULL_BEGIN
         } else if (error != nil) {
             // Invalidate data that's different from our current screen data if a Show or SetDisplayLayout fails. This will prevent subsequent `Show`s from failing if the request failed due to the developer setting invalid data or subsequent `SetDisplayLayout`s from failing if the template is not supported on the module. 
             [strongSelf sdl_resetFieldsToCurrentScreenData];
-            [strongSelf sdl_updatePendingOperationsWithFailedScreenState:error.userInfo[@"failedScreenState"]];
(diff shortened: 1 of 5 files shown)


SmartDeviceLink/private/SDLTextAndGraphicManager.m
```diff
@@ -195,12 +191,16 @@ - (void)sdl_updateAndCancelPreviousOperations:(BOOL)supersedePreviousOperations
     [self.transactionQueue addOperation:updateOperation];
 }
 
-- (void)sdl_updatePendingOperationsWithNewScreenData:(SDLTextAndGraphicState *)newScreenData {
+- (void)sdl_updatePendingOperationsWithNewScreenData:(SDLTextAndGraphicState *)newScreenData errorState:(SDLTextAndGraphicState *_Nullable)errorData {
```
joeljfischer, line 194: I _think_ that this can be separated into two different methods. It appears that the operation will only pass either an error or new current screen data. So this could be two different methods:

`sdl_updatePendingOperationsWithNewScreenState:`
and
`sdl_updatePendingOperationsWithFailedScreenState:`

SmartDeviceLink/private/SDLTextAndGraphicUpdateOperation.h
```diff
@@ -22,7 +22,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 typedef void(^SDLTextAndGraphicUpdateCompletionHandler)(NSError *__nullable error);
-typedef void(^CurrentDataUpdatedHandler)(SDLTextAndGraphicState *__nullable newScreenData, NSError *__nullable error);
+typedef void(^CurrentDataUpdatedHandler)(SDLTextAndGraphicState *__nullable newScreenData, NSError *__nullable error, SDLTextAndGraphicState *__nullable errorScreenData);
```
joeljfischer, line 25: I kind of think that instead of passing a separate handler parameter here, we should just package it in the `error.userInfo` dictionary. Key could be something like `@"failedScreenState"`. Then we can get rid of the error data.

SmartDeviceLink/private/SDLTextAndGraphicUpdateOperation.h
```diff
@@ -38,6 +38,10 @@ typedef void(^CurrentDataUpdatedHandler)(SDLTextAndGraphicState *__nullable newS
 /// @param updateCompletionHandler The handler potentially passed by the developer to be called when the update finishes
 - (instancetype)initWithConnectionManager:(id<SDLConnectionManagerType>)connectionManager fileManager:(SDLFileManager *)fileManager currentCapabilities:(SDLWindowCapability *)currentCapabilities currentScreenData:(SDLTextAndGraphicState *)currentData newState:(SDLTextAndGraphicState *)newState currentScreenDataUpdatedHandler:(CurrentDataUpdatedHandler)currentDataUpdatedHandler updateCompletionHandler:(nullable SDLTextAndGraphicUpdateCompletionHandler)updateCompletionHandler;
 
+/// Changes updated state to remove potential errors from previous update operations
+/// @param errorData A updated state that failed in a previous operation that will be used to filter out erroneous data
+- (void)updateStateDataWithErrorData:(SDLTextAndGraphicState *)errorData;
```
joeljfischer, line 43:
```suggestion
/// Changes updated state to remove failed updates from previous update operations. This will find and revert those failed updates back to current screen data so that we don't duplicate the failure.
/// @param errorState A updated state that failed in a previous operation that will be used to filter out erroneous data
- (void)updateTargetStateWithErrorState:(SDLTextAndGraphicState *)errorState;
```

SmartDeviceLink/private/SDLTextAndGraphicUpdateOperation.m
```diff
@@ -97,6 +97,23 @@ - (void)start {
     }
 }
 
+- (void)updateStateDataWithErrorData:(SDLTextAndGraphicState *)errorData {
+    self.updatedState.textField1 = errorData.textField1 == self.updatedState.textField1 ? self.currentScreenData.textField1 : self.updatedState.textField1;
```
joeljfischer, line 101:
```suggestion
    self.updatedState.textField1 = (errorData.textField1 == self.updatedState.textField1) ? self.currentScreenData.textField1 : self.updatedState.textField1;
```

etc. down the method

joeljfischer (changes requested):
Additional Notes:

1. This is not a minor version update (from the PR text). No public file had a public interface change (the operations are not public, you can check which are public by looking at the attached image)
Screen Shot 2022-08-30 at 9 01 28 AM

SmartDeviceLinkTests/DevAPISpecs/SDLTextAndGraphicManagerSpec.m
```diff
@@ -475,11 +480,22 @@ @interface SDLTextAndGraphicUpdateOperation ()
...
+                SDLTextAndGraphicState *errorState = [[SDLTextAndGraphicState alloc] init];
+                errorState.textField1 = @"Bad Data";
+                testManager.textField1 = errorState.textField1;
+                testManager.textField4 = @"Good Data";
+                testOperation3 = testManager.transactionQueue.operations[3];
+
+                NSDictionary *userInfo = @{
+                    @"failedScreenState": errorState
+                };
+                testOperation.currentDataUpdatedHandler(nil, [NSError errorWithDomain:@"any" code:1 userInfo:userInfo]);
```
joeljfischer, line 493:
```suggestion
                // Create a "bad data" text field 1, then set it in the manager, which should create an operation (op 2)
                SDLTextAndGraphicState *errorState = [[SDLTextAndGraphicState alloc] init];
                errorState.textField1 = @"Bad Data";
                testManager.textField1 = errorState.textField1;
                
                // Create a "good data text field 4, which should create a second operation (op 3)
                testManager.textField4 = @"Good Data";
                testOperation3 = testManager.transactionQueue.operations[3];

                // Simulate a failure of the first operation
                NSDictionary *userInfo = @{
                    @"failedScreenState": errorState
                };
                testOperation.currentDataUpdatedHandler(nil, [NSError errorWithDomain:@"any" code:1 userInfo:userInfo]);
```

SmartDeviceLinkTests/DevAPISpecs/SDLTextAndGraphicManagerSpec.m
```diff
@@ -475,11 +480,22 @@ @interface SDLTextAndGraphicUpdateOperation ()
...
             });
 
-            it(@"should reset the manager's data", ^{
+            it(@"should reset the manager's data and update other operations updated state", ^{
                 expect(testManager.textField1).to(equal(testManager.currentScreenData.textField1));
+                expect(testOperation3.updatedState.textField1).to(equal(@"Test1"));
```
joeljfischer, line 498:
```suggestion
                expect(testOperation3.updatedState.textField1).to(equal(testManager.currentScreenData.textField1));
```

SmartDeviceLinkTests/DevAPISpecs/SDLTextAndGraphicUpdateOperationSpec.m
```diff
@@ -780,7 +780,7 @@
                 updatedState.textField3 = field3String;
                 updatedState.textField4 = field4String;
 
-                testOp = [[SDLTextAndGraphicUpdateOperation alloc] initWithConnectionManager:testConnectionManager fileManager:mockFileManager currentCapabilities:windowCapability currentScreenData:emptyCurrentData newState:updatedState currentScreenDataUpdatedHandler:^(SDLTextAndGraphicState *_Nullable newScreenData, NSError *_Nullable error) {
+                testOp = [[SDLTextAndGraphicUpdateOperation alloc] initWithConnectionManager:testConnectionManager fileManager:mockFileManager currentCapabilities:windowCapability currentScreenData:emptyCurrentData newState:updatedState currentScreenDataUpdatedHandler:^(SDLTextAndGraphicState * _Nullable newScreenData, NSError * _Nullable error) {
```
joeljfischer, line 783:
```suggestion
                testOp = [[SDLTextAndGraphicUpdateOperation alloc] initWithConnectionManager:testConnectionManager fileManager:mockFileManager currentCapabilities:windowCapability currentScreenData:emptyCurrentData newState:updatedState currentScreenDataUpdatedHandler:^(SDLTextAndGraphicState *_Nullable newScreenData, NSError *_Nullable error) {
```

Just to revert them back

joeljfischer (changes requested):
1. The `@"failedScreenState"` string should also be a string constant (because it's in a private header file, it'll still be private). In the operation .h file:

```objc
extern NSString *const SDLTextAndGraphicFailedScreenStateErrorKey;
```

In the .m:

```objc
NSString *const SDLTextAndGraphicFailedScreenStateErrorKey = @"failedScreenState";
```

joeljfischer (changes requested):
There also need to be `SDLTextAndGraphicUpdateOperation` test changes to test `- updateTargetStateWithErrorState:` and `sdl_updateGraphicsAndShow:` changes.

Everything else looks good.

SmartDeviceLinkTests/DevAPISpecs/SDLTextAndGraphicUpdateOperationSpec.m
```diff
@@ -797,6 +797,47 @@
...
+            errorState.textField1 = nil;
+            errorState.primaryGraphic = nil;
+            errorState.secondaryGraphic = nil;
+
+            [testOp updateTargetStateWithErrorState:errorState];
+
+            expect(updatedState.textField1).toNot(beNil());
+            expect(updatedState.primaryGraphic).toNot(beNil());
+            expect(updatedState.secondaryGraphic).toNot(beNil());
+        });
```
joeljfischer, line 838: These will need to check all of the data, not just selected items. Additionally, the format of the test should be all "updated state" data is full, all "current screen data" is `nil`. Then send an error state with data that matches updated state, and now everything should be empty. Then send an error state with data that doesn't match the updated state, and everything should still match its original state. Then send an error state with everything nil, and it should still match its original state. LMK if you have questions. We should want to make sure we're covering everything here.

joeljfischer (changes requested):

joeljfischer (approved):

SmartDeviceLink/private/SDLTextAndGraphicManager.m
```diff
@@ -182,7 +178,10 @@ - (void)sdl_updateAndCancelPreviousOperations:(BOOL)supersedePreviousOperations
         } else if (error != nil) {
             // Invalidate data that's different from our current screen data if a Show or SetDisplayLayout fails. This will prevent subsequent `Show`s from failing if the request failed due to the developer setting invalid data or subsequent `SetDisplayLayout`s from failing if the template is not supported on the module. 
             [strongSelf sdl_resetFieldsToCurrentScreenData];
-            [strongSelf sdl_updatePendingOperationsWithNewScreenData:strongSelf.currentScreenData];
+            SDLTextAndGraphicState *errorState = [error.userInfo objectForKey:SDLTextAndGraphicFailedScreenStateErrorKey];
```
joeljfischer, line 181:
```suggestion
            SDLTextAndGraphicState *errorState = error.userInfo[SDLTextAndGraphicFailedScreenStateErrorKey];
```

SmartDeviceLink/private/SDLTextAndGraphicManager.m
```diff
@@ -182,7 +178,10 @@ - (void)sdl_updateAndCancelPreviousOperations:(BOOL)supersedePreviousOperations
...
+            if (errorState) {
```
joeljfischer, line 182:
```suggestion
            if (errorState != nil) {
```

What you had is fine, but the style of the project is to do an explicit check in this case.

joeljfischer (approved):

